PR #49, opened 2022-06-14
FilteredData constructor

### LINKED ISSUE #33: Change constructor of the FilteredData to not depend on TealData
part of https://github.com/insightsengineering/NEST-roadmap/issues/32

- [x] Initialization of the FilteredData should not depend on a TealData nor JoinKeys nor CodeClass. Initialize should require list of data, list of join keys.
- [x] Initialization of the FilteredDataset should not depend on a TealDataset but on a class of the dataset (data.frame or MultiAssayExperiment). FilteredDataset should receive dataname, (dataset) label, primary keys, col labels, metadata, parent. 
- [x] FilteredDataset should not hold TealDataset but raw data and it's attributes.
- [x] We don't need to keep copy of the var_labels in the class - they are stored in column attributes and they are not dropped on filter nor on merge (only on mutate).
- ~~CodeClass should not be included in the FilteredData. But it's "silently" used in get_rcode to combine reproducible code. We need to find the way how to assure reproducibility in the teal_module(s)~~ (to be addressed later)
- [x] Make sure the distinction between cdisc and non cdisc data.frames are handled correctly

So by default it should be possible to initialize FilterPanel using following:
```r
init_filtered_data(
  data = list(
    adsl = synthetic_cdisc_data("latest")$adsl
    adtte = synthetic_cdisc_data("latest")$adtte
    mae = MultiAssayExperiment::miniACC
  ), 
  join_keys = list(...)
)
```

image

Things to resolve:
- how to pass attributes of the dataset along with the list of data

nikolas-burkoff:
Closes #33 
- [ ] this PR
- [ ] https://github.com/insightsengineering/teal.data/pull/75
- [ ] https://github.com/insightsengineering/teal/pull/670
- [ ] https://github.com/insightsengineering/teal.transform/pull/73
- [ ] https://github.com/insightsengineering/teal.modules.general/pull/417


gogonzo (dismissed):
Filtering parent doesn't affect the child. See efficacy app for example. `get_call` in CDISCFilteredData has been modified but this call isn't executed anywhere. This means that relationship between reactive parent and child has been lost.

tests/testthat/test-FilteredData.R
```diff
@@ -345,35 +340,49 @@ testthat::test_that(
...
 
 testthat::test_that("get_filter_overview throws error with wrong argument input", {
+  datasets <- get_filtered_data_object()
   testthat::expect_error(isolate(datasets$get_filter_overview("AA")), "Some datasets are not available:")
   testthat::expect_error(isolate(datasets$get_filter_overview("")), "Some datasets are not available:")
   testthat::expect_error(isolate(datasets$get_filter_overview(23)), "Some datasets are not available:")
 })
 
 testthat::test_that("get_filter_overview returns overview matrix for non-filtered datasets", {
+  datasets <- get_filtered_data_object()
```
nikolas-burkoff, line 385: So this test and the following tests fail.

This is what is expected:

image

But we can no longer distinguish that `mock_iris` is not CDISC and `ADSL` is CDISC with no parent. Previously the former was `DefaultFilteredDataset` and the latter `CDISCFilteredDataset`. Now they are both `DefaultFilteredDataset` and then either they both have subjs as this branch now creates a CDISCFilteredData object:
 
image

Or if I change the creation of datasets to use FilteredData we get neither with subjs:

image

Is this change of behaviour OK? More generally it seems like the design we've changed to is fighting against the dispatch mechanisms in R (we have cdisc, non cdisc and MAE data, on the dataset level we have MAE and not MAE and on the data level we have CDISC and not CDISC). I'm OK with this if it's a stepping stone to simplifying the filter panel and removing all the data but I don't really like it  Especially trying to understand the logic to get the filter information - maybe this logic can be simplified?

R/CDISCFilteredData.R
```diff
@@ -75,21 +61,6 @@
 #' states$queue_push(filter_state_adtte, queue_index = 1L, element_id = "PARAMCD")
 #'
 #' isolate(datasets$get_call("ADTTE"))
-#'
```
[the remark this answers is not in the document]
nikolas-burkoff: done

R/FilteredData.R
```diff
@@ -315,29 +352,51 @@ FilteredData <- R6::R6Class( # nolint
...
-    set_dataset = function(dataset) {
-      stopifnot(is(dataset, "TealDataset") || is(dataset, "TealDatasetConnector"))
-      dataname <- teal.data::get_dataname(dataset)
+    #' @param dataset_args (`list`)\cr
+    #'   containing the arguments except (`dataname`)
+    #'   needed by `init_filtered_dataset`
+    #' @param dataname (`string`)\cr
+    #'   the name of the `dataset` to be added to this object
+    #' @return (`self`) invisibly this `FilteredData`
+    set_dataset = function(dataset_args, dataname) {
```
gogonzo, line 366: I'd send it to private field now

R/FilteredData.R
```diff
@@ -315,29 +352,51 @@ FilteredData <- R6::R6Class( # nolint
...
+      })
+
       invisible(self)
     },
 
+    #' @description
+    #' TODO
+    #' @param keys TODO
+    #' @return (`self`) invisibly this `FilteredData`
+    set_join_keys = function(keys) {
```
gogonzo, line 395: this should be a private now

R/FilteredData.R
nikolas-burkoff: so I was thinking about this - could there be a situation where you've created the filtereddata and then want to add more into it? And not have to recreate the whole object - if so this and join_keys would be public otherwise yup can move to private

nikolas-burkoff: though if I make it private I don't need to test it 😂

gogonzo: Good point, let's keep it as is.

nikolas-burkoff:
Should we change keys argument to primary_keys and join_keys as it can get confusing that keys sometimes refers to one and sometimes the other?

gogonzo:
@nikolas-burko

> Should we change keys argument to primary_keys and join_keys as it can get confusing that keys sometimes refers to one and sometimes the other?

@nikolas-burkoff I wonder if keys is necessary at all. primary keys are basically join_keys A<->A.

nikolas-burkoff:
True but join_keys lives in FilteredData and primary keys live in filtereddataset - and I think they are needed there for icons in filtering panel etc.?

R/FilteredData.R
```diff
@@ -315,29 +352,51 @@ FilteredData <- R6::R6Class( # nolint
...
-    #' @return (`self`) invisibly this `FilteredTealData`
-    set_dataset = function(dataset) {
-      stopifnot(is(dataset, "TealDataset") || is(dataset, "TealDatasetConnector"))
-      dataname <- teal.data::get_dataname(dataset)
+    #' @param dataset_args (`list`)\cr
+    #'   containing the arguments except (`dataname`)
+    #'   needed by `init_filtered_dataset`
+    #' @param dataname (`string`)\cr
+    #'   the name of the `dataset` to be added to this object
+    #' @return (`self`) invisibly this `FilteredData`
```
kpagacz, line 365: It is unclear from the docs what exactly goes into this list. It says needed by `init_filtered_dataset`, but it would not hurt to have it documented firstly here and then link from init_filtered_dataset to here, considering this method and its validation will be what ultimately decides what can go into it.

Also - we already pass a list of args, why not lump dataname with this list? One argument less to maintain xD

A reasonable thing to consider is to formalize this `dataset_args` as an s3 type. If it pops up anywhere else (`init_filtered_dataset`?), it can help with having validation separate from the logic here. Alternatively, drop the whole list thing and just pass the elements separately as arguments to this method? As it is now, it will be hard for users to guess what can and cannot go into this list.

kpagacz: I guess this is a private class anyway, but it would be helpful for our devs as well.

R/CDISCFilteredData.R
```diff
@@ -10,45 +10,31 @@
...
-#' It is not too difficult to make this class generalize to more than one parent dataset,
-#' provided each dataset then contains information about its parents together with the
-#' shared keys with each of them.
-#'
 #' @seealso `FilteredData` class
 #'
 #' @examples
 #' library(scda)
 #' library(shiny)
+#' library(teal.data)
```
pawelru, line 21: `shiny` will be loaded together with teal.data
https://github.com/insightsengineering/teal.data/blob/ee99ab4126b2a3523a14085cf1ed7d5451b93ce6/DESCRIPTION#L18-L20
```suggestion
#' library(teal.data)
```

R/FilteredData.R
```diff
@@ -36,18 +36,22 @@
...
 #'
 #'
 #' df <- datasets$get_data("iris", filtered = FALSE)
 #' print(df)
 #'
-#' datasets$get_metadata("iris")
+#' datasets$get_metadata("mtcars")
 #'
 #' datasets$set_filter_state(
 #'   list(iris = list(Species = list(selected = "virginica")))
```
pawelru, line 57: This is incorrect as there is no `"Species"` column in `mtcars`

R/init_filitered_data.R
```diff
@@ -0,0 +1,68 @@
+#' @export
+init_filtered_data <- function(x, ...) {
+  UseMethod("init_filtered_data")
+}
```
pawelru, line 4: please enhance the docs

nikolas-burkoff: @pawelru this is still a draft PR :)

R/FilteredData.R
gogonzo: @nikolas-burkoff I agree, docs could be updated to include what are possible fields.

@kpagacz @nikolas-burkoff  ideally `datalabel`, `metadata` and `keys` (primary) should be an attribute of raw-data. But since inclusion of these attributes are not reflected anywhere in the R code (yet), we can't do this now.
`teal` doesn't do any operation on a raw-data which would discard column attributes (labels) or data attributes (datalabel, metadata). This is why collabels are not carried in the R6 classes anymore. I'm looking forward to allow us to include metadata and datalabel also.

R/FilteredDataset.R
```diff
@@ -140,29 +117,27 @@ FilteredDataset <- R6::R6Class( # nolint
...
-        filter_call <- self$get_call()
-        eval_expr_with_msg(filter_call, env)
-        get(x = self$get_filtered_dataname(), envir = env)
-      })
-
+      private$dataname <- dataname
+      private$keys <- keys
+      private$label <- if (is.null(label)) character(0) else label
+      private$metadata <- metadata
+      private$calculate_hash()
```
pawelru, line 140: can the hash be an active field recalculated on request? - we would have less fields/slots to maintain

nikolas-burkoff: I guess it depends on how quick it is to hash a dataset - when it's large could it take too long?

R/FilteredData.R
```diff
@@ -139,10 +161,14 @@ FilteredData <- R6::R6Class( # nolint
 
     #' @description
     #' Gets the R preprocessing code string that generates the unfiltered datasets
-    #' @param dataname (`character`) name(s) of teal.data::dataset(s)
+    #' @param dataname (`character`) name(s) of dataset(s)
     #' @return (`character`) deparsed code
```
pawelru, line 165:
```suggestion
    #' @return (`character(1)`) deparsed code
```

R/FilteredDataset.R
gogonzo: @pawelru I think we don't need this field to be dynamic now as unfiltered data is set once "forever". When we finaly figure out how to reload DDL this would be needed. At this point it's not relevant to add complexity to this one.

R/FilteredData.R
```diff
@@ -181,25 +225,10 @@ FilteredData <- R6::R6Class( # nolint
...
-    #'
-    #' @param dataname (`character`) name of the dataset
-    #' @param attr (`character`) attribute to get from the data attributes of the dataset
-    #' @return value of attribute, may be `NULL` if it does not exist
-    get_data_attr = function(dataname, attr) {
-      private$check_data_varname_exists(dataname)
-      checkmate::assert_string(attr)
-      teal.data::get_attrs(self$get_filtered_dataset(dataname)$get_dataset())[[attr]]
-    },
-
```
gogonzo, line 198: This fails as get_attrs is a method for TealDataset/TealData. These datasets no longer have attributes as varlabels can be obtained via get_varlabel. Only one occurrence downstream:
https://github.com/insightsengineering/teal.modules.general/pull/417/files

R/FilteredData.R
kpagacz: Just writing tests for this method. I had to go to a private `validate_dataset_args` method to check what it validates to write a proper test case.

R/FilteredData.R
```diff
@@ -69,7 +72,35 @@ FilteredData <- R6::R6Class( # nolint
...
+    #' @param data_objects (`list`) should contain.
+    #' - `dataset` data object object supported by [`FilteredDataset`].
+    #' - `metatada` (optional) additional metadata attached to the `dataset`.
+    #' - `keys` (optional) primary keys.
+    #' - `datalabel` (optional) label describing the `dataset`.
+    #' - `parent` (optional) which `dataset` is a parent of this one.
+    #' @param join_keys (`JoinKeys`) see [teal.data::join_keys()].
+    #' @param code (`CodeClass`) see [`teal.data::CodeClass`].
+    #' @param check (`logical(1)`) whether data has been check against reproducibility.
+    initialize = function(data_objects, join_keys, code = NULL, check = FALSE) {
```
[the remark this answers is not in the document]
nikolas-burkoff: yeah I'll change it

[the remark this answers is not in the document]
nikolas-burkoff: @kpagacz I like the idea of allowing length 0 data_objects and adding datasets in later using the api - as maybe there's a case for that outside of teal.data -> do you see any downstream effects for this?

kpagacz:
What happens with MAE right now? Or are we doing it in another PR?

nikolas-burkoff:
> EDIT: I think there might be a problem with mixing data.frame and MAE objects together with some join keys between them or parent/child relationship. Pretty sure it is going to go wrong.

So I think it is going to go the same wrong as it would have done before this PR? Maybe an issue in the backlog?

R/FilteredData.R
```diff
@@ -104,7 +134,7 @@ FilteredData <- R6::R6Class( # nolint
     #' @description
     #' Gets variable names of a given dataname for the filtering.
     #'
-    #' @param dataname (`character`) name of the dataset
+    #' @param dataname (`character(1)`) name of the dataset
     #' @return (`character` vector) of variable names
     get_filterable_varnames = function(dataname) {
```
kpagacz, line 139: Btw, what is the point of this method? XD

nikolas-burkoff: it's overwritten in CDISCFilteredData...

gogonzo: - in FilteredData it doesn't do anything. colnames(data) == filterable_varnames
- In [CDISCFIltereData](https://github.com/insightsengineering/teal.slice/blob/8a09ad5aaeab460d921218cd4f80c9b0c1af1a1d/R/CDISCFilteredData.R#L157) we don't want child to have duplicated columns. It's actually users request because in ADTTE there are some duplicated columns comparing to ADSL. I foreseen that in the future it might cause some confusion as we can for example build a relationship between movie <-> actor and movie.name with actor.name are duplicated names but not the same thing. Let's discuss this in some time in the future if it will be important.

gogonzo:
> * add issue to update filter-panel.Rmd currently I've just added a note saying what is out of date
> I guess we don't need to do this? As everything is going to be changing a lot now?
We need the update for sure. I'll put this to sprint 60 when we have everything possibly clarified and done.

https://github.com/insightsengineering/teal.slice/issues/52

> * work out a nice way of specifying keys/join keys for cdisc data so you don't have to specify them all - although this could go to a different issue?
> Do you think we need to do this? We could have `join_keys = teal::get_cdisc_join_keys()` to simplify creating a `CDISCFilteredData` object - if so I'll create an issue

Let's create the issue but doesn't have to be taken in this increment. https://github.com/insightsengineering/teal.slice/issues/53

gogonzo:
> > EDIT: I think there might be a problem with mixing data.frame and MAE objects together with some join keys between them or parent/child relationship. Pretty sure it is going to go wrong.
> 
> So I think it is going to go the same wrong as it would have done before this PR? Maybe an issue in the backlog?

@nikolas-burkoff is right. Nothing new. If the MAE users complain about this we would know. At this moment it's a gap we can ignore.

R/FilteredData.R
```diff
@@ -73,17 +73,20 @@ FilteredData <- R6::R6Class( # nolint
...
     #' - `parent` (optional) which `dataset` is a parent of this one.
     #' @param join_keys (`JoinKeys`) see [teal.data::join_keys()].
-    #' @param code (`CodeClass`) see [teal.data::CodeClass].
+    #' @param code (`CodeClass`) see [`teal.data::CodeClass`].
     #' @param check (`logical(1)`) whether data has been check against reproducibility.
     initialize = function(data_objects, join_keys, code = NULL, check = FALSE) {
       checkmate::assert_list(data_objects, any.missing = FALSE, min.len = 0, names = "unique")
-      # TODO other checks but internals of data_objects are checked in set_dataset
+      # Note the internals of data_objects are checked in set_dataset
+      checkmate::assert_class(join_keys, "JoinKeys", null.ok = TRUE)
```
gogonzo, line 86: let's make arg `join_keys = NULL` by default. Otherway it would fail

gogonzo (approved):
Setting FilteredData is simplified as much as it's possible at this moment. 

image

kpagacz (approved):
I have spent the morning launching the apps from the gallery and I have not noticed anything browfurrowing.

tests/testthat/test-MAEFilteredDataset.R
```diff
@@ -319,13 +269,22 @@ testthat::test_that(
...
   testthat::expect_identical(filtered_dataset$get_filterable_varnames(), character(0))
 })
 
+testthat::test_that("MAEFilteredDataset$get_varlabels returns column variable labels", {
+  utils::data(miniACC, package = "MultiAssayExperiment")
+  x <- miniACC
+  attr(SummarizedExperiment::colData(x)$ADS, "label") <- "ADS label"
+  filtered_dataset <- MAEFilteredDataset$new(dataset = x, dataname = "miniACC")
+  labels <- filtered_dataset$get_varlabels(c("COC", "ADS"))
+  testthat::expect_equal(c("COC"= NA, ADS = "ADS label"), labels)
```
kpagacz, line 282:
```suggestion
  testthat::expect_equal(c("COC" = NA, ADS = "ADS label"), labels)
```

PR #65, opened 2022-07-06
filterPanelAPI

### LINKED ISSUE #64: Pass filter_panel_api to the module
Our objective is teal_module(s) to receive the filter_panel_api instead of taking filter-states from the `FilteredData` object. In the flow-chart below what we need to do in this issue is distinguished by "bolded" frames. This means that if `teal_module` has a `filter_panel_api` argument, we need to pass a `FilterPanelAPI` object which have a FIlteredData in the private field with limited number of methods (preferable only [filter-state-api](https://github.com/insightsengineering/teal.slice/blob/acdcb38712f4a826d6fc1bee3fec755f55dd4924/R/filtered_data_wrappers.R#L6) + [formated_filter_states](https://github.com/insightsengineering/teal.slice/blob/acdcb38712f4a826d6fc1bee3fec755f55dd4924/R/FilteredData.R#L463)) 

FIlterPanelApi should be S4 (more preferable) or R6 with wrappers get_filter_state, add_filter_state, reset_filter_state dispatched also for this class (currently they are only for FilteredData)

quosure-teal-simplify-api drawio

```r
tm_module(
  a b c <instructions for choices> # currently data_extract
  data <list of reactives, optionally with expr, metadata and join_keys attr>, 
  filter_panel_api = <fite
  # datasets - filteredData
  # reporter - if reporter-api needed 
)
```

mhallal1:
closes #64 

-`FilterPanelAPI` R6 class added with methods to get, set an remove filters.
-these methods are used by the wrapper methods in `filtered_data_wrappers.R` to access and manipulate objects of class `FilteredData` and `FilterPanelAPI`.
-The documentation of these methods was added as in `FilteredData` methods.
-Tests were added.


gogonzo (commented):
All good here, but I can't see the necessary changes in teal (to include FilterPanelApi to the modules)

gogonzo (approved):
I just tested it with several tmg modules and it works like a charm.

PR #71, opened 2022-07-14
add joinkeys into filtered data

### LINKED ISSUE #63: Pass JoinKeys to the FilteredData
Council has agreed that we should use JoinKeys in all the places. It gives a constistency and auto-validation methods.

This means that `join_keys` slot in FilteredData should contain `JoinKeys`

nikolas-burkoff:
Closes #63 but note this is not fixing https://github.com/insightsengineering/teal/issues/683 note I've checked teal.transform and all the tests pass there and the vignette behaviour is the same as on main.

Please also tests a few tmg or tmc modules

### DIFF
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -250,7 +250,11 @@ FilteredData <- R6::R6Class( # nolint
       if (is.null(private$keys)) {
         return(character(0))
       }
-      private$keys$get(dataset_1, dataset_2)
+      keys <- private$keys$get(dataset_1, dataset_2)
+      if (length(keys) == 0) {
+        return(character(0))
+      }
+      return(keys)
     },
 
     #' @description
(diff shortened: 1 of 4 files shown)


R/FilteredData.R
```diff
@@ -247,19 +247,14 @@ FilteredData <- R6::R6Class( # nolint
     #' @param dataset_2 (`character(1)`) other dataset name
     #' @return (`named character`) vector with column names
     get_join_keys = function(dataset_1, dataset_2) {
```
nikolas-burkoff, line 249: So this keeps the previous behaviour - I'm not sure that is what we want though?

nikolas-burkoff: To be done here https://github.com/insightsengineering/teal.data/issues/78

junlueZH (approved):
The logic is correct. The API is unchanged. The new tests are adequate.

PR #75, opened 2022-07-20
Update get_filter_state

### LINKED ISSUE #61: FIlterState S3 format function
For teal.reporter reasons we need a format S3 method to process filter state. 
This function have to format the object recursively.

Function format.CLASSNAME
Input:
a list of a specific S3 class (e.g. fs) with elements of a specific S3 class (e.g. fs_Date)
Output:
a nice printout of the filter state

So this item requires the get_filter_state method to return a specific S3 class now.

mhallal1:
closes #61 
related PR: 

After a discussion with @nikolas-burkoff and @Polkas , we decided to update `FilteredData::get_filter_state` by adding an attribute to the returned list which is the output of `get_formatted_filter_state`.
The list is used to append meta data and the character output of `get_formatted_filter_state` is used to append text to the reporter.

Please test with anu tmg, tmc, tmg ... module.

### DIFF
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -417,11 +417,15 @@ FilteredData <- R6::R6Class( # nolint
     #'
     #' Gets all active filters in the form of a nested list.
     #' The output list is a compatible input to `self$set_filter_state`.
+    #' The attribute `formatted` renders the output of `self$get_formatted_filter_state` which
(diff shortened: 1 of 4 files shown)


Polkas (approved):
LGTM, please add the NEWS record. Now we could easily update the `teal.reporter`.

PR #78, opened 2022-07-21
modify get_formatted_filter_state

### LINKED ISSUE #56: FilterState format for empty Filters
When we as for the styled FilterState ( the format print), we should get a simple Empty Filters when there are no Filters.
Currently we get a line for each dataset and the filter are black for them.

From

Filters for ADSL:
Filters for ADTTE:

To

Empty FIlters

mhallal1:
closes #56 
Update `FilteredData$get_formatted_filter_state` to not append empty filters.


R/FilteredData.R
```diff
@@ -452,7 +452,11 @@ FilteredData <- R6::R6Class( # nolint
     #'
     get_formatted_filter_state = function() {
       out <- c()
-      for (filtered_dataset in self$get_filtered_dataset()) out <- c(out, filtered_dataset$get_formatted_filter_state())
+      for (filtered_dataset in self$get_filtered_dataset()) {
+        if (length(filtered_dataset$get_filter_state()) != 0) {
+          out <- c(out, filtered_dataset$get_formatted_filter_state())
+        }
+      }
```
gogonzo, line 459: Maybe this is more elegant than appending vector? Leaving up to you
```suggestion
      Filter(
        length
        sapply(
          self$get_filtered_dataset(), 
          function(filtered_dataset) {
            filtered_dataset$get_formatted_filter_state()
          }
        })
      )
```

gogonzo (approved):
I approve the idea of having attribute

R/FilteredData.R
```diff
@@ -452,7 +452,11 @@ FilteredData <- R6::R6Class( # nolint
     #'
     get_formatted_filter_state = function() {
       out <- c()
-      for (filtered_dataset in self$get_filtered_dataset()) out <- c(out, filtered_dataset$get_formatted_filter_state())
+      for (filtered_dataset in self$get_filtered_dataset()) {
+        if (length(filtered_dataset$get_filter_state()) != 0) {
+          out <- c(out, filtered_dataset$get_formatted_filter_state())
```
nikolas-burkoff, line 457: Could `filtered_dataset$get_formatted_filter_state()` return `NULL` if `length(filtered_dataset$get_filter_state()) == 0` this would then simplify the logic

mhallal1: the logic is simplified now as per Dawid's suggestion.

mhallal1: `filtered_dataset$get_formatted_filter_state()` returns `NULL` now and I replaced the for loop with a `Fitler` statement.

R/FilteredData.R
```diff
@@ -451,8 +451,15 @@ FilteredData <- R6::R6Class( # nolint
...
-      for (filtered_dataset in self$get_filtered_dataset()) out <- c(out, filtered_dataset$get_formatted_filter_state())
+      out <- Filter(
+        length,
+        sapply(
+          self$get_filtered_dataset(),
+          function(filtered_dataset) {
+            filtered_dataset$get_formatted_filter_state()
+          }
+        )
+      )
```
nikolas-burkoff, line 462: Do you actually need to filter out here? If they do return NULL now

image

mhallal1: no, removed and updated `teal.reporter` PR accordingly.

nikolas-burkoff (approved):

R/FilteredDataset.R
```diff
@@ -155,9 +155,19 @@ FilteredDataset <- R6::R6Class( # nolint
     #' @return `character(1)` the formatted string representing the filter state
```
nikolas-burkoff, line 155: or NULL?

mhallal1: updated

PR #85, opened 2022-08-09
271 rna seg warning double operator@pre release

BLAZEWIM:
# Pull Request

Fixes [#271](https://github.com/insightsengineering/teal.modules.hermes/issues/271)

It looks like a mistake done while copying: https://github.com/insightsengineering/teal.slice/blob/09c6ab2e026325b62c790c3759ac414701457a62/R/FilterStates.R#L1077
double operator does not make sense here for me.

Anyway, warning was produced because of: https://stackoverflow.com/questions/72848442/r-warning-lengthx-2-1-in-coercion-to-logical1/72848495#72848495 which came with R >= 4.2


BLAZEWIM:
wrong target branch, removing!

PR #86, opened 2022-08-09
double operator changed to single for MAEFilterStates

BLAZEWIM:
# Pull Request

Fixes #[271](https://github.com/insightsengineering/teal.modules.hermes/issues/271)

It looks like a mistake done while copying:
https://github.com/insightsengineering/teal.slice/blob/09c6ab2e026325b62c790c3759ac414701457a62/R/FilterStates.R#L1077

double operator does not make sense here for me on line 1456.

Anyway, warning was produced because of: https://stackoverflow.com/questions/72848442/r-warning-lengthx-2-1-in-coercion-to-logical1/72848495#72848495 which came with R >= 4.2

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,11 +5,11 @@
 * The filtered data is now stored in `FilteredData` not `FilteredDataset`.
 * The join keys stored inside `FilteredData` are now `JoinKeys` objects.
 * Updated `get_filter_state` to return a list of active filter states and an attribute with the character form of the filter states.
+* In `FilterStates`, `varlabels` are now described as character vector, also removed usage of `||` to acomodate for vector input.
 
 # Bug fixes
 
 * Fixed a bug when the filter panel overview would not refresh if the panel was hidden during a transition between active modules.
-* Fixed a bug where `varlabels` were assumed to be a character in `FilterStates` (now it is a character vector) and a `||` was used.
 
 # teal.slice 0.1.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@
 * The filtered data is now stored in `FilteredData` not `FilteredDataset`.
 * The join keys stored inside `FilteredData` are now `JoinKeys` objects.
 * Updated `get_filter_state` to return a list of active filter states and an attribute with the character form of the filter states.
-* In `FilterStates`, `varlabels` are now described as character vector, also removed usage of `||` to accommodate for vector input.
+* Updated `get_varlabels` method in the `FilterStates` classes, accepts a vector input.
 
 # Bug fixes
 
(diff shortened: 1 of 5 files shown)


Polkas (changes requested):
ok, so this method looks to be used improperly as it is NOT vectorized. 
Unfortunately sb not added a proper validation that the variable argument should be a string.
Please find where this method is used for a vector `variables` arg and if we need to support such scenario (one or many places).

For sure:
- [ ] add a validation about the `variables` argument, depending what we will decide about vectorization
- [x] we need a NEWS file entry with an information about the update
- [x] please do not evaluate the same expression twice

We have to check if we should expand this function and start to support the vectorized version. You will know it after the usage check. If we go this direction then we need:

- [x] Update the roxygen2 
- [x] Update the function what you propose

Polkas:
I will give more hints so the concerned method (`get_varlabels`) is a private one so could be used only internally.
You have 6 usages of `private$get_varlabels` in the teal.slice/R/FilterStates.R file.

BLAZEWIM:
With help from @gogonzo we decided to allow varlabels to be a vector. This use case is now only in MAEFilteredStates.
Checked locally, warning is gone.

NEWS.md
```diff
@@ -9,6 +9,7 @@
 # Bug fixes
 
 * Fixed a bug when the filter panel overview would not refresh if the panel was hidden during a transition between active modules.
+* Fixed a bug where `varlabels` were assumed to be a character in `FilterStates` (now it is a character vector) and a `||` was used.
```
Polkas, line 12: In my opinion it is an enhancement not a bug fix as the method was working properly under the defined in documentation usage.

tests/testthat/test-MAEFilterStates.R
```diff
@@ -3,7 +3,7 @@ testthat::test_that("The constructor does not throw", {
     input_dataname = "test",
     output_dataname = "test",
     datalabel = character(0),
-    varlabels = character(0),
+    varlabels = "varlabel1",
```
Polkas, line 6: I will comment it here as a remind, please revert these changes. We should stay with the default for the old tests.

NEWS.md
BLAZEWIM: moved to info part

tests/testthat/test-MAEFilterStates.R
BLAZEWIM: done in new commit

NEWS.md
```diff
@@ -5,6 +5,7 @@
 * The filtered data is now stored in `FilteredData` not `FilteredDataset`.
 * The join keys stored inside `FilteredData` are now `JoinKeys` objects.
 * Updated `get_filter_state` to return a list of active filter states and an attribute with the character form of the filter states.
+* In `FilterStates`, `varlabels` are now described as character vector, also removed usage of `||` to accomodate for vector input.
```
Polkas:
```suggestion
* Updated `get_varlabels` method in the `FilterStates` classes, accepts a vector input.
```

BLAZEWIM: ok

Polkas (approved):
LGTM:)

Issue #84, opened 2022-08-04
Introduce disable filters in FilterPanelApi/FilteredData


gogonzo:
Please add option to disable filters for particular module. This particular module should get unfiltered data when toggle filter through FilterPanelApi

nikolas-burkoff:
Should this be module specific? Isn't it easier for both app developers and app users to have one global toggle which turns the filter panel on/off for all modules? i.e. [here](https://github.com/insightsengineering/teal.slice/pull/91)

donyunardi:
Hi @kumamiao, can you help us on this? Do you think user would prefer global on/off or module-level on/off? 
Or maybe both?

kumamiao:
I feel like should we consider putting a hold on this issue, as we're planning for a bigger design change for the filter panel (which could potentially involve the setup for global & module-specific filters)?

gogonzo:
@kumamiao we prefer to go through as we are working on the feature branch. This will be easier for developers to move forward with having quosure implemented in all modules. Disable/enable filters can be reverted and modified at any point.

kumamiao:
In this case let's move forward with this issue. I would propose to go with what Nik suggested, have one global toggle which turns the filter panel on/off for all modules - although one concern here, if switched from on->off->on, would the default choices come back to previous selected filters, or everything would be wiped off?

In the future if we want to enable and then implement module-level toggle, then that toggle would be only applicable for module-specific filters and not impact global filters.

nikolas-burkoff:
> although one concern here, if switched from on->off->on, would the default choices come back to previous selected filters, or everything would be wiped off? The former would be preferred if possible

So I agree when turning the filter panel back on - the filters which were active at the time the filter panel was turned off should be turned back on again - as is the case [here](https://github.com/insightsengineering/teal.slice/pull/91)

PR #91, opened 2022-08-22
turn off on panel

Polkas:


nikolas-burkoff:
Does this close https://github.com/insightsengineering/teal.slice/issues/84

R/FilteredData.R
```diff
@@ -759,6 +771,23 @@ FilteredData <- R6::R6Class( # nolint
             }
           )
 
+          observeEvent(
+            eventExpr = list(input[["filter_turn_onoff"]]),
+            handlerExpr = {
+              if (isFALSE(input[["filter_turn_onoff"]])) {
+                shinyjs::hide("filter_add_vars")
+                shinyjs::hide("filter_active_vars")
```
nikolas-burkoff, line 779: shinyjs::disable instead of hide? I think that's nicer - though that would then mean you'd need some more logic `if input$switch-button` somewhere

nikolas-burkoff:
We also need to correctly (whatever that means) handle the case where the filter panel is disabled but the module developer calls the filter panel api to set states

R/FilteredData.R
Polkas: I thought of it, as you like it more (and I am neural) so I will apply it:)

Polkas:
> Does this close #84 - I guess you also need the trigger in FilteredDataAPI

No, this is a button in the global scope not local.

nikolas-burkoff:
- Do you think there's a way of linking all these together (i.e. some panel with light shading to show that the whole of this is the filter panel?) - and maybe a title "filtering" 

image

- Also I don't like the word "Turn" there maybe An inline label "Filtering:" then the switch you have with labels "Enabled" + "Disabled"

- There's a picture of the filter panel in the teal vignette which I think should be updated (or a ticket added to update it)

I'm aware there is a big filter panel UI change coming so not sure whether these should be done here or not

gogonzo (commented):
tests are missing, besides looks good to me

R/FilteredData.R
```diff
@@ -891,6 +945,12 @@ FilteredData <- R6::R6Class( # nolint
     # private attributes ----
     filtered_datasets = list(),
 
+    # turn on / off filter panel
+    filter_turn = TRUE,
```
pawelru, line 949: 1. I dislike the "turn" term here. You can _turn_ left or right. I propose _state_ / _status_ / _active_ instead. Please change other places as well including method docs.

2. What's the point of this field? I ask from the class simplicity point of view.
I can see that it's used in the api class - there is a check whether it's active and if yes then you can do set/get/etc. Is that needed? The functionality would be anyway disabled from the GUI perspective (vide: disabling inputs). What I am suggesting here is that the big disable/enable button would _just_ propagate this action to all relevant inputs + save/restore action and that's it. An object would not store its state as this opens a door for unnecessary complexity I believe. I am happy to discuss it further if needed.

R/FilteredData.R
```diff
@@ -573,6 +577,33 @@ FilteredData <- R6::R6Class( # nolint
     restore_state_from_bookmark = function(state, check_data_hash = TRUE) {
       stop("Pure virtual method")
     },
+    #' @description disable the filter panel
+    filter_panel_disable = function() {
+      private$filter_turn <- FALSE
+      shinyjs::disable("filter_add_vars")
+      shinyjs::disable("filter_active_vars")
```
pawelru, line 584: why there is no `shiny::ns()` here?

R/FilteredData.R
```diff
@@ -759,6 +801,18 @@ FilteredData <- R6::R6Class( # nolint
...
 
+          private$filter_panel_ui_id <- session$ns(NULL)
+          observeEvent(
+            eventExpr = input[["filter_turn_onoff"]],
+            handlerExpr = {
+              if (isTRUE(input[["filter_turn_onoff"]])) {
+                self$filter_panel_enable()
+              } else {
+                self$filter_panel_disable()
+              }
```
pawelru, line 812: I would suggest to put all of `shinyjs::disable/enable` calls here and use save/restore methods. This would keep `shiny` functionalities nicely decoupled - methods relates only to _data fields_ and `shiny` parts lives inside the modules.
Moreover, I do not see (and I couldn't imagine) other place where we would call `self$filter_panel_enable/disable()`.

pawelru, line 812: I would suggest to put all of `shinyjs::disable/enable` calls here and use save/restore methods. This would keep `shiny` functionalities nicely decoupled - methods relates only to _data fields_ and `shiny` parts lives inside the modules.
Moreover, I do not see (and I couldn't imagine) other place where we would call `self$filter_panel_enable/disable()`.

pawelru: On another side - please put logger entries

R/FilteredData.R
```diff
@@ -759,6 +801,18 @@ FilteredData <- R6::R6Class( # nolint
             }
           )
 
+          private$filter_panel_ui_id <- session$ns(NULL)
+          observeEvent(
+            eventExpr = input[["filter_turn_onoff"]],
+            handlerExpr = {
```
pawelru, line 807: please put logger entries

R/FilteredData.R
```diff
@@ -589,6 +620,17 @@ FilteredData <- R6::R6Class( # nolint
...
+          id = ns("switch-button"),
+          class = "inline-block",
+          shinyWidgets::switchInput(
+            ns("filter_turn_onoff"),
+            label = "TURN",
+            value = TRUE,
+            inline = FALSE,
+            handleWidth = "80px"
+          )
+        ),
```
pawelru, line 633: Please spend more time on UX here. IMO right now it is too maximal

R/FilteredData.R
Polkas: "other place where we would call" - DRY rule is not the only reason to separate the code. Nice that you are raising it, please check how overcrowded is the server of filter panel now. WE should work to make it more clear and readable.

R/FilteredData.R
Polkas: I like the current UI as is obvious and the functionality important, please give me your idea then.

R/FilteredData.R
Polkas: By default `shinyWidgets::switchInput` use the logical(1) for value so this should not be surprising I stay with this type.

`filter_turn` variable is needed only for filter_panel_api, so we could provide a proper warning if sb want to update the state when filter panel is turn off.

I like the Turn Off/On naming, although I see you agree with Nik in this area so please update to the one you are preferring.

R/FilteredData.R
Polkas: done

R/FilteredData.R
pawelru: > DRY rule is not the only reason to separate the code

Yes you are definitely right here but my first point remains.
Long time ago we have a lengthy discussions with @gogonzo and Konrad about decoupling shiny-related parts from the class definitions. Unfortunately there is no good design pattern and we decided to still keep it inside the class definition but only inside one module method or two ui/server methods. This indeed means that those the method(s) definitions would be lengthy but it also assures nice decoupling of class' methods from the shiny part which should serve just as an add-on which calls those methods. Separate shiny-methods would break that and make the code less separated.

R/FilteredData.R
pawelru: It's not about the importance of the feature - it's more how it fits into the overall GUI. Right now it doesn't fit well.
Maybe something like below? It's a minimal switch with decreased opacity.
image

donyunardi: I think the button that @Polkas proposed could work if we label it properly:

Screen Shot 2022-09-22 at 10 44 40 PM

But I also think the button that @pawelru proposed looks good too. 

@kumamiao do you have any opinion on how which button that makes the most sense to the user?

donyunardi: Also, @kumamiao, can I also hear your opinion on the wording 'On/Off' or 'Enabled/Disabled' for the button toggle? Which one do you think user will prefer?

kumamiao: I am a fan of Pawel's proposition of the On/Off switch for Active filtering which is quite straightforward
image
The current button can be a bit misleading and takes more space

pawelru:
Thank you - this is much better now!

The screenshot you provided made me think about yet another idea - how about placing the switch next to the show/hide button (hamburger one). Those are conceptually functionalities affecting the same object - filter. WDYT?

Polkas:
@pawelru very interesting proposition. 
So the **hamburger button is a part of teal::init not the filtering panel by itself**. 
Currently we are adding the activate button directly to the filter panel, so under current design I could not bind them together directly. I could make the switch button to align right by itself.

I apply your proposition and I remove the text and leave it only as a title which will be printed on hover. I very like this scenario.

Screenshot 2022-10-17 at 18 00 49

with regular text:

Screenshot 2022-10-17 at 18 05 01

PR #93, opened 2022-08-23
add set filterable

### LINKED ISSUE #70: [Feature Request]: Enable app developers to limit variables available in filter panel
### Feature description

from @wangh107 

Provide a way for app developers to be able to limit filter panel variables choices to pre-specified variables. For example, I have an `ADSL` with 100 variables, as app developer, I wish to have a way to limit the available filtering variables to only allow some variables to be used as filters. i.e., I only want the app users to choose from `SEX`, `AGE`, `RACE`, `PDL1` as possible filters from `ADSL` on the filter panel.

```
### Code of Conduct

### Contribution Guidelines

### Security Policy

```

nikolas-burkoff:
Closes #70 

Also requires https://github.com/insightsengineering/teal/pull/724 - see that PR for example


tests/testthat/test-FilteredDataset.R
```diff
@@ -49,6 +49,54 @@ testthat::test_that("get_keys returns the keys passed to the constructor", {
...
+  expected <- c("Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width", "Species")
+
+  filtered_dataset$set_filterable_varnames(NULL)
+  expect_setequal(filtered_dataset$get_filterable_varnames(), expected)
+
+  filtered_dataset$set_filterable_varnames(character(0))
+  expect_setequal(filtered_dataset$get_filterable_varnames(), expected)
+})
+
+testthat::test_that("setting filterable varnames which include columns that do not existignores these columns", {
```
nikolas-burkoff, line 75:
```suggestion
testthat::test_that("setting filterable varnames which include columns that do not exist ignores these columns", {
```

nikolas-burkoff:
I wonder what should happen if we programmatically try to add a filter to a non filterable variable - at the moment you can
i.e. [here](https://github.com/insightsengineering/teal.slice/blob/main/R/FilteredDataset.R#L607)

R/CDISCFilteredData.R
```diff
@@ -160,7 +160,9 @@ CDISCFilteredData <- R6::R6Class( # nolint
       varnames <- self$get_filtered_dataset(dataname)$get_filterable_varnames()
       parent_dataname <- self$get_parentname(dataname)
       parent_varnames <- if (length(parent_dataname) > 0) {
-        super$get_filterable_varnames(parent_dataname)
+        # cannot call get_filterable_varnames on the parent filtered_dataset in case
+        # some of its variables are set to be not filter-able
```
Polkas, line 164:
```suggestion
        # some of its variables are set to be non-filter-able
```

R/FilteredDataset.R
```diff
@@ -458,6 +478,10 @@ FilteredDataset <- R6::R6Class( # nolint
     metadata = NULL,
     hash = character(0),
 
+    # if this has length > 0 then only varnames in this list
```
Polkas, line 481:
```suggestion
    # if this has length > 0 then only varnames in this vector
```

Polkas:
Because of `vars_include` https://github.com/insightsengineering/teal.slice/blob/80aa3c2c6943fb41f5204edd0ff61a07b123c3c0/R/FilterStates.R#L831

we do not have to adjust filter_panel_api for this case

R/FilteredData.R
```diff
@@ -140,6 +140,18 @@ FilteredData <- R6::R6Class( # nolint
       self$get_filtered_dataset(dataname)$get_filterable_varnames()
     },
 
+    #' @description
+    #' Set the variable names of a given dataset for the filtering
+    #' @param dataname (`character(1)`) name of the dataset
+    #' @param varnames (`character` or `NULL`) The variables which
```
mhallal1, line 146: The variables which?!

nikolas-burkoff: fixed

Polkas (approved):
LGTM

R/CDISCFilteredData.R
```diff
@@ -160,7 +160,9 @@ CDISCFilteredData <- R6::R6Class( # nolint
       varnames <- self$get_filtered_dataset(dataname)$get_filterable_varnames()
       parent_dataname <- self$get_parentname(dataname)
       parent_varnames <- if (length(parent_dataname) > 0) {
-        super$get_filterable_varnames(parent_dataname)
+        # cannot call get_filterable_varnames on the parent filtered_dataset in case
+        # some of its variables are set to be non-filterable
+        self$get_filtered_dataset(parent_dataname)$get_supported_filter_varnames()
```
pawelru, line 165: This is the only place (other than from within the class itself) where we are calling `$get_supported_filter_varnames()`. Can we call S3 method directly and make this method private? Can we remove this method completely and replace all occurrences with S3 method?
Having two separate methods `get_supported_filter_varnames` and `get_filterable_varnames` might be confusing and somewhat complex.

R/FilteredDataset.R
```diff
@@ -314,9 +314,38 @@ FilteredDataset <- R6::R6Class( # nolint
...
+    #' It takes the intersection of the column names
+    #' of the data and `private$.filterable_varnames` if
+    #' `private$.filterable_varnames` has positive length
+    #'
     #' @return (`character` vector) of variable names
     get_filterable_varnames = function() {
-      get_filterable_varnames(self$get_dataset())
+      varnames <- self$get_supported_filter_varnames()
+      if (length(private$.filterable_varnames) > 0) {
+        return(intersect(varnames, private$.filterable_varnames))
```
pawelru, line 325: Have you think about preserving the order of filterable varnames? 

```
> intersect(letters, c("c", "b", "a"))
[1] "a" "b" "c"
> intersect(c("c", "b", "a"), letters)
[1] "c" "b" "a"
```

R/FilteredDataset.R
```diff
@@ -314,9 +314,38 @@ FilteredDataset <- R6::R6Class( # nolint
...
+    get_supported_filter_varnames = function() {
+      get_supported_filter_varnames(self$get_dataset())
+    },
+
+    # setters ------
+    #' @description
+    #' Set the allowed filterable variables
+    #' @param varnames (`character` or `NULL`) The variables which can be filtered
+    #' See `self$get_filterable_varnames` for more details
+    #' @return invisibly this `FilteredDataset`
```
pawelru, line 344: I can see that you do intersect on get and no input validation. I think it's a good decision given the _connectors_. Would be good to document that.

R/FilteredDataset.R
```diff
@@ -314,9 +314,38 @@ FilteredDataset <- R6::R6Class( # nolint
...
+    #' @return (`character` vector) of variable names
+    get_supported_filter_varnames = function() {
+      get_supported_filter_varnames(self$get_dataset())
+    },
+
+    # setters ------
+    #' @description
+    #' Set the allowed filterable variables
+    #' @param varnames (`character` or `NULL`) The variables which can be filtered
+    #' See `self$get_filterable_varnames` for more details
```
pawelru, line 343: would be great to have a link but not sure if it's working for R6 class docs

R/FilteredDataset.R
```diff
@@ -458,6 +487,10 @@ FilteredDataset <- R6::R6Class( # nolint
     metadata = NULL,
     hash = character(0),
 
+    # if this has length > 0 then only varnames in this vector
+    # can be filtered
+    .filterable_varnames = NULL,
```
pawelru, line 492: why `.xyz`? All of other _data_ fields do not have dot in its names so it's somewhat breaking a convention

R/FilteredDataset.R
nikolas-burkoff: So I don't think so  see here r-lib/ roxygen2/issues/955

R/FilteredDataset.R
nikolas-burkoff: changed

R/FilteredDataset.R
nikolas-burkoff: I've switch them round so the filtereable_varnames comes first

R/FilteredDataset.R
pawelru: ok I see - thanks for the reference

pawelru (approved):

R/FilteredDataset.R
pawelru: thanks - that makes more sense to me tbh

Polkas:
I tested it once more after the update, still working correctly.

PR #96, opened 2022-09-09
cran fix

mhallal1:
related to https://github.com/insightsengineering/coredev-tasks/issues/334

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Type: Package
 Package: teal.slice
-Title: Filter module for teal applications
+Title: Filter Module for `teal` Applications
 Version: 0.1.1.9016
 Date: 2022-09-02
 Authors@R: c(


donyunardi:
Trying to see if we can do something about the elapsed time for FilteredData example.
I did look around to see if there's any opportunity to do so but I wasn't able to spot anything.
Maybe someone else can suggest something?

image

image

mhallal1:
> Trying to see if we can do something about the elapsed time for FilteredData example. I did look around to see if there's any opportunity to do so but I wasn't able to spot anything. Maybe someone else can suggest something?
> 
> image
> 
> image

I guess it is system dependent, here are the results on mine:
Screenshot from 2022-09-12 09-35-10

donyunardi:
Thanks for the snapshot. My computer is rather old.

donyunardi (approved):

PR #97, opened 2022-09-13
teal.slice - bs345

Polkas:
linked to https://github.com/insightsengineering/teal/issues/719

- [x] redesign the count bars for the checkbox input in the filter panel
- [x] redesign datetime filterpanel input to use flex
- [x] theme in the fluidPage try to use the teal.bs_theme option if it is there. the new utils function get_teal_bs_version is needed here.

Redesign of the bars for filter panel checkboxes:
Now we use a separate bar per each checkbox, and the bars are a simple div's with a proper background-color. The positions for the elements are done with flex. It is important to test it for variables with only one level values, it is an edge case.
Logical and Choices States are updated.

bs3:
Screenshot 2022-09-19 at 14 06 14
bs4:
Screenshot 2022-09-19 at 14 07 12
bs5:
Screenshot 2022-09-19 at 14 07 55


R/FilterState.R
```diff
@@ -1828,26 +1819,39 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
+            class = "choices_state",
             checkboxGroupInput(
               ns("selection"),
               label = NULL,
-              choices =  private$choices,
               selected = self$get_selected(),
-              width = "100%"
+              width = "100%",
+              choiceNames = labels,
+              choiceValues = as.character(private$choices),
```
Polkas, line 1848:
```suggestion
              choiceNames = labels,
              choiceValues = as.character(private$choices),
              width = "100%"
```

R/FilterState.R
```diff
@@ -1066,19 +1070,35 @@ LogicalFilterState <- R6::R6Class( # nolint
...
+          style = sprintf("width:%s%%", l_freq * 100),
+          span(
+            class = "choices_state_label_text",
+            sprintf(
+              "%s (%s)",
+              private$choices[i],
+              l_count
+            )
+          )
+        )
```
mhallal1, line 1091: this can be probably defined as a function in `utils` as it is called twice.

Polkas: It is not so easy as these are 2 separate classes and we could not add a private method to the base class as it make no sense. A utils function is not a general for me too.

mhallal1 (approved):
LGTM

inst/css/filter-panel.css
```diff
@@ -90,15 +106,29 @@ a.remove_all:hover {
...
-  bottom: -2px;
-  left: 16px;
-  animation: 0.75s ease-out 0s 1 shinyDataFilterEnlargeX,
-    0.5s ease-in 0s 1 shinyDataFilterFadeIn;
-  transform-origin: left;
+  padding: 2px;
+}
+
+.choices_state_label {
+  background-color:lightblue;
```
asbates, line 128: Linking this to insightsengineering/teal/issues/735. If we match the whole app theme to the user set bootstrap theme, we will want to update the color here.

asbates:
Issue noted: in the example app in Getting Started vignette, we use `shiny::dataTableOutput/renderDataTable`. However, these are marked superseded and Shiny recommends using the `DT` package.

After discussing with the team, we will make a note of this in the vignette. But will not use `DT` functions as this will add another dependency. I also added #103 in case we want a different solution in the future.

donyunardi (approved):

PR #98, opened 2022-09-19
Fix rounding for pretty formatting of sliderInput steps

### LINKED ISSUE #41: Better rounding in sliders
Surely we can do better than this @kumamiao, @gogonzo 

image

See maybe https://github.com/rstudio/shiny/issues/1006 and https://stackoverflow.com/questions/42675102/r-shiny-slider-input-round

asbates:
# Pull Request

Fixes #41. 

This PR implements nicer displayed numbers for slider inputs. It uses essentially the [same method as Shiny](https://github.com/rstudio/shiny/blob/c21ba0baca5a3eeaeabbdbcb8e9e83a6bd9a0bb9/R/input-slider.R#L255-L274).

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -1662,11 +1662,34 @@ RangeFilterState <- R6::R6Class( # nolint
     # @return numeric(3) with names min, max, step - relevant for sliderInput
     get_pretty_range_inputs = function(values) {
       v_pretty_range <- pretty(values, n = 100)
+      min <- min(v_pretty_range)
+      max <- max(v_pretty_range)
+
+      step <- private$get_pretty_range_step(min, max, v_pretty_range)
+
       c(
-        min = v_pretty_range[1],
-        max = v_pretty_range[length(v_pretty_range)],
-        step = `if`(private$is_integer, 1L, v_pretty_range[2] - v_pretty_range[1])
+        min = min,
+        max = max,
+        step = step
       )
+    },
+    # @description gets pretty step size for range slider
+    #  adaptation of shiny's method (see shiny/R/input-slider.R function findStepSize)
+    # @param min (numeric) minimum of pretty values
+    # @param max (numeric) maximum of pretty values
+    # @param pretty_range (numeric(n)) vector of pretty values
+    # @return numeric(1) pretty step size for the sliderInput
+    get_pretty_range_step = function(min, max, pretty_range) {
+      range <- max - min
+
+      if (private$is_integer && range > 2) {
+        return(1L)
+      } else {
+        n_steps <- length(pretty_range) - 1
+        return(
+          signif(digits = 10, (max - min) / n_steps)
+        )
+      }
     },
     validate_selection = function(value) {
       if (!is.numeric(value)) {
--- a/.lintr
+++ b/.lintr
@@ -1,4 +1,4 @@
-linters: with_defaults(
+linters: linters_with_defaults(
   line_length_linter = line_length_linter(120),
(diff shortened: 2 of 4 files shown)


donyunardi (approved):

donyunardi:
Maybe we should investigate if we could just pass the ranges directly to `optionalSliderInput` and remove `min`, `max`, and `step` argument to optimize our code.

I'll make a new issue so we can look into this.

Polkas (changes requested):
Please add the NEWS record and add tests
I will check the code in the moment

R/FilterState.R
```diff
@@ -1664,10 +1664,23 @@ RangeFilterState <- R6::R6Class( # nolint
     # @return numeric(3) with names min, max, step - relevant for sliderInput
     get_pretty_range_inputs = function(values) {
       v_pretty_range <- pretty(values, n = 100)
+      min <- min(v_pretty_range)
```
Polkas, line 1667: Think about a separate private method/utils function which could be properly tested and the code more clear. Even linked shiny use such `findStepSize`.

R/FilterState.R
```diff
@@ -1664,10 +1664,23 @@ RangeFilterState <- R6::R6Class( # nolint
...
+      max <- max(v_pretty_range)
+      range <- max - min
+
+      # ensure number of digits is reasonable
+      # essentially the same as shiny's implementation
```
Polkas, line 1672: if you think the solution is "essentially the same", then you have to leave a proper reference.

asbates:
> Maybe we should investigate if we could just pass the ranges directly to `optionalSliderInput` and remove `min`, `max`, and `step` arguments to optimize our code.
> 
> https://github.com/insightsengineering/teal.slice/blob/a449203ab5267fd208fe68d606a89ead1cc9be6a/R/FilterState.R#L1381-L1403
> 
> I'll make a new issue so we can look into this.

This seems like the easiest solution to me. I would also add that we may not need the separate issue. Is there a particular reason why we have `FilterState$get_pretty_range_inputs`? If not, then instead of making the current proposed fix and then coming back and undoing it later (#101), why not forgo the custom method? It would be simpler both in terms of code and administrative overhead to make the 'correct' change now.

mhallal1:
> > Maybe we should investigate if we could just pass the ranges directly to `optionalSliderInput` and remove `min`, `max`, and `step` arguments to optimize our code.
> > https://github.com/insightsengineering/teal.slice/blob/a449203ab5267fd208fe68d606a89ead1cc9be6a/R/FilterState.R#L1381-L1403
> > 
> > I'll make a new issue so we can look into this.
> 
> This seems like the easiest solution to me. I would also add that we may not need the separate issue. Is there a particular reason why we have `FilterState$get_pretty_range_inputs`? If not, then instead of making the current proposed fix and then coming back and undoing it later (#101), why not forgo the custom method? It would be simpler both in terms of code and administrative overhead to make the 'correct' change now.
> 
> @mhallal1 What do you think? Should we pass the values directly to `optionalSliderInput` (therefore `shiny::sliderInput`) and let Shiny handle the formatting?

@asbates I like the proposed solution here, I would respond to Maciej's comments on citing the original work (add the link in the code) and adding some tests if possible.
Regarding the comment above, as we are doing an agile setup, we can fix this task and close it and investigate the need (or not) of `FilterState$get_pretty_range_inputs` in another issue. This is to keep track of what is done in the required time space so we do not take much more time for an issue assigned less time/complexity.

mhallal1 (approved):
LGTM

PR #100, opened 2022-09-21
export init_filtered_states

nikolas-burkoff:
Needed with https://github.com/insightsengineering/teal.modules.helios/pull/32


danielinteractive (approved):

danielinteractive:
Hi @nikolas-burkoff , can we please also export `SEFilterStates`? we need it in the tmhelios PR

nikolas-burkoff:
> Hi @nikolas-burkoff , can we please also export SEFilterStates? we need it in the tmhelios PR

I really don't want to do that as then we have to maintain and deprecate it over the next few increments - as the filter panel is likely to change a lot - also we would need to export all the other filter states as well - I'll let @pawelru decide

danielinteractive:
@nikolas-burkoff hm ok. the problem is that we need that in tmhelios, and we cannot use `:::` since then `R CMD check` will fail.

danielinteractive:
thanks @pawelru for laying out these two options, that description is very clear!
I would definitely recommend to work towards option 1. Especially as we will need other generalizations, and when you want to make teal a long living product, then you will need this extensibility. 
I totally get that this is a bigger change and needs to be well thought through, especially in light of other refactoring going on or planned etc. 
And the good news is that it seems `R CMD check` does not complain about the `:::` - maybe because we are not using `as-cran` or whatever, but because of that we can keep it as is for now but you could consider working towards option 1 in the refactoring?

nikolas-burkoff:
>  but you could consider working towards option 1 in the refactoring?

So it might not be exactly the same API in the future depending on how things are refactored but yes I believe it is essential that people can inherit from and customize the filter states for their own data types. @donyunardi can you capture this in any filter panel refactoring tickets and docs.

nikolas-burkoff:
I'd also argue the the filteredstates for SE and MAE don't actually belong in teal.slice (which should handle base types only) but the general mechanism for creating filtered states is exposed in teal.slice and then these live in say tmhermes. 

Similarly MAEDataset shouldn't live in teal.data, teal.data should expose the way of creating TealDataset objects (or whatever they are replaced by when simplifying stuff) and then the specializations for hermes/helios live in their appropriate place - this then mean @danielinteractive can create/update/change their behaviour without changes to the core of teal - splitting the concerns in a much better way.

Polkas:
"`R CMD check` does not complain about the `:::`"

to solve the problem  we could simply use `utils::getFromNamespace`, e.g. `SEFilterStates_base <- utils::getFromNamespace("SEFilterStates", "teal.slice")`.

Source: my old post https://stackoverflow.com/a/63023801/5442527

pawelru:
Thank you for the discussion. It seems that we are going more into 1 which is also good. If dev team is happy to execute these design changes then let's go ahead with it.

For the dataset classes - I do agree it should be basically the same mechanism and non-general child classes should be defined in the respective package that extends general functionality defined in `teal.slice`. Let's put that in our plans @donyunardi

NEWS.md
```diff
@@ -9,6 +9,7 @@
 * The join keys stored inside `FilteredData` are now `JoinKeys` objects.
 * Updated `get_filter_state` to return a list of active filter states and an attribute with the character form of the filter states.
 * Updated `get_varlabels` method in the `FilterStates` classes, accepts a vector input.
+* Exported `S3` method `init_filtered_states` so that it can be used in other packages.
```
danielinteractive, line 12:
```suggestion
* Exported `S3` generic function `init_filtered_states` so that it can be used in other packages.
```

nikolas-burkoff:
So this should hopefully be green now we don't have capacity to change this any further now but this issue is in the backlog to discuss how the export is handled more generally https://github.com/insightsengineering/teal.slice/issues/102

Issue #109, opened 2022-10-04
White spacing on filter state in report card


nikolas-burkoff:
From https://github.com/insightsengineering/teal.modules.general/issues/469

![image](https://user-images.githubusercontent.com/15201933/193821159-d47bb574-6538-4118-9bcf-4c02a6984935.png)

Can this be changed so either there is much less white space or maybe one entry per line?

donyunardi:
Thanks Nik.
Proposing we move to backlog since this is not critical.
Also think this is a good first issue.

PR #112, opened 2022-10-11
change formatting of selected values

chlebowa:
# Pull Request

solves this issue:
https://github.com/insightsengineering/teal.slice/issues/109

- set `format(justify = "none") to avoid extra whitespace
- values are comma-separated
- lines are wrapped

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal.slice
 Title: Filter Module for `teal` Applications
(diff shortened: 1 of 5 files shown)


Polkas (commented):
lack of the new record in the NEWS file.
You should add a test for wrap functionality which is new.
The nested call could be broken to smaller pieces so will be much more readable.
Please add L when you using integers, like 6L.

Most important we are not adding new functionalities during UAT period.

PR #123, opened 2022-11-08
Filter Panel - global turn on/off button

### LINKED ISSUE #84: Introduce disable filters in FilterPanelApi/FilteredData

Polkas:
closes #84 

Global toggle for Filter Panel, where the hidden state is recovered when turn on.
The PR is taken into account the filter_panel_api which is linked with it.


R/filter_panel_api.R
```diff
@@ -83,8 +91,26 @@ FilterPanelAPI <- R6::R6Class( # nolint
...
+        warning("Filter Panel is turn off so the action can not be applied with api.")
+      }
+      invisible(NULL)
+    },
+    #' @description Toggle the state of the Filter Panel turn on/off button.
+    #' @param id (`character(1)`)\cr
+    #'  `id` to the toggle button.
+    #'
+    #' @return `NULL`
+    filter_panel_toggle = function() {
```
gogonzo, line 107: It could be hard to track down by the developer. Please explain the technical process of the toggle
That a toggle of the state indirectly changes the filter states by switching button. Switch button is observed by the server which disables/enables all filter states. Disabled Filter states are saved so they are restored when filter is again enabled. One can't neither change nor add the filter states if they are disabled

gogonzo (commented):
I like the idea and implementation. I also prefer to remove and restore filters than having two reactives for enable/disable.

R/FilteredData.R
```diff
@@ -586,6 +590,33 @@ FilteredData <- R6::R6Class( # nolint
     restore_state_from_bookmark = function(state, check_data_hash = TRUE) {
       stop("Pure virtual method")
     },
+    #' @description disable the filter panel
```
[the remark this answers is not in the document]
donyunardi:
```suggestion
    #' @description 
    #' Disable the filter panel by adding `disable` attribute to `filter_add_vars` and `filter_active_vars ` id in the UI element.
    #' In addition, it will store the existing filter states in a private field called `cached_states` 
    #' before remove all filter states from the object.
```

R/filter_panel_api.R
donyunardi: I am still a bit confused about this part mainly because I don't see the `filter_panel_toggle()` method is being called anywhere. Also, the roxygen note saying that it takes `id` as parameter but I don't see any argument.

I removed this method, `load_all()`, run teal.gallery exploratory app, and was still able to toggle the filter panel successfully. Do we actually need this method?

R/filter_panel_api.R
```diff
@@ -83,8 +91,26 @@ FilterPanelAPI <- R6::R6Class( # nolint
...
+        private$filtered_data$remove_all_filter_states(datanames = datanames_to_remove)
+      } else {
+        warning("Filter Panel is turn off so the action can not be applied with api.")
+      }
+      invisible(NULL)
+    },
+    #' @description Toggle the state of the Filter Panel turn on/off button.
+    #' @param id (`character(1)`)\cr
+    #'  `id` to the toggle button.
+    #'
```
Polkas, line 105:
```suggestion
    #'
```

R/filter_panel_api.R
Polkas: You are right there is no need for id argument here, +10 points for finding. 
So the whole filter_panel_api seems to be a weight without a proper value added, especially as a exported to end users item.
If we remove toggle method method then we could simplify the FilteredData class as the ui id does not have to be recorded. In my opinion till there is no decision to stop export/support filter_panel_api till then we should not limit its usage/functionality.

R/FilteredData.R
```diff
@@ -586,6 +590,36 @@ FilteredData <- R6::R6Class( # nolint
...
+    #' before removing all filter states from the object.
+    filter_panel_disable = function() {
+      private$filter_turn <- FALSE
+      shinyjs::disable("filter_add_vars")
+      shinyjs::disable("filter_active_vars")
+      private$cached_states <- self$get_filter_state()
+      self$remove_all_filter_states()
+      invisible(NULL)
+    },
+    #' @description enable the filter panel
```
Polkas, line 605:
```suggestion
    #' @description enable the filter panel
        #' Enable the filter panel by adding `enable` class to `filter_add_vars` and `filter_active_vars ` tags in the UI.
    #' In addition, it will restore the filter states from a private field called `cached_states`.
```

R/FilteredData.R
```diff
@@ -586,6 +590,36 @@ FilteredData <- R6::R6Class( # nolint
     restore_state_from_bookmark = function(state, check_data_hash = TRUE) {
       stop("Pure virtual method")
     },
+    #' @description 
+    #' Disable the filter panel by adding `disable` attribute to `filter_add_vars` and `filter_active_vars ` id in the UI element.
+    #' In addition, it will store the existing filter states in a private field called `cached_states` 
+    #' before removing all filter states from the object.
```
Polkas, line 596:
```suggestion
    #' Disable the filter panel by adding `disable` class to `filter_add_vars` and `filter_active_vars ` tags in the UI.
    #' In addition, it will store the existing filter states in a private field called `cached_states` 
    #' before removing all filter states from the object.
```

R/filter_panel_api.R
Polkas: @pawelru we had some discussion about filter panel api and how far we want to support/maintain it. 
It is needed e.g. for teal.reporter and getting filter states in the modules but should it be exported and continuously improved.

nikolas-burkoff: > It is needed e.g. for teal.reporter and getting filter states in the modules

For now it is :)

R/filter_panel_api.R
```diff
@@ -83,8 +91,24 @@ FilterPanelAPI <- R6::R6Class( # nolint
...
-      private$filtered_data$remove_all_filter_states(datanames = datanames_to_remove)
+      if (private$filtered_data$get_filter_turn()) {
+        datanames_to_remove <- if (missing(datanames)) private$filtered_data$datanames() else datanames
+        private$filtered_data$remove_all_filter_states(datanames = datanames_to_remove)
+      } else {
+        warning("Filter Panel is turn off so the action can not be applied with api.")
+      }
+      invisible(NULL)
+    },
+    #' @description Toggle the state of the Filter Panel turn on/off button.
```
donyunardi, line 102:
```suggestion
    #' @description 
    #' Toggle the state of the Filter Panel turn on/off button by running javascript code to click the toggle button with id `filter_turn_onoff`.
    #' This button is observed in `srv_filter_panel` method that executes `filter_panel_enable()` or `filter_panel_disable()` method depending on the toggle state.
```

donyunardi: adding more description as suggested by @gogonzo

R/FilteredData.R
```diff
@@ -904,6 +972,12 @@ FilteredData <- R6::R6Class( # nolint
     # private attributes ----
     filtered_datasets = list(),
 
+    # turn on / off filter panel
+    filter_turn = TRUE,
```
pawelru, line 976: (copied from closed PR as I still think it's relevant)

1. I dislike the "turn" term here. You can _turn_ left/right or on/off. It's a verb whereas you are looking for a noun. I propose _state_ / _status_ / _active_ instead. Please change other places as well including method docs.

2. What's the point of this field? I ask from the class simplicity point of view.
I can see that it's used in the api class - there is a check whether it's active and if yes then you can do set/get/etc. Is that needed? The functionality would be anyway disabled from the GUI perspective (vide: disabling inputs). What I am suggesting here is that the big disable/enable button would _just_ propagate this action to all relevant inputs + save/restore action and that's it. An object would not store its state as this opens a door for unnecessary complexity I believe. I am happy to discuss it further if needed.

_Originally posted by @pawelru in https://github.com/insightsengineering/teal.slice/pull/91#discussion_r961522114_

----

By default `shinyWidgets::switchInput` use the logical(1) for value so this should not be surprising I stay with this type.

`filter_turn` variable is needed only for filter_panel_api, so we could provide a proper warning if sb want to update the state when filter panel is turn off.

I like the Turn Off/On naming, although I see you agree with Nik in this area so please update to the one you are preferring.

_Originally posted by @Polkas in https://github.com/insightsengineering/teal.slice/pull/91#discussion_r961543287_

R/filter_panel_api.R
```diff
@@ -62,7 +62,11 @@ FilterPanelAPI <- R6::R6Class( # nolint
     #'
     #' @return `NULL`
     set_filter_state = function(filter) {
-      private$filtered_data$set_filter_state(filter)
+      if (private$filtered_data$get_filter_turn()) {
+        private$filtered_data$set_filter_state(filter)
+      } else {
+        warning("Filter Panel is turn off so the action can not be applied with api.")
```
pawelru, line 68: "turned off"
Please apply it into other places as well

BUT: please also see my previous comment - maybe it's not needed at all.

R/FilteredData.R
Polkas: 1. I am ok with the active naming convention.
2. I feel you need a little bit of background and support. I am happy to help you here.

R/filter_panel_api.R
Polkas: obsolate

mhallal1 (commented):
Can we enhance the toggle a bit to make it clear what it does?
Screenshot from 2022-11-14 12-12-00

Polkas:
@mhallal1 the help text will appear when you hover the slider. The minimalist design was applied on purpose.

R/FilteredData.R
Polkas: @pawelru you not answer to more important question which is if we want to support filter panel api at all or in what scope.

DESCRIPTION
```diff
@@ -60,4 +60,4 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.1
+RoxygenNote: 7.2.2
```
Polkas, line 63:
```suggestion
RoxygenNote: 7.2.1
```

Polkas:
linked to https://github.com/insightsengineering/teal.slice/issues/127

R/FilteredData.R
```diff
@@ -602,6 +640,22 @@ FilteredData <- R6::R6Class( # nolint
       div(
         id = ns(NULL), # used for hiding / showing
         include_css_files(pattern = "filter-panel"),
+        div(
+          id = ns("switch-button"),
+          class = "flex justify-content-right",
+          div(
+            title = "Active Filtering",
```
Polkas, line 647:
```suggestion
            title = "Enable/Disable filtering",
```

Issue #15, opened 2021-08-05
Enable / disable filters on a variable level (but not removing it)


cicdguy:
temporarily disable filter but not removing it

this would need a special api in filtered data - some disabled state of each of the filter

PRIORITY IS TO HAVE A BUTTON TO DISABLE (NOT REMOVE) ALL FILTERS AT ONCE (on the left of the top garbage button), then try to work on per variable scope.

![image](https://user-images.githubusercontent.com/12943682/128366489-20c9e46f-9187-4ee9-bce6-25ecb4e262cf.png)

Provenance: 
```
Creator: pawelru
```

gogonzo:
@pawelru I guess this is might be still valid

asbates:
@donyunardi @gogonzo Can you check my understanding of this please.

1. Right now we have a disable filter panel switch. When the switch is off (disabled) the filter states are cached and then removed. The change here is to not cache and remove all the filters but disable them.
2. We also want to be able to disable a _single_ filter state while keeping the others enabled.

gogonzo:
> 1. Right now we have a disable filter panel switch. When the switch is off (disabled) the filter states are cached and then removed. The change here is to not cache and remove all the filters but disable them.
> 2. We also want to be able to disable a single filter state while keeping the others enabled.

(2) is what we need. Issue refers to a single card. Disable all filters (1) is implemented at this moment in the way you described and we haven't concluded if we want to change the behaviour - if you see the room for improvement please write your comment. So, please focus on (2)

PR #124, opened 2022-11-08
15 Enable / disable filters on a variable level (but not removing it)

BFalquet:
https://github.com/insightsengineering/teal.slice/issues/15

review together with he corresponding PR in teal.modules.helios https://github.com/insightsengineering/teal.modules.helios/pull/35

- expose init_filtered_dataset to be able to extend the method to other classes (helios::HeliosData).

thank you for the review


nikolas-burkoff:
So I am OK with exporting this function but I think I want an extra note in the roxygen saying that 

"Although this function is exported for use in other packages it may be changed or removed in a future release at which point any code which relies on this exported function will need to be changed"

The reason for this is that during the refactor of the filter panel and upstream of teal it is entirely likely that this will be changed completely and we're not going to be backward compatible or continue to support this (of course helios will at that point need updating with whatever changes we make)

danielinteractive:
thanks @nikolas-burkoff sounds good to us - also good to keep in mind that `teal.modules.helios` is not open source, so nobody externally can learn from our tricks there directly :-)

nikolas-burkoff (approved):

Issue #6, opened 2022-01-25
Improve filter panel UI to make it clear the active filter panel is collapsed


nikolas-burkoff:
I had collapsed the filter panel then did some testing, I then adding a filter and couldn't understand why it wasn't being added 
![image](https://user-images.githubusercontent.com/15201933/150983045-44260daa-cbe0-472c-92a8-9af245918fe1.png)

Can we either make it clearer it's closed or maybe even pop it open if we add a new filter to it?

PR #146, opened 2023-01-04
Collapse filters for a single dataset

### LINKED ISSUE #14: Enable to collapse and expand filters per dataset
As the number of filters grows it becomes increasingly annoying to scroll down to add more filter or change a filter in a dataset. It could be somewhat mitigated by allowing to collapse and expand filters from a dataset.

New buttons like this:
image

nikolas-burkoff:
Closes #14 

Test with example apps including MAE also test different cases make sense (removing all filters with the filters collapsed etc.)

I added a summary of number of filters applied as without that it is hard to tell whether the filters have been collapsed or there are no filters at all - which is particularly confusing if you add a new filter when they are collapsed.

image
image

```
library(teal)

data(miniACC, package = "MultiAssayExperiment")

app <- init(
  data = teal_data(
    dataset("IRIS", iris),
    dataset("MTCARS", mtcars),
    dataset("MAE", miniACC)
  ),
  modules = modules(example_module())
)

runApp(app)
```


R/FilteredDataset.R
```diff
@@ -417,11 +442,20 @@ FilteredDataset <- R6::R6Class( # nolint
           shiny::observeEvent(self$get_filter_state(), {
             if (length(self$get_filter_state()) == 0) {
               shinyjs::hide("remove_filters")
+              shinyjs::hide("collapse")
+              shinyjs::hide("filter_count_ui")
+              shinyjs::show("filters")
```
nikolas-burkoff, line 447: If we move these two out of the `if` (but keep them in the `observeEvent` then this would mean the filters always stop being collapsed when a filter is added or removed - which may be what we want.

R/FilteredDataset.R
```diff
@@ -407,6 +423,15 @@ FilteredDataset <- R6::R6Class( # nolint
           logger::log_trace("FilteredDataset$server initializing, dataname: { deparse1(dataname) }")
           checkmate::assert_string(dataname)
           shiny::setBookmarkExclude("remove_filters")
+
+          output$filter_count <- renderText(
+            sprintf(
+              "%d filter%s applied",
+              length(self$get_filter_state()),
```
nikolas-burkoff, line 430: so in MAE case this isn't quite right:

image

image

BLAZEWIM: Maybe just leave a "filter applied" darkred info? It might be also misleading in when the filter was added, but not actual filtering was done (nothing selected).

R/FilteredDataset.R
BLAZEWIM: +1 I would prefer the filters to be re-collapsed then, the other way it feels like nothing has happened when you add the filter.

nikolas-burkoff: regarding "(but not when Active Filter Variables is collapsed)." - we have https://github.com/insightsengineering/teal.slice/issues/6 it might be something you want to take on?

nikolas-burkoff: changed

R/FilteredDataset.R
nikolas-burkoff: I could change "applied" to "selected" ?

> can we have this info also after Active Filter Variables collapse?

I'm going to defer that to #6

gogonzo, line 430: I think you need a private method `n_active_filter_states` for both `FilteredDataset` and `MAEFilteredDataset`. In the near future it wont be needed as we plan to unnest these filter-states in the API
```suggestion
              private$n_active_filter_states(),
```

gogonzo (commented):
Wundabar 👍 Let's fix the MAE case

R/FilteredDataset.R
nikolas-burkoff: So:

```
# how many filters are currently applied
    n_active_filter_states = function() {
      sum(vapply(names(self$get_filter_state()), function(state_name) {
        if (state_name == "subjects") {
          length(self$get_filter_state()[[state_name]])
        } else {
          length(self$get_filter_state()[[state_name]]$subset)
        }
      }, FUN.VALUE = numeric(1)))
    },
```

seems to work?

nikolas-burkoff: actually I want to move this to FilterStates I think

gogonzo: MAE is ugly itself so I'm not judging this. I'm good with the current state (with n_active_filter_states), but if you'd like to move it to filter_states I'm not against this. But I just wonder how are you going to adjust the icon (which is in the filetredDataset) with label produced by FilterStates

gogonzo (approved):
👍

PR #150, opened 2023-01-05
Fix bs4 error

### LINKED ISSUE #147: Bootstrap 4 RangeFilterState graph is incorrect
From https://github.com/insightsengineering/coredev-tasks/issues/397

```
options(shiny.useragg = FALSE)
options("teal.bs_theme" = bslib::bs_theme(version = "4"))

library(teal)

app <- init(
  data = teal_data(
    dataset("IRIS", iris)
  ),
  modules = modules(example_module()),
  filter = list(IRIS = list(Sepal.Length = c(5, 7)))
)

runApp(app)
```

Note on bs3 and bs5 the graph is correct:

Image

Image

On bs4 it is in the wrong place:
Image

Also gives a warning:

image

**Acceptance Criteria**
- Identify why the warning occurs and fix it.
- If unfixable (bootstrap related issue) then we should probably drop support for bs4?

nikolas-burkoff:
Closes #147 

image

Test with:

```
options(shiny.useragg = TRUE)
options("teal.bs_theme" = bslib::bs_theme(version = "4"))

library(teal)

app <- init(
  data = teal_data(
    dataset("IRIS", iris)
  ),
  modules = modules(example_module()),
  filter = list(IRIS = list(Sepal.Length = c(5, 7)))
)

runApp(app)
```

Also test that it still works on bs3 + bs5 and for different range filters


inst/css/filter-panel.css
```diff
@@ -156,6 +156,8 @@ a.remove_all:hover {
   animation: 0.75s ease-out 0s 1 shinyDataFilterEnlargeY,
     0.5s ease-in 0s 1 shinyDataFilterFadeIn;
   transform-origin: bottom;
+  width: 100%;
```
nikolas-burkoff, line 159: I think we should probably revisit how we do this overlay as it feels very unstable @gogonzo

inst/css/filter-panel.css
```diff
@@ -156,6 +156,8 @@ a.remove_all:hover {
...
+  max_width: 100%;
```
nikolas-burkoff, line 160:
```suggestion
  max-width: 100%;
```

inst/css/filter-panel.css
nikolas-burkoff: For me bs4 image size is now correct in this case:

image

but bs3 is too big:
image

as is bs5:
image


so either we can leave it as the histogram is meant as a rough guide only so it's ok if it's not exactly right, or we should have a rethink

donyunardi: That shift was caused by the margin setup here:

https://github.com/insightsengineering/teal.slice/blob/4de856b16662d8051e44ecd8d0fe24a2090f32f5/inst/css/filter-panel.css#L154

Not sure why we have the left and right margin but if I set to `auto` or `0px`, the div box width is correct.

```css
margin: 10px auto -25px;
```
or 
```css
margin: 10px 0px -25px;
```

Tested against all bs version, the range graphs and div boxes look correct.

donyunardi (commented):
All graphs look consistent in all bs versions.

I added a note about left and right margin setup.
If there's a specific reason for those margins, then we can leave it as is. 

Without the margin, I can see the graph shift 11px to the left.

Here's the video when I turn left and right margins on and off
https://user-images.githubusercontent.com/8597300/210860732-225c38a9-4f18-4d53-85b5-b4aac2b2692c.mov

inst/css/filter-panel.css
nikolas-burkoff: so without margin although the divs line up, the image doesn't line up with the range of the slider:

No margin
image

margin:
image


I'll see if there's a better way of matching them up...

nikolas-burkoff: ok so I fixed this by on these slider (and only these ones) removing the extra bits of the slider beyond the range:
image

This doesn't affect the users ability to choose their range

inst/css/filter-panel.css
```diff
@@ -151,11 +151,19 @@ a.remove_all:hover {
...
+  margin: 10px 0px -25px;
   height: 25px;
   animation: 0.75s ease-out 0s 1 shinyDataFilterEnlargeY,
     0.5s ease-in 0s 1 shinyDataFilterFadeIn;
   transform-origin: bottom;
+  width: 100%;
+  max-width: 100%;
+}
+
+/* used to enusre the rangew histogram exactly matches the slider */
```
nikolas-burkoff, line 163:
```suggestion
/* used to enusre the range histogram exactly matches the slider */
```

inst/css/filter-panel.css
```diff
@@ -151,11 +151,19 @@ a.remove_all:hover {
...
     0.5s ease-in 0s 1 shinyDataFilterFadeIn;
   transform-origin: bottom;
+  width: 100%;
+  max-width: 100%;
+}
+
+/* used to ensure the range histogram exactly matches the slider */
+.filterRangeSlider .irs-grid {
+  width:100% !important;
+  left:0% !important;
```
donyunardi, line 166: I noticed the draggable circle position is not matching with the grid.



How about we update the `.irs-line` instead?

```css
.filterRangeSlider .irs-line {
  margin-left: 10px;
  margin-right: 10px;
}
```

The circle is more centered to the value on the grid:


And our grid still look okay.

R/FilterStateRange.R
```diff
@@ -155,21 +155,25 @@ RangeFilterState <- R6::R6Class( # nolint
...
-              pretty_range_inputs[c("min", "max")]
-            } else {
-              self$get_selected()
-            }
-          }),
-          width = "100%",
-          step = pretty_range_inputs["step"]
+        div(
+          class = "filterRangeSlider",
+          style = "width:100%", # needed for bs4
```
gogonzo, line 160: can it be moved to css file?
```suggestion
```

inst/css/filter-panel.css
nikolas-burkoff: yup that's nicer

R/FilterStateRange.R
nikolas-burkoff: moved

inst/css/filter-panel.css
nikolas-burkoff: ok so I see the problem as well but replacing the .irs-grid with .irs-line makes the graph be too big for me - this is bs3
image

nikolas-burkoff: I wonder if we need our own widget here...

asbates: To make things weider, swapping `.irs-grid` with `.irs-line` looked fine for me on all BS versions.

asbates:
@nikolas-burkoff What do you think about waiting on this until more of the filter panel redesign is finished? I have a feeling even if we get this working now, we may run into the same problem later depending on the html & css of the rest of the filter panel.  

Either way I think the best route might be to adjust the html structure. I don't feel great about changing the slider components (`.irs-grid`, or `.irs-line`), maybe we should let that do it's thing. And then we can change how we overlay the plot. I think css grid might help here because it allows you to overlap elements.

nikolas-burkoff:
> @nikolas-burkoff What do you think about waiting on this until more of the filter panel redesign is finished? I have a feeling even if we get this working now, we may run into the same problem later depending on the html & css of the rest of the filter panel.

Fair enough, I've now made the minimum changes to stop the bs4 error happening and we can defer the alignment until later

donyunardi (approved):
I agree with the suggestion. The update looks good to me.

PR #155, opened 2023-01-11
Simplified reactivity of the FilterState

### LINKED ISSUE #129: [Feature Request]: FilterState to not show shiny inputs by default
part of #48 
### Feature description

Please see the [wireframe](https://drive.google.com/file/d/1jYjYZR6Yqtfv7ZzMfBMa2T5xYIwWmVL7/view?usp=sharing)

1. Shiny inputs should not be visible by default. So called filter-card-summary (ui of the FilterState) should displayed - showing summarized information of the filer-state.
2. User clicks the active button in filter-card-summary to trigger change from filter-card-summary to filter-card-interactive. (Button can be a small gear icon or the card itself).
3. filter-card-interactive is rendered each time it's triggered to show (selection is based on the current state). User can change the values in filter-card-interactive and they are not propagated immediately until one clicks save. Once saved card turns back again to filter-card-summary with a new values. Thanks to updating values of the interactive selection after save only, we don't need a set_selected_reactive anymore (please see below).

--------

Important: Please be aware of how the FilterState$private$selected can be changed. Please see `help(teal.slice:::FilterState)` and go to the modyfying state, you'll see following image:

image

Filter states can be changed in three ways. See example [FilterState$server](https://github.com/insightsengineering/teal.slice/blob/e866d4dd6f7d3d3cec23725b1df8adabc8e506b6/R/FilterState.R#L1140):
- In the interactive session by directly specifying values of selected, keep_na or keep_inf using set_state method (to update all at once), or using set_selected, set_keep_na or set_keep_inf
- In a teal application by changing appropriate inputs
- Using methods set_selected_reactive, set_keep_na_reactive, set_keep_inf_reactive which serves as a programmatic API.

Important improvement would be to remove area in the dashed rectangle which will simplify [this](https://github.com/insightsengineering/teal.slice/blob/e866d4dd6f7d3d3cec23725b1df8adabc8e506b6/R/utils.R#L196) process.

gogonzo:
Partially fixes #129

### Scenarios

Before we dive deeper I'd like to explain how states of the single filters can be changed. "state of the filter" are specifically reactive fields in the FIlterState object (see [FilterState$private$selected and FilterState$private$keep_na](https://github.com/insightsengineering/teal.slice/blob/main/R/FilterState.R#L471)). There are three situations where states can be changed:

1. In the interactive session by directly specifying values of `selected`, `keep_na` or `keep_inf` using `$set_state` method (to update all at once), or using `$set_selected`, `$set_keep_na` or `$set_keep_inf`.
2. In a teal application by changing appropriate inputs. This is done in following way: user select values, FilterState$server_inputs detect that input changed and changes the values using `$set_selected`
3. In a teal application by using filter panel API using `get_filter_state`, `set_filter_state`, `remove_filter_state`. See the example in the bottom of the comment to see how module can interact with the filter-panel. 

### Problematic case

**Case (3) is a reason of this PR and overcomplicated reactivity** because the way how it influences the state of the filter is not direct. Please focus on the proces:
- module interacts with the `filter_panel_api` and changes the state using `set_filter_state`
- state is sent to the FilterState which can't directly change reactive field directly, because values of the state will not match that is selected in the shiny UI - so we can't update values without updating UI. So especially for this situation we introduced extra fields `selected_reactive`, `keep_na_reactive`, `keep_inf_reactive` being equivalent of `selected`, `keep_na`, `keep_inf`. When `*_reactive` fields are changed, then `updateXXXInput` is run, to update UI inputs which then triggers what has been described in (2). Process is depicted on the image below. See the details of [FilterState$server_inputs](https://github.com/insightsengineering/teal.slice/blob/41242cbeb151980ae894c2295087aa9e7f491477/R/FilterStateChoices.R#L270) again to see that this server does following: `selected_reactive changed` > `updateXXXInput` > `input has changed` > `change the state`
 
image

### Solution

Instead of having two copies of reactive fields (`selected`, `keep_na`, `keep_inf`) it's enough to have one copy of the reactive fields and change the related UI element if reactive val has been changed using API. I've added a protection to make sure that values will be updated only if `!setequal(self$get_selected(), input$selection)`

This PR removes all code connected to this two-staged process explained in case (3) - part of the process in the dashed-rectangle on the above image.

### Example

Simple module which prints the filter states using `get_filter_state(filter_panel_api)`. There is also a option to reset filters to the default values by clicking "reset to default" button.

```r
#options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
devtools::load_all("teal")
library(scda)

ADSL$empty <- NA
ADSL$logical1 <- FALSE
ADSL$numeric <- seq_len(nrow(ADSL))
ADSL$logical <- rep(c(TRUE, FALSE), length.out = nrow(ADSL))
ADSL$datetime <- Sys.time() + seq_len(nrow(ADSL))

ADSL$numeric[1:10] <- NA
ADSL$numeric[11:20] <- Inf
ADSL$logical[20:30] <- NA
ADSL$date[30:40] <- NA
ADSL$datetime[40:50] <- NA

funny_module <- function (label = "Filter states") {
  module(
    label = label,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      observeEvent(input$reset, {
        set_filter_state(filter_panel_api, default_filters)
      })

      })
    },
    filters = "all"
  )
}

default_filters <- list(
  ADSL = list(
    numeric = list(selected = c(0, 140), keep_na = TRUE, keep_inf = TRUE),
    logical = list(selected = FALSE),
    date = list(selected = c("2023-12-18", "2050-10-12")),
    datetime = list()
  ),
  ADTTE = list(
    AVAL = list(selected = c(10, 1000), keep_na = TRUE, keep_inf = TRUE)
  )
)

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL),
    cdisc_dataset("ADRS", ADRS)
  ),
  modules = list(funny_module()),
  filter = default_filters
)

runApp(app)
```


R/FilterStateChoices.R
```diff
@@ -301,7 +279,7 @@ ChoicesFilterState <- R6::R6Class( # nolint
               ))
             }
           )
-          private$observe_keep_na(input)
+          private$keep_na_server("keep_na")
```
nikolas-burkoff, line 282:
```suggestion
          private$keep_na_srv("keep_na")
```

nikolas-burkoff (commented):
My main problem seems to be with DateTime - the app in the PR description doesn't work for me as `datetime = list()` filter crashes the app and gives the following error:

image

And if I do remove that but add a datetime filter in the app then the "from" part keeps jumping forward by an hour without any user input:

image

and then a bit later with no user input in the meantime:
image

nikolas-burkoff (commented):
In the app in the PR description - the reset to default button does not reset the keep NA and keep Inf checkboxes even in the case where default_filters explicitly includes their values 

```
numeric = list(selected = c(0, 140), keep_na = TRUE, keep_inf = TRUE),
``` 

Also I'm not sure exactly what is meant to happen here - having a choices filter (with enough choices for pickerinput) - doesn't reset to "all selected"

Edit: Fixed in this PR

nikolas-burkoff (commented):
App crashes if "to" date is earlier than "from" date in DateFilterState (and also in DateTime as well)

image

This also happens on main - I think check_in_range shouldn't give a hard stop and when we move to having a "save" button then we can validate the ui before pressing save to make sure it's valid 


Issue: https://github.com/insightsengineering/teal.slice/issues/158

nikolas-burkoff (approved):
Remaining issue already existed on main and tickets have been put in

PR #156, opened 2023-01-12
display filter count

chlebowa:
Closes [this issue](https://github.com/insightsengineering/teal.slice/issues/6).

Adds element to Active Filter Summary panel that displays number of active filters (if any) when the panel is minimized.


R/FilteredData.R
```diff
@@ -826,15 +827,30 @@ FilteredData <- R6::R6Class( # nolint
             }
           )
 
+          output$teal_filters_count <- renderText({
+            n_filters_active <- reactive({
+              length(unlist(self$get_filter_state(), recursive = FALSE))
+            })
+            # req(n_filters_active > 0L)
+            sprintf("%s filter%s active", n_filters_active(), ifelse(n_filters_active() == 1, "", "s"))
```
nikolas-burkoff, line 835: - So I think it makes more sense to have a private (or public) method n_filters_active() and then here you can just call `private$n_filters_active()`
- I think it's fine to have "0 filters applied" so can remove the commented out code
- Did we use "active" or "applied" in the earlier case? We should be consistent here
- I'm not sure that `length(unlist(self$get_filter_state(), recursive = FALSE))` is the right number in the MAE case. I would much rather do:

```
lapply(self$datanames(), function(dataname) self$get_filtered_dataset(dataname)$n_active_filter_states())
```

nikolas-burkoff (commented):
A comment from me inline and I haven't tested yet but we should make sure everything behaves sensible when we click on these two icons as well (if you're not already)

image

R/FilteredData.R
chlebowa: > * So I think it makes more sense to have a private (or public) method n_active_filter_states() and then here you can just call `private$n_active_filter_states()`

`length(self$get_filter_state())` is used in two places: in a method and in a server function. Shall I replace those with calls to this new method?

> * I think it's fine to have "0 filters applied" so can remove the commented out code

I slightly disagree, I think a clean output looks neater. I suggest we seek a third opinion.

> * Did we use "active" or "applied" in the earlier case? We should be consistent here

👍 

> * I'm not sure that `length(unlist(self$get_filter_state(), recursive = FALSE))` is the right number in the MAE case. I would much rather do:
> 
> ```
> lapply(self$datanames(), function(dataname) self$get_filtered_dataset(dataname)$n_active_filter_states())
> ```

Indeed. Updated.

> we should make sure everything behaves sensible when we click on these two icons as well (if you're not already)

The "Enable filter panel" button works automagically because it clears filters (and stashes them in another private field).

asbates:
When filters are collapsed, can we change the icon from "-" to "+"? I think it would be clearer if "-" means collapse and "+" means expand.

image

chlebowa:
> When filters are collapsed, can we change the icon from "-" to "+"? I think it would be clearer if "-" means collapse and "+" means expand.
> 
> image

Here is a draft method:
Use `shinyjs::runjs` to execute jQuery expressions that change icons or titles. This is wrapped in two helper functions (first draft) that either toggle between two values or change the first into the second.

There is a bug: when panel is minimized and filters are removed, the icon changes byt the title doesn't change even though `expr` is constructed properly (see print in R console).

R/FilteredData.R
```diff
@@ -795,12 +797,23 @@ FilteredData <- R6::R6Class( # nolint
...
             } else {
               shinyjs::show("remove_all_filters")
             }
+            shinyjs::show("filter_active_vars_contents")
+            shinyjs::hide("filters_active_count")
+            toggle_icon(session$ns("minimise_filter_overview"), c("fa-angle-right", "fa-angle-down"), TRUE)
+            toggle_title(session$ns("minimise_filter_overview"), c("Restore panel", "Minimise Panel"), TRUE)
+          })
+
+          shiny::observeEvent(input[["minimise_filter_overview"]], {
```
gogonzo, line 812: please stick to one getElement function 
```suggestion
          shiny::observeEvent(input$minimise_filter_overview, {
```

gogonzo (commented):
> When filters are collapsed, can we change the icon from "-" to "+"? I think it would be clearer if "-" means collapse and "+" means expand.

What @asbates @chlebowa do you think about fa-window-minimize and fa-window-maximize instead?

> There is a bug: when panel is minimized and filters are removed, the icon changes byt the title doesn't change even though expr is constructed properly (see print in R console).

Could you please investigate it further why renderText doesn't react to the change of the count?

R/FilteredData.R
chlebowa: The file was already inconsistent, see lines 873 and 853. Should I change all instances to `$`?

chlebowa:
> > When filters are collapsed, can we change the icon from "-" to "+"? I think it would be clearer if "-" means collapse and "+" means expand.
> 
> What @asbates @chlebowa do you think about fa-window-minimize and fa-window-maximize instead?

Paweł requested chevrons: https://github.com/insightsengineering/teal.slice/issues/6#issuecomment-1082838849

R/helpers.R
```diff
@@ -0,0 +1,52 @@
+toggle_icon <- function(inputId, icons, one_way = FALSE) {
```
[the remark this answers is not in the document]
chlebowa: I will move to utils, but wouldn't we want these functions to show up on the package index?

chlebowa:
> Could you please investigate it further why renderText doesn't react to the change of the count?

@gogonzo please clarify, the count seems accurate to me.

```
library(teal)

mae_d <- dataset("MAE", miniACC, metadata = list(type = "example"))

app <- init(
  data = teal_data(mae_d),
  modules = example_module()
)

```

R/helpers.R
```diff
@@ -0,0 +1,52 @@
...
+          "var button_id = 'a#%1$s';",
+          "var curr = $(button_id).attr('title');",
+          "if (curr == '%2$s') { $(button_id).attr('title', '%3$s');",
+          "} else { $(button_id).attr('title', '%2$s');",
+          "}"
+        ),
+        inputId, titles[1], titles[2]
+      )
+    }
+if (one_way) print(expr)
```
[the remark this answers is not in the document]
chlebowa: This will go when all bugs are removed.

R/FilteredData.R
gogonzo: Yup, please

R/utils.R
```diff
@@ -104,3 +104,130 @@ get_teal_bs_theme <- function() {
     bs_theme
   }
 }
+
+
+
+
+#' Toggle button properties.
```
gogonzo, line 111: please move the functions to the FilteredData-utils.R. Please add tag `@keywords internal` - this does not include function in the doc index (important for private functions)

R/FilteredData.R
```diff
@@ -726,6 +720,11 @@ FilteredData <- R6::R6Class( # nolint
                 }
               )
             )
+          ),
+          div(
+            id = ns("filters_active_count"),
+            style = "display: none;",
+            textOutput(ns("teal_filters_count"))
```
gogonzo, line 727:
```suggestion
          shinyjs::hidden(div(
            id = ns("filters_active_count"),
            textOutput(ns("teal_filters_count"))
```

gogonzo: reason for this is that `shinyjs::show/hide` has it's own way of handling this. At this moment they also append `style = "display:none` but also change the class of the element. If anything in the future change with `shinyjs` our `display: "none"` might affect how `shinyjs` works.

R/FilteredData.R
```diff
@@ -795,12 +794,31 @@ FilteredData <- R6::R6Class( # nolint
...
+            toggle_icon(session$ns("minimise_filter_active"), c("fa-angle-right", "fa-angle-down"))
+            toggle_title(session$ns("minimise_filter_active"), c("Restore panel", "Minimise Panel"))
+          })
+
+          shiny::observeEvent(private$get_filter_count(), {
+            if (private$get_filter_count() == 0) {
               shinyjs::hide("remove_all_filters")
             } else {
               shinyjs::show("remove_all_filters")
             }
```
gogonzo, line 813: Please rather use `condition` argument instead
```suggestion
            shinyjs::hide("remove_all_filters", condition = private$get_filetr_count() == 0)
```

asbates:
> What @asbates @chlebowa do you think about fa-window-minimize and fa-window-maximize instead?

@gogonzo That would work as well. I'm good with any icon as long as it's clear the purpose.

gogonzo (approved):
👍

PR #160, opened 2023-01-13
[don't bother] Filter summary and filter inputs

### LINKED ISSUE #129: [Feature Request]: FilterState to not show shiny inputs by default

gogonzo:
closes #129 

To be discussed with @donyunardi and @lcd2yyz 
Todo:
- ~~better formatting of the summary~~
- removing keep_na and keep_inf if there are no NA or Inf in the column.
- adding optimal delay to the sliderInput 
- ~~adding numericRangeInput as a alternative, additional way of specifying range.~~

#### Solution 
Inputs can be shown by clicking the inputs icon. I prefer to click the icon than clicking the whole summary-view card as the latter is not obvious and complicates the UX. Issue I have with clicking the whole card is that it's not obvious first to click it and how possibly one could go back from the inputs-view back to summary-view. I've tried to implement dropdown-menu with inputs instead. I can present it 

image

One can switch back to the summary by clicking the same element.

image

#### Alternative

I've tried to implement dropdown-menu with inputs instead. Here is the screenshot how it looks. It can be improved, but I dislike this option as it will be difficult to adjust this to the filter-panel size and follow parent DOM.

image

#### Example

```r
#options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
devtools::load_all("teal")
library(scda)

ADSL$empty <- NA
ADSL$logical1 <- FALSE
ADSL$numeric <- seq_len(nrow(ADSL))
ADSL$logical <- rep(c(TRUE, FALSE), length.out = nrow(ADSL))

ADSL$numeric[1:10] <- NA
ADSL$numeric[11:20] <- Inf
ADSL$logical[20:30] <- NA
ADSL$date[30:40] <- NA
ADSL$datetime[40:50] <- NA

funny_module <- function (label = "Filter states") {
  module(
    label = label,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      observeEvent(input$reset, {
        set_filter_state(filter_panel_api, default_filters)
      })

      })
    },
    filters = "all"
  )
}

default_filters <- list(
  ADSL = list(
    numeric = list(selected = c(0, 140), keep_na = TRUE, keep_inf = TRUE),
    logical = list(selected = FALSE),
    date = list(selected = c("2023-12-18", "2050-10-12")),
    datetime = list()
  ),
  ADTTE = list(
    AVAL = list(selected = c(10, 1000), keep_na = TRUE, keep_inf = TRUE)
  )
)

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL),
    cdisc_dataset("ADRS", ADRS)
  ),
  modules = list(funny_module()),
  filter = default_filters
)

runApp(app)

```


nikolas-burkoff:
> More generally, I think styling and interaction would be a lot easier if the UI was not part of the classes. I think this is a long term goal? So maybe we leave it for now. But right now it's really hard to style because different pieces of the HTML live with different classes.

So I wonder if we should do this now - as the data flow is also complicated by having the server functions in the classes - cleaning it now may make it easier to add the new functionality in? Also any other simplifications (@asbates mentions lots of html which is not needed) may also make sense to do now as well?

PR #169, opened 2023-01-23
update init_filtered_data

mhallal1:
related to https://github.com/insightsengineering/teal.data/issues/120

Updated `teal.slice::init_filtered_data` to remove `CDISCTealData` S3 method and to account for `cdisc` status in the `TealData` S3 method.


R/FilteredData-utils.R
```diff
@@ -32,50 +32,32 @@ init_filtered_data.TealData <- function(x, # nolint
                                         code = x$get_code_class(),
                                         cdisc = FALSE,
                                         check = x$get_check()) {
+  cdisc <- if (length(join_keys$get_parents()) > 0) TRUE else FALSE
```
nikolas-burkoff, line 35:
```suggestion
  cdisc <- length(join_keys$get_parents()) > 0 
```

R/FilteredData-utils.R
```diff
@@ -32,50 +32,32 @@ init_filtered_data.TealData <- function(x, # nolint
...
     }
 
-    list(
-      dataset = dataset$get_raw_data(),
-      keys = dataset$get_keys(),
-      metadata = dataset$get_metadata(),
-      label = dataset$get_dataset_label(),
-      parent = parent
-    )
+    if (cdisc) {
```
nikolas-burkoff, line 45: I think you can make this nicer by doing:

sensible_name <- list(dataset = ..., label = ...)
if (cdisc) sensible_name[["parent"]] <- parent

Or - if it's ok for parent to be NULL as opposed to not be in names(list) for non cdisc case:

list(
  dataset = ....,
  ...
  parent = if (cdisc) parent else NULL
)

(or if parent is NULL in non-cdisc case then it's even simpler)

R/FilteredData-utils.R
```diff
@@ -32,50 +32,32 @@ init_filtered_data.TealData <- function(x, # nolint
...
+      list(
+        dataset = dataset$get_raw_data(),
+        keys = dataset$get_keys(),
```
nikolas-burkoff, line 48: I think it's fine to get keys like this for now, in the future we'll get them directly from the join_keys - similarly for parent in line 40 we'll get from join_keys when creating tdata object.

nikolas-burkoff (commented):
A few small comments - when we use tdata to create FilteredData we may want to change this constructor a bit - but that can happen later

I'd also add news item "Updated `init_filtered_data` to take into account the removal of `CDISCTealData` from `teal.data` package"

R/FilteredData-utils.R
mhallal1: updated.

R/FilteredData-utils.R
mhallal1: yes that step can be done with the tdata switch.

R/FilteredData-utils.R
```diff
@@ -32,51 +32,27 @@ init_filtered_data.TealData <- function(x, # nolint
...
       metadata = dataset$get_metadata(),
-      label = dataset$get_dataset_label(),
-      parent = parent
+      label = dataset$get_dataset_label()
     )
+
+    if (cdisc) return_list[["parent"]] <- parent
+    return_list
   })
+
```
chlebowa, line 55: Is it possible to have `return_list` always include the "parent" field, just have it empty and ignored for non-cdisc data?

R/FilteredData-utils.R
```diff
@@ -32,51 +32,27 @@ init_filtered_data.TealData <- function(x, # nolint
...
-                                             join_keys = x$get_join_keys(),
-                                             code = x$get_code_class(),
-                                             cdisc = TRUE,
-                                             check = x$get_check()) {
-  data_objects <- lapply(x$get_datanames(), function(dataname) {
-    dataset <- x$get_dataset(dataname)
-
-    # CDISCTealData can contain TealDataset and CDISCTealDataset objects
-    # the former do not have a get_parent() call
     parent <- if (inherits(dataset, "CDISCTealDataset")) {
```
nikolas-burkoff, line 39: shouldn't we be getting the parents from the join_keys here?

mhallal1: updated.

R/FilteredData-utils.R
mhallal1: this will require doing some changes in the downstream methods.

nikolas-burkoff (approved):

PR #171, opened 2023-01-25
remove superfluous methods

chlebowa:
Rather than have `state_list_initialize` methods in `FilterStates` classes create lists of `reactiveVal`s, use `initialize` to create those lists in `private$state_list`.

Removed some unit tests for the `FilterStates` abstract class are superfluous.

Also cleaned up some leftover "queue" nomenclature in comments and log messages.


R/FilterStatesDF.R
```diff
@@ -396,6 +397,8 @@ DFFilterStates <- R6::R6Class( # nolint
       )
     }
   ),
+
+  # private membres ----
```
nikolas-burkoff, line 401:
```suggestion
  # private members ----
```

nikolas-burkoff (commented):
Just confirming that the tests removed for the abstract class do have equivalent tests in the appropriate childclass test file (I assume DFFilterStates) - if so then this can be merged after small typo fix.

Please also merge into feature branch

chlebowa:
Unfortunately some of the tests do not have child class equivalents.

Also, [this](https://github.com/insightsengineering/coredev-tasks/issues/430).

tests/testthat/test-init_filter_state.R
```diff
@@ -20,34 +20,28 @@ test_that("'extract_type' must be specified with 'input_dataname'", {
 })
```
chlebowa, line 20: This code raises an error because `ADSL` is not defined. Any suggestions would be welcome.

R/FilterState.R
```diff
@@ -75,13 +75,13 @@ FilterState <- R6::R6Class( # nolint
...
       checkmate::assert_character(extract_type, max.len = 1, any.missing = FALSE)
-      stopifnot(
-        length(extract_type) == 0 ||
-          length(extract_type) == 1 && !is.null(input_dataname)
-      )
-      stopifnot(extract_type %in% c("list", "matrix"))
+      if (length(extract_type) == 1)
+        checkmate::assert_choice(extract_type, choices = c("list", "matrix"))
+      if (length(extract_type) == 1 && is.null(input_dataname))
+        stop("if extract_type is specified, input_dataname must also be specified")
```
gogonzo, line 83: We want to check also if it is a scalar.
```suggestion
      if (length(extract_type) != 0)
        checkmate::assert_choice(extract_type, choices = c("list", "matrix"))
      if (length(extract_type) == 1 && is.null(input_dataname))
        stop("if extract_type is specified, input_dataname must also be specified")
```

tests/testthat/test-DFFilterStates.R
```diff
@@ -99,38 +99,38 @@ testthat::test_that("DFFilterStates$set_filter_state updates filter state which
...
-      varlabels = character(0),
-      keys = character(0)
-    )
-    fs <- list(
-      c(5.1, 6.4),
-      Species = c("setosa", "versicolor")
-    )
-    testthat::expect_error(dffs$set_filter_state(state = fs, data = iris))
-  }
+                    code = {
```
gogonzo, line 102: Please make this indent smaller
```suggestion
testthat::test_that("DFFilterStates$set_filter_state throws error when using an unnamed list", {
```

gogonzo (approved):
👍 Thanks. When merged please merge `main->refactor_filter_panel@main`

R/FilterState.R
chlebowa: The case of `extract_type` being longer than 1 is forbidden and covered by `checkmate::assert_character(extract_type, max.len = 1, any.missing = FALSE)` above.

tests/testthat/test-DFFilterStates.R
```diff
@@ -251,21 +253,27 @@ testthat::test_that(
...
+      dffs$ui_add_filter_state("id", data.frame(A = numeric(0))),
+      div("no samples available")
+    )
   }
 )
 
 # Format
 testthat::test_that("$format() is a method of DFFilterStates", {
-  testthat::expect_error(isolate(
+  testthat::expect_no_error(isolate(
```
nikolas-burkoff, line 269: it's not especially important but I would try to isolate as little as possible (so we get failed tests if we add reactivity to somewhere new - but it's not worth changing this at this time

DF <- DFFilterStates$new(...)
testthat::expect_no_error(isolate(DF$format

nikolas-burkoff (approved):
Cool - lets merge into feature branch as well

tests/testthat/test-DFFilterStates.R
chlebowa: `isolate` IS necessary here because there is a `reactiveVal` in `private$state_list` (formerly `queue`) now:
```
DFFilterStates$new(
      input_dataname = "test",
      output_dataname = "test",
      datalabel = character(0),
      varlabels = "test",
      keys = "test"
    )$format()
```
```
Error in `.getReactiveEnvironment()$currentContext()`:
! Operation not allowed without an active reactive context.
• You tried to do something that can only be done from inside a reactive consumer.
Run `rlang::last_error()` to see where the error occurred.
```

Issue #175, opened 2023-01-27
Replace `input_dataname` and `output_dataname` in FIlterStates class with `dataname`


gogonzo:
In the `FIlterStates` class we use `input_dataname` and `output_dataname` fields which are the same. We have two fields as historiacaly we used to assign filtered dataset into a new binding `ADSL_FILTERED <- ADSL |> filter(...)`.
Please try to change class of stored `private$dataname` into character (currently it's of type `name` or `call`).

PR #176, opened 2023-01-27
175 replace input output@main

BLAZEWIM:
# Pull Request

Fixes #[175](https://github.com/insightsengineering/teal.slice/issues/175)

also varname

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,8 +22,6 @@
 
 * Calculation of step in slider for `RangeFilterState` now uses `checkmate::test_integerish` instead of `is.integer`.
 * Updated `init_filtered_data` to take into account the removal of `CDISCTealData` from `teal.data` package.
-* Parameters `input_dataname` and `output_dataname` of `FilterStates` are now combine into one character parameter `dataname`.
-* Parameter `varname` of `FilterState` is not a character.
 
 # teal.slice 0.2.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,8 @@
 
 * Calculation of step in slider for `RangeFilterState` now uses `checkmate::test_integerish` instead of `is.integer`.
 * Updated `init_filtered_data` to take into account the removal of `CDISCTealData` from `teal.data` package.
+* Parameters `input_dataname` and `output_dataname` of `FilterStates` are now combine into one character parameter `dataname`.
+* Parameter `varname` of `FilterState` is not a character.
 
 # teal.slice 0.2.0
 
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -47,19 +47,19 @@ FilterState <- R6::R6Class( # nolint
     #' Initialize a `FilterState` object
     #' @param x (`vector`)\cr
     #'   values of the variable used in filter
-    #' @param varname (`character`, `name`)\cr
+    #' @param varname (`character`)\cr
     #'   name of the variable
     #' @param varlabel (`character(1)`)\cr
     #'   label of the variable (optional).
-    #' @param input_dataname (`name` or `call`)\cr
+    #' @param dataname (`character(1)`)\cr
     #'   name of dataset where `x` is taken from. Must be specified if `extract_type` argument
     #'   is not empty.
     #' @param extract_type (`character(0)`, `character(1)`)\cr
     #' whether condition calls should be prefixed by dataname. Possible values:
     #' \itemize{
     #' \item{`character(0)` (default)}{ `varname` in the condition call will not be prefixed}
-    #' \item{`"list"`}{ `varname` in the condition call will be returned as `<input_dataname>$<varname>`}
-    #' \item{`"matrix"`}{ `varname` in the condition call will be returned as `<input_dataname>[, <varname>]`}
+    #' \item{`"list"`}{ `varname` in the condition call will be returned as `<dataname>$<varname>`}
+    #' \item{`"matrix"`}{ `varname` in the condition call will be returned as `<dataname>[, <varname>]`}
     #' }
     #'
     #' @return self invisibly
@@ -67,27 +67,21 @@ FilterState <- R6::R6Class( # nolint
     initialize = function(x,
                           varname,
                           varlabel = character(0),
-                          input_dataname = NULL,
+                          dataname = NULL,
                           extract_type = character(0)) {
-      checkmate::assert(
-        checkmate::check_class(varname, "name"),
-        checkmate::check_class(varname, "call"),
-        checkmate::check_string(varname)
-      )
+      checkmate::assert_string(varname)
       checkmate::assert_character(varlabel, max.len = 1, any.missing = FALSE)
-      checkmate::assert_multi_class(input_dataname, c("name", "call"), null.ok = TRUE)
+      checkmate::assert_string(dataname, null.ok = TRUE)
       checkmate::assert_character(extract_type, max.len = 1, any.missing = FALSE)
-      if (length(extract_type) == 1)
+      if (length(extract_type) == 1) {
         checkmate::assert_choice(extract_type, choices = c("list", "matrix"))
-      if (length(extract_type) == 1 && is.null(input_dataname))
-        stop("if extract_type is specified, input_dataname must also be specified")
-
-      private$input_dataname <- input_dataname
-      private$varname <- if (is.character(varname)) {
-        as.name(varname)
-      } else {
-        varname
       }
+      if (length(extract_type) == 1 && is.null(dataname)) {
+        stop("if extract_type is specified, dataname must also be specified")
+      }
+
+      private$dataname <- dataname
+      private$varname <- varname
       private$varlabel <- if (identical(varlabel, as.character(varname))) {
         # to not display duplicated label
         character(0)
@@ -98,13 +92,12 @@ FilterState <- R6::R6Class( # nolint
       private$selected <- reactiveVal(NULL)
       private$na_count <- sum(is.na(x))
       private$keep_na <- reactiveVal(FALSE)
-
       logger::log_trace(
         sprintf(
           "Instantiated %s with variable %s, dataname: %s",
           class(self)[1],
-          deparse1(varname),
-          deparse1(private$input_dataname)
+          private$varname,
+          private$dataname
         )
       )
       invisible(self)
@@ -140,7 +133,7 @@ FilterState <- R6::R6Class( # nolint
       values <- paste(format(self$get_selected(), nsmall = 3L, justify = "none"), collapse = ", ")
       paste(c(
         strwrap(
-          sprintf("Filtering on: %s", self$get_varname(deparse = TRUE)),
+          sprintf("Filtering on: %s", self$get_varname()),
           width = wrap_width,
           indent = indent
         ),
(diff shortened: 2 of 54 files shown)


chlebowa:
Was #175 to add support for a character string in `dataname` or to change to only handle a character string?

BLAZEWIM:
> Was #175 to add support for a character string in `dataname` or to change to only handle a character string?

Second one, change it.

nikolas-burkoff:
Please merge into filterpanel_refactor@main once this has been merged 👍

R/FilterStates.R
```diff
@@ -161,25 +140,16 @@ FilterStates <- R6::R6Class( # nolint
         # below code translates to call by the names of filter_items
         rhs <- call_with_colon(
           self$get_fun(),
-          private$input_dataname,
+          `if`(is.call(private$dataname), private$dataname, as.name(private$dataname)),
```
nikolas-burkoff, line 143: if dataname is only a character then we don't need this? And on line 148

R/FilterStates.R
```diff
@@ -118,11 +97,11 @@ FilterStates <- R6::R6Class( # nolint
...
     #'
     #' Builds \emph{subset expression} from condition calls stored in `FilterState`
-    #' objects selection. The `lhs` of the expression is `private$output_dataname`.
-    #' The `rhs` is a call to `self$get_fun()` with `private$input_dataname`
+    #' objects selection. The `lhs` of the expression is `private$dataname`.
+    #' The `rhs` is a call to `self$get_fun()` with `private$dataname`
     #' as argument and a list of condition calls from `FilterState` objects
     #' stored in `private$state_list`.
-    #' If `input_dataname` is the same as `output_dataname` and no filters are applied,
+    #' If no filters are applied,
```
nikolas-burkoff, line 104: @gogonzo maybe another issue but on line 109 do you think we could do the following:

get_call = function(call_dataname = character(0)) {
  if (length(call_dataname) == 0) {
    call_dataname <- private$dataname
  } 
  checkmate::assert_character(call_dataname)  
  // replace private$dataname with call_dataname in the rest of this function
},

nikolas-burkoff (commented):
A couple of small comments

R/FilterStates.R
nikolas-burkoff: I see this is likley to be too hard

tests/testthat/test-init_filter_state.R
```diff
@@ -8,29 +8,29 @@ testthat::test_that("init_filter_state accepts a character vector of length 0 or
...
   testthat::expect_error(
-    init_filter_state(7, varname = "test", input_dataname = quote(test)),
+    init_filter_state(7, varname = "test", dataname = quote(test)),
     NA
   )
   testthat::expect_error(
-    init_filter_state(7, varname = "test", input_dataname = call("test")),
+    init_filter_state(7, varname = "test", dataname = call("test")),
     NA
   )
```
chlebowa, line 35: remove

tests/testthat/test-init_filter_state.R
```diff
@@ -42,16 +42,16 @@ testthat::test_that("init_filter_state accepts character as extract_type", {
...
 
   testthat::expect_error(
-    init_filter_state(7, varname = "test", input_dataname = as.name("test"), extract_type = "list"),
+    init_filter_state(7, varname = "test", dataname = as.name("test"), extract_type = "list"),
```
chlebowa, line 45: `dataname = "test"`

tests/testthat/test-init_filter_state.R
```diff
@@ -42,16 +42,16 @@ testthat::test_that("init_filter_state accepts character as extract_type", {
...
     NA
   )
   testthat::expect_error(
-    init_filter_state(7, varname = "test", input_dataname = as.name("test"), extract_type = "matrix"),
+    init_filter_state(7, varname = "test", dataname = as.name("test"), extract_type = "matrix"),
```
chlebowa, line 49: `dataname = "test"`

chlebowa (changes requested):
Some assertions change with #171, wait for that to be merged (~2 hrs) before you apply.

R/FilterStates.R
BLAZEWIM: Reason is MAE - in this case it still can be a call. BUT - maybe it would be better to reverse it and create a custom get_call for FilterStatesMAE.R.

BLAZEWIM:
> Some assertions change with #171, wait for that to be merged (~2 hrs) before you apply.

Ok, will wait, let me know.

R/FilterStates.R
nikolas-burkoff: at the very least add a comment here so when I look next time I see why it's like this :)

R/FilterState.R
```diff
@@ -51,38 +51,38 @@ FilterState <- R6::R6Class( # nolint
     #'   name of the variable
     #' @param varlabel (`character(1)`)\cr
     #'   label of the variable (optional).
-    #' @param input_dataname (`name` or `call`)\cr
+    #' @param dataname (`name` or `call`)\cr
```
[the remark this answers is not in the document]
donyunardi: I assumed you were referring to making `dataname` argument as character class.

If we do this, we need to make sure that `$get_dataname()` is behaving properly.

https://github.com/insightsengineering/teal.slice/blob/aab1255ec1cbb6442f7d3cef1271eed873aec0a7/R/FilterState.R#L180-L186

If we treat `dataname` as character, then it will have backslash `\` when calling `$get_dataname`
```r
# default behavior
r$> filter_state$get_dataname()
[1] "data"

r$> deparse1(as.name("data"))
[1] "data"

# if we treat dataname as character
r$> deparse1("data")
[1] "\"data\""


```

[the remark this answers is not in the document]
chlebowa: Yes, the `deparse` in the method should be removed.

Note the slash is just an esacape character because you get a string containing quote marks.

R/FilterStates.R
gogonzo: Why is custom call needed? If we have `dataname = character(1)` in all filter-states then we can still use abstract `get_call`, right? If you need to turn `character` to call please just use `str2lang()`

`language` is used only in `get_call` methods and we needed to convert dataname into character using `deparse1(private$input_dataname)` everywhere else. Point of the issue is to simplify the classes - let's keep this field as a `character` only and use `character` only to initialize the class.

Just a brief look how faster is turning from string to language. Comparison is still on the level of microseconds, but still shows that keeping `private$dataname = character(1)` and converting it to the `language` when using `get_call` is not a big deal.

```r
call <- quote(MAE[[xxx]])
microbenchmark::microbenchmark(
  str2lang("MAE[[xxx]]"),
  deparse1(call)
)
#                   expr    min     lq     mean median     uq    max neval
# str2lang("MAE[[xxx]]")  1.401  1.601  1.83901  1.701  1.801 12.201   100
#         deparse1(call) 15.500 15.901 17.13894 16.201 16.751 74.501   100
```

R/FilterStates-utils.R
```diff
@@ -7,13 +6,10 @@
 #' @param data (`data.frame`, `MultiAssayExperiment`, `SummarizedExperiment`, `matrix`)\cr
 #'   the R object which `subset` function is applied on.
 #'
-#' @param input_dataname (`character(1)` or `name` or `call`)\cr
-#'   name of the data used on lhs of the expression
+#' @param dataname (`character(1)` or `name` or `call`)\cr
```
chlebowa, line 9:
```suggestion
#' @param dataname (`character(1)`)\cr
```

tests/testthat/test-FilterState.R
```diff
@@ -58,18 +41,12 @@ testthat::test_that("get_varlabel returns a string passed to the constructor", {
 
 testthat::test_that("get_varname(deparse = FALSE) returns a name if varname passed to the constructor is a string", {
```
chlebowa, line 42: Amend test descriptions.

chlebowa (approved):
Almost there :+1:

R/FilterStateDate.R
```diff
@@ -66,7 +66,7 @@ DateFilterState <- R6::R6Class( # nolint
       sprintf(
         "%sFiltering on: %s\n%1$s  Selected range: %s - %s\n%1$s  Include missing values: %s",
         format("", width = indent),
-        self$get_varname(deparse = TRUE),
+        self$get_varname(),
```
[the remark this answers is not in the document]
BLAZEWIM: It can, the condition in it was a leftover, I will check whether we can get rid of it

R/FilterState.R
```diff
@@ -170,18 +163,15 @@ FilterState <- R6::R6Class( # nolint
...
     #'
-    get_dataname = function(deparse = TRUE) {
-      if (isTRUE(deparse)) {
-        deparse1(private$input_dataname)
+    get_dataname = function() {
+      if (!is.null(private$dataname)) {
+        private$dataname
       } else {
-        private$input_dataname
+        "NULL"
```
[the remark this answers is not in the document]
BLAZEWIM: Thinking about it again, I will change it to empty string, this "NULL" looks strange. This was the default behavior previously (see modified tests_init_filter_states.R, `init_filter_state provides default values for varlabel, dataname` ), but not neccessary one I suppose.

R/FilterState.R
```diff
@@ -470,16 +460,19 @@ FilterState <- R6::R6Class( # nolint
...
       } else if (isTRUE(private$extract_type == "matrix")) {
-        call_extract_matrix(dataname = private$input_dataname, column = as.character(private$varname))
+        call_extract_matrix(
+          dataname = private$dataname,
+          column = as.character(private$varname)
+        )
       } else {
-        private$varname
+        str2lang(private$varname)
       }
```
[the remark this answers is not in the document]
BLAZEWIM: yes, but get_varname_prefixed() returns a call, so this is the place it stops being one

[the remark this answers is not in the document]
chlebowa: Oh, I see!

R/FilteredDataset.R
```diff
@@ -463,15 +464,15 @@ FilteredDataset <- R6::R6Class( # nolint
...
             lapply(
               self$get_filter_states(),
               function(x) x$state_list_empty()
             )
-            logger::log_trace("FilteredDataset$server@1 removed filters, dataname: { deparse1(dataname) }")
+            logger::log_trace("FilteredDataset$server@1 removed filters, dataname: { dataname }")
           })
 
-          logger::log_trace("FilteredDataset$initialized, dataname: { deparse1(dataname) }")
+          logger::log_trace("FilteredDataset$initialized, dataname: { (dataname }")
```
gogonzo, line 475:
```suggestion
          logger::log_trace("FilteredDataset$initialized, dataname: { dataname }")
```

R/FilterStateDate.R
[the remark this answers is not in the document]
BLAZEWIM: I decided to keep the function, but change the use among the class to be consistent with the use of get_dataname and private$dataname

NEWS.md
```diff
@@ -22,6 +22,8 @@
 
 * Calculation of step in slider for `RangeFilterState` now uses `checkmate::test_integerish` instead of `is.integer`.
 * Updated `init_filtered_data` to take into account the removal of `CDISCTealData` from `teal.data` package.
+* Parameters `input_dataname` and `output_dataname` of `FilterStates` are now combine into one character parameter `dataname`.
+* Parameter `varname` of `FilterState` is not a character.
```
gogonzo, line 26: both classes and their constructor wrappers (init_filter_state, init_filter_states) are internal. They shouldn't be mentioned in the changelog

gogonzo: There are also other NEWS about theese clases - we will remove them later

gogonzo (approved):
👍 Tested with the apps everything works. No occurrences of the input_ and output_dataname in the code. Calls are generated properly, logs looks good.

PR #180, opened 2023-01-31
change default argument reactive(NULL)

### LINKED ISSUE #178: Change default `x_reactive` argument of `FilterState` to `NULL`
The `x_reactive` argument to FilterState classes controls the "blue" histogram of selected values in RangeFilterState and the left count ie. (100 in the count "100/200") in Choices, Logical, keep_NA and keep_Inf 

image

FilterStates passes this information to FilterState - after https://github.com/insightsengineering/teal.slice/pull/172 the default of `x_reactive` is `reactive(x)` but it should be NULL and if x_reactive is NULL then do not show the filtered count i.e. don't show (NULL/200) just "200" and don't show the blue filtered histogram.

When doing this optimize the code in the initialize of all FilterState classes (ideally we don't want lots of if, else statements throughout the class)

Also explore if the default  data_reactive argument of FilterState should be changed

nikolas-burkoff:
Closes #178 

- I prefer reactive(NULL) than NULL
- We are limited in what we do in initialize as if we check the value of x_reactive then we need isolate around the constructor
- I'm open to suggestions regarding making this nicer
- Added the css into the filterstates ui as well as we need it for the RangeFilterState histograms

Default of FilterStates is also reactive(NULL)


nikolas-burkoff:
I guess if we change from reactive(NULL) to just NULL we can do something smarter in initialize as then the check if(is.null(x_reactive)) doesn't need an isolate...

gogonzo:
I these `get_choices_label` functions are too complicated. I think we can simplify and tidy up the way how we obtain counts. In the init of `choicesFilterState` there is `private$histogram_data` which isn't used at all. Having `private$choices` where value is a factor level and name is a count is also not elegant. I think we can introduce 
```
private$choices # factor levels
private$choices_count # unfiltered count
private$f_choices_count # reactive
private$calculate_choices_count(x) 
```

That is just a proposition, I think it would be nice if we tidy up these `initialize` methods and how we preserve counts.

R/FilterStatesMAE.R
```diff
@@ -155,7 +155,9 @@ MAEFilterStates <- R6::R6Class( # nolint
         } else {
           fstate <- init_filter_state(
             x = SummarizedExperiment::colData(data)[[varname]],
-            x_reactive = reactive(SummarizedExperiment::colData(data_reactive())[[varname]]),
+            x_reactive = reactive( if (!is.null(data_reactive()))
```
nikolas-burkoff, line 158: Need this as cannot call colData on NULL

gogonzo (approved):
👍 Thanks

Issue #179, opened 2023-01-31
Add Helper apps into roxygen


nikolas-burkoff:
I think these 2 apps should be included in the roxygen documentation as they make it clearer how FilterState and FilterStates work:

- https://github.com/insightsengineering/teal.slice/pull/172#issuecomment-1406468757
- https://github.com/insightsengineering/teal.slice/pull/172#issuecomment-1406636803

donyunardi:
Acceptance Criteria:
* Add helper document in two classes FilterState and FilterStates
* Consolidate if related example exist elsewhere in the package

gogonzo:
I think each class (eg. `RangeFilterState`, `ChoicesFIlterState`) should have own example, simple single `fs$ui()` and `fs$server()` call, wrapped in the simple module outputing `fs$get_call` and `fs$get_selected`. We should not create app which consolidate multiple `FilterState` modules. Same regarding `FilterStates` class. One class - one example.

If examples are written for init_filter_states and init_filter_state, examples should be done for each class - the examples will get consolidated automatically as they sharing the same documentation page.

@donyunardi @nikolas-burkoff @chlebowa

donyunardi:
Thanks @gogonzo 

I understand your perspective, and I agree that although it may require considerable effort, implementing the change that you suggested would significantly enhance the user experience with our documentation by eliminating the need for them to navigate through various files to locate examples.

If there are no objections, let's proceed with this updated acceptance criteria.

PR #165, opened 2023-01-19
Refactor of the filter-panel

gogonzo:
Feature branch for recent changes related with [refactor of the filter-panel](https://github.com/insightsengineering/teal.slice/milestone/1)

Todo before merge:
- [x] Make range graph align with slider (or put issue in backlog)
- [ ] Add formal tests and check coverage
- [ ] Refactor?
- [x] https://github.com/insightsengineering/teal.slice/issues/337
- [x] https://github.com/insightsengineering/teal.slice/issues/129 
- [ ] Update pictures in teal, vignettes etc.
- [x] https://github.com/insightsengineering/teal.slice/issues/166
- [x] https://github.com/insightsengineering/teal.slice/issues/167
- [x] #173 
- [x] #174 
- [x] #187 
- [ ] #298 
- [ ] Run and pass R CMD Check 
 -----

## testing module
```

funny_module <- function (label = "Filter states", datanames = "all") {
  checkmate::assert_string(label)
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderPrint({
        logger::log_trace("rendering text1")
        filter_panel_api %>% get_filter_state()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}
```

## MAE example
```r
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
#options("teal.bs_theme" = bslib::bs_theme(version = 5))
#options(shiny.trace = TRUE)
devtools::load_all("teal.slice")
devtools::load_all("teal")
library(scda)
library(SummarizedExperiment)
utils::data(miniACC, package = "MultiAssayExperiment")

default_filters <- filter_settings(
  filter_expr(id = "F", title = "females", dataname = "MAE", datalabel = "subjects", expr = 'MAE$gender == "female"'),
  filter_var(dataname = "MAE", datalabel = "subjects", varname = "race"),
[...]

Issue #135, opened 2022-11-28
Enable multiple filter_panel module calls


gogonzo:
Why?

We are moving towards module-specific-filter-panel which means that controlling module filters through single UI element is either impossible or too complicated to address.

First obstacle is that we have a single filter panel UI and we are inserting new filters (see observer in [FilterStates$srv_active](https://github.com/insightsengineering/teal.slice/blob/17a8fc65deb4c8cfe948948a6e993b1af2ef496d/R/FilterStatesDF.R#L224)). If you look at this line in [FilterStates$insert_filter_state_ui](https://github.com/insightsengineering/teal.slice/blob/17a8fc65deb4c8cfe948948a6e993b1af2ef496d/R/FilterStates.R#L380) you will understand that location of the HTML element where we insert new card is fixed in the class (so we have a single filter ui module for all teal_modules)

- It should be possible to call filter-panel module multiple times
- `FilteredData` should be still a single-global object
- `private$cards_container_id` should be removed

donyunardi:
Acceptance Criteria:
* When the module changes, the filter panel should show filters related to the module.
* This will be in teal.slice for now and available for programmer to activate if they want.
* Teal still only call the filter panel once.

gogonzo:
closed by #300

PR #181, opened 2023-01-31
show/hide filter state inputs

asbates:
This is the first part of #129, allowing filter states to show either the inputs or a summary, with only one input visible at a time. This is done through a redesign of the filter cards and their container. The cards now have a header that displays the summary information. When a header is clicked, the input expands. Clicking a different header closes the first and expands the second. The 2 step process is so that the functionality can be introduced so other issues are not held up.

For part 2, I am planning on improving the styling. More specifically, truncating the filter state value when it is long. e.g. categorical with 20 levels selected needs to be shortened somehow. I would also like to make some general style updates to make it look nicer.


R/FilterStateChoices.R
```diff
@@ -136,6 +136,98 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
+              class = "filter-card-remove"
+            )
+          )
+        ),
+        tags$div(
+          class = "filter-card-body",
+          private$ui_inputs(ns("inputs"))
+        )
+        )
+      )
```
gogonzo, line 230: I think we can keep ui in the FilterState as before but header could be delegated to the child-classes - in the FIlterState it could look like this:
```r
ui = div(
  class = "filter-card-header",
  private$ui_header(ns("header")),
  actionLink(ns("remove"),...)
)
div(
  private$ui_inputs(ns("inputs"))
)
```

R/FilterStateRange.R
```diff
@@ -214,6 +214,117 @@ RangeFilterState <- R6::R6Class( # nolint
...
+      
+      tags$li(
+        id = id,
+        tags$div(
+        class = "filter-card",
+        tags$div(
+          class = "filter-card-header",
+          uiOutput(ns("header_name_value"), inline = TRUE),
+          uiOutput(ns("header_keep_na"), inline = TRUE),
+          uiOutput(ns("header_keep_inf"), inline = TRUE),
```
gogonzo, line 309: Maybe instead of: 
`<varname>: min-max NA: TRUE Inf: TRUE`
`<varname>: min-max NA: FALSE Inf: FALSE`
 
following could be better:
`<varname>: min-max, NA, Inf`
`<varname>: min-max`

inst/css/filter-panel.css
```diff
@@ -218,3 +218,78 @@ a.remove_all:hover {
...
+
+.fa-check {
+  color: var(--bs-success, #3c763d);
+}
+
+.fa-xmark {
+  color: var(--bs-danger, #a94442);
+}
+
+.filter-card-body {
```
gogonzo, line 288: image

We need some margin

gogonzo (commented):
Looks good. This is a good way to solve out problems. Let's discuss the way how new FS are going to be added/inserted.

I think that to complete this issue we will have to use renderUI instead of insertUI/removeUI https://github.com/insightsengineering/teal.slice/issues/135
I'm happy to discuss this on the call

nikolas-burkoff:
On the refactor branch I had to add `include_css_files(pattern = "filter-panel"),` inside the UI of the FilterState class [see here](https://github.com/insightsengineering/teal.slice/pull/165/commits/4d7d029ff034943debafb0e1a0d7e39e9789f052) to make sure that the apps referenced [here](https://github.com/insightsengineering/teal.slice/issues/179) have appropriate css. 

Sorry that has caused a conflict here. 

We should make sure these apps work with the new js/css

nikolas-burkoff:
Looking nice!

I've been playing with examples https://github.com/insightsengineering/teal.slice/pull/165

Might be worth thinking of: 
- rounding of numbers
- I like an arrow for a range so "10 -> 1000" rather than just (10 - 1000)
- We can open multiple accordions at once:
image
I don't know if that's a problem or not @gogonzo 

For the header which is too long - I quite like the idea of "VARNAME x, y,... (+23)" or something like that

R/FilterStateChoices.R
donyunardi: I also see reusable code in server method. Maybe we can move that to FilterState class too?

asbates:
> For now, I think we should focus on the the default bootstrap (for shiny 1.6.0, it's bootstrap 4) that is being used by shiny/teal. This way we won't be distracted as we continue on with our progress. Once we're done with the code foundation of the upgraded filter panel, we can revisit this and see how teal reacts with other bs version.

So FYI as of 1.6.0 Shiny _supports_ bootstrap 4. But it still uses 3 as the default.

donyunardi:
> > For now, I think we should focus on the the default bootstrap (for shiny 1.6.0, it's bootstrap 4) that is being used by shiny/teal. This way we won't be distracted as we continue on with our progress. Once we're done with the code foundation of the upgraded filter panel, we can revisit this and see how teal reacts with other bs version.
> 
> So FYI as of 1.6.0 Shiny _supports_ bootstrap 4. But it still uses 3 as the default.

Ah okay, thanks. Yes, let's focus on 3 then.

R/FilterStateChoices.R
asbates: Definitely some repeated code. I think I misunderstood a conversation the other day.

R/FilterStateRange.R
asbates: Not sure what you mean here. Could you expand on this?

lcd2yyz:
> Unless there's a way to make this more pleasant, I am leaning toward not include the selection values in the title. Since this is user facing, tagging @lcd2yyz for opinion.

I also feel including selected values in the variable names make it way to crowded. We don't need to go so extreme to collapse to one-liner, but being able to show the 1) variable label and 2) selected values or ranges, will be informative and important.

So this is what I had in mind initially, it should also allow bootstrap theme to come through
image

I'm less concerned about showing NA/Inf in the collapsed mode - if visually doesn't fit (in a visually pleasing way), we don't need to show. But I think we might need to design different displays for different variable class. For example
* Boolean var: show any selected
* Categorical var:
  * Binary: show any selected (same as boolean)
  * Multiple: show selected if only 1 category checked, show as "Multiple" if >1 category selected
* Numerical var: show selected range, auto formatted to scientific notation if significant digit is >4
* Date var: show selected date range, formatted to ymd only (do not show time part)

R/FilterStateRange.R
gogonzo: `NA: TRUE` and `Inf: TRUE` takes some space so I think about alternative display. Instead of having `NA: TRUE` as a separate "span" we can include it in the choices, so we could have:
- `SEX: F, M, NA` when NA is TRUE
- `AGE: 18-64, Inf` when Inf is TRUE
- `AGE: 18-64` when Inf is FALSE

gogonzo:
We gathered with the team and decided to move on with the issue (on the feature branch). So that:
- summary card will contain header + one-line-summary
- header will be always the same for all filter states- Title + label + remove button
- summary will always be a one-liner which will be produced by the XyzFilterState class. So each XyzFilterState class will be responsible to produce the summary. 

So far we concluded to have following summary:
- RangeFIlterState: `$min-$max`, `if $na then "NA" else NULL`, `if $inf then "Inf" else NULL`
- DateFilterState: `$min-$max`, `if $na then "NA" else NULL`
- ChoicesFilterState: `$n-levels selected`
- DatetimeFilterState: `$min-$max`, `if $na then "NA" else NULL`
- LogicalFilterState: `$value`, `if $na then "NA" else NULL`

If we stick to the convention that summary is always a one liner produced by the classes than it will be very easy to change particular summary module for a single data type without changing the design for all of them.

R/FilterState.R
```diff
@@ -399,35 +402,34 @@ FilterState <- R6::R6Class( # nolint
...
-                class = "filter_panel_varname"
-              ),
-              if (checkmate::test_character(self$get_varlabel(), min.len = 1) &&
-                tolower(self$get_varname()) != tolower(self$get_varlabel())) {
-                tags$span(self$get_varlabel(), class = "filter_panel_varlabel")
-              }
-            )
+              class = "filter-card-title",
+              tags$span(self$get_varname(deparse = TRUE)),
+              if (length(self$get_varlabel())) {
```
donyunardi, line 415: The varname and varlabel show twice when it holds the same value:

image

In the previous state of the code, we have the if statement comparing `$get_varname` and `$get_varlabel`, which will work, but I don't think this is the root cause.

private$varlabel is initialized with condition to not hold the same value with varname:
https://github.com/insightsengineering/teal.slice/blob/aab1255ec1cbb6442f7d3cef1271eed873aec0a7/R/FilterState.R#L91-L96

However, when I `browser()` this, I see the following:
image

varlabel is the result of `get_varlabels()` method from FilteredDataset class:
https://github.com/insightsengineering/teal.slice/blob/aab1255ec1cbb6442f7d3cef1271eed873aec0a7/R/FilteredDatasetDefault.R#L41-L50

I think we need to review this method:
https://github.com/insightsengineering/teal.slice/blob/aab1255ec1cbb6442f7d3cef1271eed873aec0a7/R/FilteredDataset.R#L292-L301

R/FilterStateRange.R
```diff
@@ -515,6 +514,27 @@ RangeFilterState <- R6::R6Class( # nolint
         )
         invisible(NULL)
       })
+    },
+    ui_summary = function(id) {
+      ns <- NS(id)
+      uiOutput(ns("summary"), class = "filter-card-summary")
+    },
+    server_summary = function(id) {
```
donyunardi, line 522: We should hide the summary when the filter card is clicked:
image

inst/css/filter-panel.css
```diff
@@ -218,3 +218,90 @@ a.remove_all:hover {
...
+.filter-card-header {
+  display: grid;
+  grid-template-rows: 1fr 1fr;
+  color: var(--bs-body-color, #333333);
+  border: 1px solid var(--bs-gray-200, rgba(0, 0, 0, 0.03));
+  padding: 5px 10px;
+  border-radius: 0.25rem;
+}
+
+.filter-card-header:hover {
```
donyunardi, line 268:
```suggestion
.filter-card:hover {
```

On hover, I can see the header changes.

Screen Shot 2023-02-06 at 2 04 19 PM

What do you think about applying this to the filter-card instead?
image

donyunardi (changes requested):
Thanks @asbates , provided my feedback.

R/FilterState.R
donyunardi: Maybe something changes in `formatters::var_labels`?

lcd2yyz:
> * RangeFIlterState: `$min-$max`, `if $na then "NA" else NULL`, `if $inf then "Inf" else NULL`
> * DateFilterState: `$min-$max`, `if $na then "NA" else NULL`
> * ChoicesFilterState: `$n-levels selected`
> * DatetimeFilterState: `$min-$max`, `if $na then "NA" else NULL`
> * LogicalFilterState: `$value`, `if $na then "NA" else NULL`

Few minor questions on the defined filter states
1) any particular reason for differentiate date vs datetime as two filter states? 
2) Assuming `RangeFilterState` is to be used for numerical variables, I have some concern on the naming itself, because ultimately dates are also expected as a range with min-max. Should we just call it "numerical" or are there some additional considerations?

R/FilterState.R
asbates: Wow, thanks for the details! I'll look into this.

inst/css/filter-panel.css
asbates: Good call. I'll look into this. Your screenshots look odd to me though because the hover color is supposed to apply to the title and summary. The idea is to focus on a card when it is collapsed. May be better on the whole card.

R/FilterStateRange.R
asbates: Will do.

R/FilterStates.R
```diff
@@ -367,12 +367,13 @@ FilterStates <- R6::R6Class( # nolint
     #'
     ui = function(id) {
       ns <- NS(id)
-      private$cards_container_id <- ns("cards")
+      private$cards_container_id <- ns("filter-card-container")
```
gogonzo, line 370: `filter-card-container` uses `-` which is a shiny namespace separator. I'm not sure whether it has consequences or risks.

nikolas-burkoff: Yeah I'd change that to "_"

lcd2yyz:
@gogonzo Thanks for the explanation, makes sense. Let's keep your proposed design then.

inst/js/filter-panel.js
```diff
@@ -0,0 +1,87 @@
...
+// there are 2 cases for the clicked card:
+// 1. it is currently inactive
+// 2. it is currently active
+
+// if 1., we need to set the active card, which was not clicked, to inactive
+//  this is handled in the first 'if' block
+
+// if 2., we need to set the active card, which was clicked, to inactive
+//  this is handled in the second 'if' block
+toggleFilterCard = function (el, containerId) {
```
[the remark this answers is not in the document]
asbates: For border, I've made them darker so the visual separation is more obvious.

inst/css/filter-panel.css
asbates: I looked at coloring the whole card on hover. This made it hard to separate the header from the body on an expanded card. I've added hover color for the body on an expanded card. This visually separates the header and body better. Please let me know what you think about this.

inst/js/filter-panel.js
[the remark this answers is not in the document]
asbates: For overlap, that's going to take some digging. More generally the cards don't look right on Chrome so I'm guessing there's something going on with browser support.

Firefox:
image


Chrome:

image

asbates:
@nikolas-burkoff 

- We can open multiple accordions at once

I was under the impression the one-at-a-time only applies within an accordion not across accordions. @gogonzo what do you think?

- When you first open a range filter I don't see the histogram until I start moving the slider

I suspect this may be due to different browsers. Hoping this goes away with addressing other issues but keeping track of it just in case.

asbates:
@lcd2yyz Regarding "auto formatted to scientific notation if significant digit is >4".

I've added rounding via `sprintf("%.4g", value)`. R describes the "g" flag as "Double precision value, in %e or %E format if the exponent is less than -4 or greater than or equal to the precision, and %f format otherwise. (The precision (default 6) specifies the number of significant digits here, whereas in %f, %e, it is the number of digits after the decimal point.)".

I think this is close to what you're going for but not sure if it's exact. What do you think?

inst/js/filter-panel.js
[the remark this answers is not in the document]
gogonzo: Yes, there is something going on with the browser, but we can't use this excuse towards end-users. I'm fine with having this look on the feature branch "just to move on" but we have to be sure that we are able to fix it quickly. @asbates how do you think, is it doable to fix the appearance for chrome relatively fast?

gogonzo:
> I was under the impression the one-at-a-time only applies within an accordion not across accordions. @gogonzo what do you think?

Across accordions would be better, but still we can live with accordion within a dataset. But on the other side, if we are running custom js and css just to create accordions within each FilterState, why not to use something ready-to-use like [shiny.semantic::accordion](https://rdrr.io/github/Appsilon/shiny.semantic/man/accordion.html) or the bootstrap accordion:
@asbates let's meet tomorrow so we can both discuss this.

gogonzo:
> I've added rounding via sprintf("%.4g", value)

I think for a decimal point we can use options that app developer can change it if has a specific data which require better precision. We shouldn't fix it to one value

https://github.com/insightsengineering/teal.slice/blob/8a09ad5aaeab460d921218cd4f80c9b0c1af1a1d/R/zzz.R#L3

inst/js/filter-panel.js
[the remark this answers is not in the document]
asbates: What would be quick to you? I'm thinking the end of this week if I had to say. What's the alternative?

asbates:
> > I was under the impression the one-at-a-time only applies within an accordion not across accordions. @gogonzo what do you think?
> 
> Across accordions would be better, but still we can live with accordion within a dataset. But on the other side, if we are running custom js and css just to create accordions within each FilterState, why not to use something ready-to-use like [shiny.semantic::accordion](https://rdrr.io/github/Appsilon/shiny.semantic/man/accordion.html) or the bootstrap accordion: @asbates let's meet tomorrow so we can both discuss this.

We can absolutely use bootstrap instead. To be honest I didn't know that existed in bootstrap. The only thing is we would have to account for versions 3, 4, 5. I don't mind using another package like `shiny.semantic` either. Personally I would try to avoid adding that as a dependency just for this but I'll go with it if that's what's decided. I agree a discussion on implementation would be good.

inst/js/filter-panel.js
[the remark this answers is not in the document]
gogonzo: It's looking good now!
image

nikolas-burkoff:
> Across accordions would be better

So the issue with having the possibility of each accordion having something open concerns cdisc data where filtering on ADSL triggers changes to the counts in other datasets - which brings in more reactivity (and possibly something similar within summarized experiments). However, if there's at most one filter card open on each dataset that will limit the reactivity substantially so it's probably ok

nikolas-burkoff:
> I think for a decimal point we can use options so the app developer can change it if has a specific data which require better precision. We shouldn't fix it to one value

And we should remember when possible to add the option into [this](https://github.com/insightsengineering/teal/blob/main/vignettes/teal-options.Rmd) vignette.

inst/js/filter-panel.js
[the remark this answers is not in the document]
gogonzo: Almost there - pickerInput space inside of the filter-card-body is trimmed to the size of the card

image

asbates:
> > > I've added rounding via sprintf("%.4g", value)
> > 
> > 
> > I think for a decimal point we can use options so the app developer can change it if has a specific data which require better precision. We shouldn't fix it to one value
> > https://github.com/insightsengineering/teal.slice/blob/8a09ad5aaeab460d921218cd4f80c9b0c1af1a1d/R/zzz.R#L3
> 
> I feel this example is slightly different - this is controlling a conscious switch from one variable class to another, that is not the inherent property of the variable. The decimal rounding is somewhat "native" to the property of numerical variables itself. While we can definitely add it as optional argument to teal start-up setting and make it as flexible as possible, on the other hand, too many choices/options isn't necessary a good thing. We've just talked about the amount of complexity we already have in the product, then following the path of simplification, I'm okay with implementing a fixed value for now, then come back in the future if there are actual user request for such flexibility.
> 
> In conclusion, what @asbates has proposed looks good to me.

@lcd2yyz I think this is just an example of using an option, not the option that would apply to this case. I will continue with a fixed value. I opened #190 to keep track of using an option.

Issue #137, opened 2022-12-08
[Feature Request]: Implement the mechanism to flag filter-states with some id


gogonzo:
### Feature description

**Purpose:**
We need a mechanism which allow to group filter-state(s) to obtain:
- module specific code `get_data(<module id>)`, `get_code(<module id>)`.
- module specific data 
- display only module specific filter cards
- get_filter_state should return only module specific filters

![image](https://user-images.githubusercontent.com/6959016/226813787-a3836029-d800-49a2-a70d-b2986802f58c.png)

We need a mechanism within `FilterStates` which controls which filter-state belong to which module. Name "module" in the filter-panel shouldn't be used, we need a name of this attribute to be more general (flag, target, you name it).

What I know for sure is that this attribute should be a reactive value, because in the app user will be able to enable/disable this filter in any module.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

donyunardi:
Acceptance Criteria:
* Each filter state should have a relation to a teal module

gogonzo:
Added blocked label because we need to decide how to set the information, through:
- `teal::init(filters = filter_settings(...))`
- set_filter_state
- how to set the information in the running application

gogonzo:
@lcd2yyz @donyunardi @asbates @chlebowa 

Please imagine two modules groups `modules1` and `module2` and some set of filters which are mapped with modules. There are (at least) two ways to achieve module-specific.

1. Two group of modules have a separate set of filters and changing one filter in one group doesn't influence filters in another group. API could look like below. This could be implementent solely in `teal` as we would create n-FilteredData for n-groups (n independent filter panels for n-groups)

```r
teal::init(
  modules(
    # modules,
    filters = filter_settings(
      filter_var(...),
      filter_var(...)
    )
  ),
  modules(
    #modules,
    filter_settings(
      filter_var(...),
      filter_var(...)
    )
  )
)
```

2. Two groups of modules share some filters. Then single `FilterState` need to be linked with several modules. Some first thoughts below:

```r
filter_settings(
  filter_var(..., id = "filter1"),
  filter_var(..., id = "filter2"),
  filter_var(..., id = "filter3"),
  filter_var(..., id = "filter4"),
)


teal::init(
  modules = list(
[...]

PR #186, opened 2023-02-03
Multiple filter panels

### LINKED ISSUE #135: Enable multiple filter_panel module calls

gogonzo:
See issue for details why - closes #135 

feature-branch          |  this branch
:-------------------------:|:-------------------------:
local-filter-panel-insertUI drawio  |  local-filter-panel-insertUI drawio (1)

On feature-branch (or main) filter-cards are displayed in following process:
1. New `FilterState` is added to the `state_list` 
2. `state_list` is reactive and it's observed in `FilterStates$srv_active`, which checks if something new has been added. 
3. Added `FilterState` object are passed to `FilterStates$add_filter_state_ui` 
4. In the `FilterStates$add_filter_state_ui` `srv` from added `FilterState` is called 
5. In the `FIlterStates$add_filter_state_ui` `ui` from added `FilterState` is inserted into div which `id = private$cards_container_id` 

On this branch
1. New `FilterState` is added to the `state_list` - same as 1 above
2. `state_list` is reactive and it's observed in `FilterStates$srv_active`, which checks if something new has been added. It's similar to (2)
3. `srv` from added `FilterState` is being called. Look carefully, there is `added_state_name()` observed which means that `srv` is called only once (srv can't be called multiple times).
4. `ui` of all `FilterState` objects kept in `state_list` are rendered

What change gives:
- filter-panel modules can be called multiple times in the app, so each `teal_module` can have own display of the filter-panel. There will be still single `FilteredData` object keeping all the filters, but filter cards can be displayed independently when switching tabs.
- Able to remove `insert_filter_state_ui`, `remove_filter_state_id`, `map_vars_to_html_ids` which were needed to call new `FilterState$srv` and insert `FilterState$ui` into `private@cards_container_id`
- removed `private@cards_container_id` as multiple filter panels can't have a single and fixed location of the UI (there will be multiple)

There is an alternative solution possible:
- `FilterStates$srv_active` can receive a `reactive` (passed from outside `FilteredDdata`) indicating what group of filters to display. `ui` which doesn't fit some criteria will be hidden and others will be shown. Need [this](https://github.com/insightsengineering/teal.slice/issues/137) to understand more.

Needs work: 
❌ I broke accordion which allows to open multiple cards in the same time

--------

Check the example below and `teal.gallery` apps.

```r
devtools::load_all("teal")
library(shiny)

funniest_module <- function(label = "Filter states", datanames = "all") {
  module(
    label = label,
    filters = datanames,
    ui = function(id, datasets) {
      ns <- NS(id)
      fluidPage(
        column(4, datasets$ui_filter_panel(ns("fp2"))),
        column(8, verbatimTextOutput(ns("filters")))
      )
    },
    server = function(id, datasets) {
      moduleServer(id, function(input, output, session) {
        datasets$srv_filter_panel("fp2")
        output$filters <- renderText(datasets$get_filter_state() |> yaml::as.yaml())
      })
    }
  )
}

app <- init(
  data = list(iris = iris, mtcars = mtcars),
  modules = list(funniest_module(), funniest_module("module 2", datanames = "iris")),
  filter = list(iris = list(Sepal.Length = list()))
)

runApp(app)

```


nikolas-burkoff (commented):
- cdsic example here https://github.com/insightsengineering/teal.slice/pull/165#issue-1548941880 doesn't work for me - it gets stuck in some loop 
image


- The disable button doesn't work:

image

(Not sure what should happen when there are two filter panels but it doesn't work just on one)

- The toggles work apart from the ActiveFilter Summary one:
image

R/FilterStateChoices.R
```diff
@@ -292,12 +292,15 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
                     inputId = "selection",
+                    choiceNames = private$get_choice_labels(),
+                    choiceValues = as.character(private$choices),
                     selected = self$get_selected()
                   )
                 } else {
                   teal.widgets::updateOptionalSelectInput(
                     session = session,
                     inputId = "selection",
+                    choices = stats::setNames(private$choices, private$get_choice_labels()),
```
nikolas-burkoff, line 303: what's the reason for removing choices from the updateCheckbox and adding it to updateOptionalSelect ? Similarly in the logical case below

R/FilterStatesSE.R
```diff
@@ -337,6 +252,103 @@ SEFilterStates <- R6::R6Class( # nolint
...
+
+    #' @description
+    #' Shiny server module.
+    #'
+    #' @param id (`character(1)`)\cr
+    #'   shiny module instance id
+    #'
+    #' @return `moduleServer` function which returns `NULL`
+    #'
+    server = function(id) {
```
nikolas-burkoff, line 285: Is the reason we need this only here  because self$state_list_get(1) works for other filterstate classes but there are two queues here so we need a different server? I'm a little lost tbh

R/FilterStateChoices.R
gogonzo: Hmm I didn't removed any choices. I just add everything available to updateXxx (choices, labels, selected) whenever update is run. I did it as I noticed [here](https://github.com/insightsengineering/teal.slice/pull/180/commits/d4f5acc18e20c344867265137fc7244c45787a8d) when I added 
`label = private$get_inf_label(),` to updateCheckboxInput call and it prevented  one reactive cycle thanks to this (label didn't need to recompute).

gogonzo: Hmm I just realized that we have two observers to updateXyz - I will try to combine them into one

nikolas-burkoff: ah yeah, my mistake

gogonzo:
I'll wait with this PR until we have stable refactor branch.

nikolas-burkoff:
Before this goes in we should make sure that disabling/enabling the filter panel(s) works sensibly - at  the moment it's not quite right

asbates:
> * Clicking on the cards does not open them (even when there's only one filter panel - @asbates thoughts)?

Not sure. Categorical cards are working for me but not numeric which is odd. I did notice this branch is a little behind so maybe that will resolve it.

The collapse/expand functionality is mostly class based so it shouldn't matter if there are 1 or 10 filter panels. I think the only place an id is used is the clickable area (card header) needs to reference the expandable area (card body). So there may need to be some adjustments to make sure these ids are unique.



image

Issue #184, opened 2023-02-01
Range slider moves


nikolas-burkoff:
If you move the left edge of the slider away from 0 and then back to 0 it will jump out to 0.01 which is confusing.
Please test with this app but also other apps and values of the range to make sure this hasn't caused unforeseen issues.

Probably best to start this one from the refactor branch even though the issue is on main as well

```
library(shiny)
library(teal.slice)
library(shinyjs)

x <- c(0.0091, 0.01 + runif(50, max= 0.99))

my_filter_state <- teal.slice:::init_filter_state(
  x = x,
  varname = "VARNAME",
  varlabel = "A label"
)

#####

ui <- fluidPage(
  useShinyjs(),
  column(3,
         div(  # only need a div so that we can hide the ui when the close button is pressed
           id = "x",
           my_filter_state$ui("fs"),
           h4("Condition (i.e. call)"), # show the condition generated by this FilterState
           textOutput("condition"),
           h4("Formatted state"), # show the human readable state
           textOutput("formatted"),
           h4("Unformatted state"), # show the raw state
           textOutput("unformatted"),
         )
  )
)

server <- function(input, output, session) {

  output$condition <- renderPrint(my_filter_state$get_call())
  output$formatted <- renderText(my_filter_state$format())
  output$unformatted <- renderPrint(my_filter_state$get_state()) # also get_selected(), get_keep_na

  signal_to_remove <- my_filter_state$server("fs")  # filterstate$server returns a trigger when the close button is pressed
  observeEvent(signal_to_remove(), shinyjs::hide("x")) # and shinyjs removes the ui
}

shinyApp(ui, server)
```

nikolas-burkoff:
So I don't know why we actually need to do this:
![image](https://user-images.githubusercontent.com/15201933/216101302-a6a86518-9218-441b-8615-392e1954eafc.png)

Surely `dplyr::filter(df, x > 0.0991)` and `dplyr::filter(df, x > 0)` are the same if there is no data between 0 and 0.0991 and users used the slider to select x = 0.

If we call `self$set_selected(input$selection)` instead of what's done (and remove the warnings and errors from `remove_out_of_bound_values` and `check_in_range` which reset the choices /crash the app respectively then it all seems to be fine

chlebowa:
[...]

Issue #133, opened 2022-11-24
Add option to zoom in sliderInput 


gogonzo:
Our sliderInput needs improvement for distributions where values are concentrated in a narrow range. It's hard to figure out where to set slider range. I propose option to zoom in the histogram by the user. If user selects a range on the distribution, little popup will show up with the distribution of the zoomed range. 

![image](https://user-images.githubusercontent.com/6959016/203741468-1a1da6ed-c556-4667-9634-a8bfe9ba0ebe.png)

lcd2yyz:
It would also be very convenient for users to be able to enter the desired value range as text input.
Something like the numerical filter slider in RStudio SDE could be pretty nice.
It allows user to drag and drop to select on a range, or enter the value directly in the text box below.
![image](https://user-images.githubusercontent.com/78242757/204711512-241f938a-c5a3-45b4-8d9f-19d018c9c6d9.png)

nikolas-burkoff:
Yes - I was thinking that being able to add the range _might_ be all we actually need @lcd2yyz  (i.e. behave like the slider in teal.goshawk - if you want to see the zoomed in distribution you can always use the variable browser to see what you are filtering) - it would then be much simpler both to use and to implement.

![image](https://user-images.githubusercontent.com/15201933/210327540-57c03409-af54-4d9d-b9fa-846b30ce5645.png)
![image](https://user-images.githubusercontent.com/15201933/210327578-d1277ee7-d2c1-48c2-80f0-a090bd6179eb.png)

and if this was a widget in teal.widgets it could be reused in teal.goshawk...

One thing we need to be careful with is at the moment it's not possible to create invalid filters in the ui of the filter panel, but by adding input boxes it will be possible - so we should add shinyvalidate (?) and handle the case where the filter panel filtering would produce an error carefully (e.g. not apply filter until it's valid, have a save filter button which is disabled until it's valid)

gogonzo:
@nikolas-burkoff Looks like we agree on this https://github.com/insightsengineering/teal.goshawk/issues/205

nikolas-burkoff:
See https://github.com/insightsengineering/teal.slice/pull/144

nikolas-burkoff:
Note 
![image](https://user-images.githubusercontent.com/15201933/211526896-7ce4c1a8-16cc-42c3-86ac-7df133eb7ab9.png)

nikolas-burkoff:
shinyvalidate plays nicely with the various inputs:


[...]

PR #189, opened 2023-02-03
slider snaps from edge in `RangedFilterState`

chlebowa:
Solves #184.

Range of numeric variable is prettified first, before anything is created.
Slider is created based on prettified range (as previously).
Initial selection is set based on prettified range (previously on actual range).
Condition call is based on prettified range (previously on actual range).
Formatted output is based on actual range and values (as previously).

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -315,8 +315,8 @@ FilterState <- R6::R6Class( # nolint
         )
       )
       value <- private$cast_and_validate(value)
-      value <- private$remove_out_of_bound_values(value)
-      private$validate_selection(value)
+      # value <- private$remove_out_of_bound_values(value)
+      # private$validate_selection(value)
       private$selected(value)
       logger::log_trace(sprintf(
         "%s$set_selected selection of variable %s set, dataname: %s",
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -2,7 +2,9 @@
 #' @docType class
 #'
 #'
-#' @title Abstract class to encapsulate filter states
+#' @title FilterState Abstract Class
+#'
+#' @description Abstract class to encapsulate filter states
 #'
 #' @details
 #' This class is responsible for managing single filter item within
@@ -298,7 +300,7 @@ FilterState <- R6::R6Class( # nolint
     #' @param value (`vector`)\cr
     #'   value(s) that come from filter selection; values are set in the
     #'   module server after a selection is made in the app interface;
-    #'   values are stored in `private$selected`n which is reactive;
+    #'   values are stored in `private$selected` which is reactive;
     #'   value types have to be the same as `private$choices`
     #'
     #' @return NULL invisibly
@@ -399,11 +401,9 @@ FilterState <- R6::R6Class( # nolint
             width = 10,
             class = "no-left-right-padding",
             tags$div(
-              tags$span(self$get_varname(),
-                class = "filter_panel_varname"
-              ),
+              tags$span(self$get_varname(), class = "filter_panel_varname"),
               if (checkmate::test_character(self$get_varlabel(), min.len = 1) &&
-                tolower(self$get_varname()) != tolower(self$get_varlabel())) {
+                  tolower(self$get_varname()) != tolower(self$get_varlabel())) {
                 tags$span(self$get_varlabel(), class = "filter_panel_varlabel")
               }
             )
@@ -424,7 +424,7 @@ FilterState <- R6::R6Class( # nolint
     }
   ),
 
-  # private members ----
+    # private members ----
   private = list(
     choices = NULL, # because each class has different choices type
     input_dataname = character(0),
@@ -437,11 +437,12 @@ FilterState <- R6::R6Class( # nolint
     varlabel = character(0),
     extract_type = logical(0),
 
-    #' description
-    #' Adds `is.na(varname)` before existing condition calls if `keep_na` is selected.
-    #' Otherwise, if missings are found in the variable `!is.na` will be added
-    #' only if `private$na_rm = TRUE`
-    #' return (`call`)
+    # private methods ----
+    # @description
+    # Adds `is.na(varname)` before existing condition calls if `keep_na` is selected.
+    # Otherwise, if missings are found in the variable `!is.na` will be added
+    # only if `private$na_rm = TRUE`
+    # @return (`call`)
     add_keep_na_call = function(filter_call) {
       if (isTRUE(self$get_keep_na())) {
         call(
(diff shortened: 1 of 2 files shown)


chlebowa:
Use this app to test (thanks @nikolas-burkoff):

```
library(shiny)
library(shinyjs)

repeat {
  x <- c(0.0091, 0.01 + runif(50, max= 0.99))
  if (identical(range(pretty(x, 100)), c(0, 1))) {
    print(range(pretty(x, 100)))
    break
  }
}

my_filter_state <- teal.slice:::init_filter_state(
  x = x,
  varname = "VARNAME",
  varlabel = "A label"
)

#####

ui <- fluidPage(
  useShinyjs(),
  column(4,
         div(  # only need a div so that we can hide the ui when the close button is pressed
           id = "x",
           my_filter_state$ui("fs"),
           h4("Actual data range"),
           textOutput("selection_range"),
           h4("Condition (i.e. call)"), # show the condition generated by this FilterState
           textOutput("condition"),
           h4("Formatted state"), # show the human readable state
           textOutput("formatted"),
           h4("Unformatted state"), # show the raw state
           textOutput("unformatted"),
         )
  ),
  # column(6, verbatimTextOutput("selection_range")),
  NULL
)

  output$condition <- renderPrint(my_filter_state$get_call())
  output$formatted <- renderText(my_filter_state$format())
  output$unformatted <- renderPrint(my_filter_state$get_state()) # also get_selected(), get_keep_na
  
  output$selection_range <- renderPrint(range(x))
  
  signal_to_remove <- my_filter_state$server("fs")  # filterstate$server returns a trigger when the close button is pressed
  observeEvent(signal_to_remove(), shinyjs::hide("x")) # and shinyjs removes the ui
}

shinyApp(ui, server)
```

R/FilterState.R
```diff
@@ -313,8 +315,8 @@ FilterState <- R6::R6Class( # nolint
         )
       )
       value <- private$cast_and_validate(value)
-      value <- private$remove_out_of_bound_values(value)
-      private$validate_selection(value)
+      # value <- private$remove_out_of_bound_values(value)
+      # private$validate_selection(value)
```
nikolas-burkoff, line 319: remove :) - also are we OK removing this? Is it a problem for other classes as it lives in the base class?

R/FilterStateRange.R
```diff
@@ -271,13 +280,8 @@ RangeFilterState <- R6::R6Class( # nolint
...
     },
     cast_and_validate = function(values) {
-      tryCatch(
-        expr = {
-          values <- as.numeric(values)
-          if (any(is.na(values))) stop()
-        },
-        error = function(error) stop("The array of set values must contain values coercible to numeric.")
-      )
+      values <- as.numeric(values)
```
nikolas-burkoff, line 283: do we need a suppresswarnings here?

R/FilterStateRange.R
```diff
@@ -67,21 +69,25 @@ RangeFilterState <- R6::R6Class( # nolint
...
-    #' Default: 0
+    #' @param indent (`numeric(1)`)
+    #'        the number of spaces before after each new line character of the formatted string.
+    #'        Default: 0
     #' @return `character(1)` the formatted string
     #'
     format = function(indent = 0) {
       checkmate::assert_number(indent, finite = TRUE, lower = 0)
 
+      # vals <- pmax(pmin(self$get_selected(), private$choices[2]), private$choices[1])
```
nikolas-burkoff, line 82: I think format should match condition?

image

nikolas-burkoff, line 82: also remove commented code

nikolas-burkoff (commented):
As we're going into main we should have NEWS? We may need a test or two?
Also merge into refactor branch

Could there be a problem in teal that app developers have specified a specific filter i.e. `(list(ADSL = list(x = selected(0.0009, 1))` and that gets automatically rounded for users? 

Add ` actionButton("set_filter", "Set filter state to [0.0091, 1]")` into the test app UI and ` observeEvent(input$set_filter, my_filter_state$set_selected(c(0.0091, 1)))` into its server function

I wouldn't spend too much time on this though as we're moving to allow users to enter a range and not just pick from a slider and at that point we can handle this fully. Please add your thoughts about this (and the complexities of jumping between slider and range [here](https://github.com/insightsengineering/teal.slice/issues/133))

R/FilterState.R
chlebowa: This was actually moved to the `format` method of `FilterStateRanged`, so other `format` methods should (and will) be modified accordingly.

R/FilterStateRange.R
chlebowa: In the previous version these warnings were never actually triggered.
They could be suppressed or they could be logged, depending on how we want to report the set range.

R/FilterStateRange.R
chlebowa: This should be discussed.
Do we want to print the range selected by the user, e.g. _AGE 30-39_, or the range of the actual data, e.g. _AGE 30-38_, because the user selected `30-39` but there are no patients older than 38 in the dataset?

I think the former (_30-39_) as that way different documents can be compared without confusion, although there is an argument to be made that the report should reflect the data and not the user's wishes.

chlebowa:
> Could there be a problem in teal that app developers have specified a specific filter i.e. `(list(ADSL = list(x = selected(0.0009, 1))` and that gets automatically rounded for users?
> 
> Add ` actionButton("set_filter", "Set filter state to [0.0091, 1]")` into the test app UI and ` observeEvent(input$set_filter, my_filter_state$set_selected(c(0.0091, 1)))` into its server function

Works just as well with these modifications, `remove_out_of_bound_values` does raise warnings. I guess you are worried that the lower limit is set to 0.0009 but the slider displays 0.01?

R/FilterStateRange.R
chlebowa: Having discussed with @nikolas-burkoff and @mhallal1 we decided that the format should reflect the selection (input state), not the data.

chlebowa:
Unfortunately, the `initialize` method cannot be simplified that much because there is an edge case of the variable having range 0 (i.e. all values are the same).

chlebowa:
See [here](https://github.com/insightsengineering/teal.slice/blob/184_slider_moves%40main/R/FilterStateRange.R#L279) for how Nik's button problem is "solved".

This means the _subsetting call_ need not contain the exact values passed to the filter API. E.g. one may wish to filter on this:
```
list(
      mtcars = list(
        hp = list(selected = c(52, 65), keep_na = FALSE, keep_inf = FALSE)
      )
    )
```
but the variable in question is:
```
52 62 65 66 66 91 (...)
```
thus the prettified range is:
```
50 55 60 65 70 75 (...)
```
and so the cal will be `>=52` rather than `>=50`.

R/FilterStateRange.R
```diff
@@ -318,18 +313,11 @@ RangeFilterState <- R6::R6Class( # nolint
...
-                pretty_range_inputs[c("min", "max")]
-              } else {
-                self$get_selected()
-              }
-            }),
-            width = "100%",
-            step = pretty_range_inputs["step"]
+            min = private$choices[1],
+            max = private$choices[2],
+            value = private$choices,
```
gogonzo, line 318:
```suggestion
            value = isolate(private$selected()),
```

gogonzo (approved):
Neat 👍 please merge main to the feature branch afterwards

PR #193, opened 2023-02-10
drop unused(?) internal (191)

### LINKED ISSUE #191: drop unused(?) internal
Since #176 made it so that `varname` is always a character string, is [this function](https://github.com/insightsengineering/teal.slice/blob/main/R/call_utils.R#L7) necessary?

### LINKED ISSUE #215: [Bug]: ChoiceFilterState produces invalid call for date and datetime
### What happened?

FilterState is initialized in the specific way. Normally R6 class follows the base class, so that numeric -> RangeFilterState, date -> DateFilterState, .... But, If there are less than `getOption("teal.threshold_slider_vs_checkboxgroup")` unique values, then ChoicesFilterState is initialized. Please analyse the example below, ADSL has extra columns (`numeric_choices`, `date_choices`, `datetime_choices`) and these filters are initialized in the filter-panel.
The call which is produced for all is `<column name> %in% c(<selected character>)`, which means that it looks like this:

- numeric_choices %in% c("1", "2", "3") ✅ 
- date_choices %in% c("2022-02-23", "2022-02-24", "2022-02-25") ❌ 
- datetime_choices %in% c("2022-02-23 22:22:22", "2022-02-24 10:22:22", "2022-02-24 22:22::22") ❌ 

`get_call` for date_choices datetime_choices produces a wrong condition which is FALSE for anything selected. 

```r
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
# options("teal.bs_theme" = bslib::bs_theme(version = 4))
# options(shiny.trace = TRUE)
library(teal)
library(scda)

dates <- as.Date("2022-02-22") + 1:4
datetimes <- as.POSIXct("2022-02-22 22:22:22") + 1:4 * 43200

ADSL$numeric_choices <- sample(1:4, nrow(ADSL), replace = TRUE)
ADSL$date_choices <- sample(dates, nrow(ADSL), replace = TRUE)
ADSL$datetime_choices <- sample(datetimes, nrow(ADSL), replace = TRUE)

funny_module <- function(label = "Filter states", datanames = "all") {
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      output$filter_states <- renderText({
        filter_panel_api |>
          get_filter_state() |>
          yaml::as.yaml()
      })
        attr(data, "code")()
      })
    }
  )
}

default_filters <- list(
  ADSL = list(
    numeric_choices = list(selected = 1:3),
    date_choices = list(selected = dates[1:3]),
    datetime_choices = list(selected = datetimes[1:3])
  )
)


data <- cdisc_data(
  cdisc_dataset("ADSL", ADSL),
  cdisc_dataset("ADRS", ADRS)
)

app <- init(
  data = data,
  filter = default_filters
)

runApp(app)
```

chlebowa:
Closes #191.
Also fixes #215. 

1. Changed internals functions that are used to compose subsetting calls.

Removed functions:
* `call_check_parse_varname`
* `call_extract_list`
* `call_extract_matrix`
* `call_extract_array`
* `call_with_colon`

Removed functions, their actions placed in respective `$get_call` methods:
* `condition_call_logical`
* `condition_call_choice`
* `condition_call_range`
* `condition_call_range_date`
* `condition_call_range_posixct`

Modified functions:
* `call_combine_by`
* `$get_call` methods in `FilterState` children
* `$get_call` method in `FilteredDataCDISC`

2. Switching from range to choices for a low number of possible items that was previously built in to `init_filter_state.numeric` has been expanded dates and datetimes.

3. `POSIXct` and `POSIXlt` are both handled explicitly by `DatetimeFilterState` and `ChoicesFilterState`.


R/FilterStateEmpty.R
```diff
@@ -77,10 +77,11 @@ EmptyFilterState <- R6::R6Class( # nolint
     #'
     get_call = function() {
       filter_call <- if (isTRUE(self$get_keep_na())) {
-        call("is.na", private$get_varname_prefixed())
+        sprintf("is.na(%s)", private$get_varname_prefixed())
```
gogonzo, line 80: I don't think `sprintf` makes sense in this. We don't need to make a string just to convert it to the `lang` instantly. 

`"FALSE"` could be `as.name("FALSE")` then.

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
-    choices <- format(choices)
-  } else if (inherits(choices, c("POSIXct", "POSIXlt"))) {
-    choices <- format(choices)
-  }
-
+  checkmate::assert_string(varname)
+  checkmate::assert_multi_class(
+    choices,
+    c("integer", "numeric", "character", "factor", "Date", "POSIXct", "POSIXlt")
+  )
```
chlebowa, line 38: Should allow all types if, according to documentation of `FilterStateChoices` states that it is only used for characters and factors?

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
-    varname
-  } else if (!choice) {
-    call("!", varname)
-  } else {
-    stop(
-      "Unknown filter state", toString(choice),
-      " for logical var ", as.character(varname)
-    )
-  }
+  sprintf("%s == %s", varname, choice)
```
gogonzo, line 105: I personally prefer `!varname` instead of `varname != logical`. The most robust is `isTRUE(varname)` which wouldn't need `!is.na(varname)` if any value in the vector is missing

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
+    c("integer", "numeric", "character", "factor", "Date", "POSIXct", "POSIXlt")
+  )
 
-  if (length(choices) == 1) {
-    call("==", varname, choices)
+  if (is.numeric(choices)) {
+    if (length(choices) == 1) {
+      sprintf("%s == %s", varname, sprintf("%s", choices))
+    } else {
+      sprintf("%s %%in%% c(%s)", varname, toString(sprintf("%s", choices)))
```
gogonzo, line 44:
```suggestion
      sprintf("%s %%in%% c(%s)", varname, toString(choices))
```

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
+  checkmate::assert_multi_class(
+    choices,
+    c("integer", "numeric", "character", "factor", "Date", "POSIXct", "POSIXlt")
+  )
 
-  if (length(choices) == 1) {
-    call("==", varname, choices)
+  if (is.numeric(choices)) {
+    if (length(choices) == 1) {
+      sprintf("%s == %s", varname, sprintf("%s", choices))
```
gogonzo, line 42:
```suggestion
      sprintf("%s == %s", varname, choices)
```

`%s` causes conversion to string. Alternatively you can use `as.character(choices)`

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
-    c_call <- do.call(
-      "call",
-      append(list("c"), choices)
-    )
-    # c_call needed because it needs to be vector call
-    # instead of vector. SummarizedExperiment.subset
-    # handles only vector calls
-    call("%in%", varname, c_call)
+    if (length(choices) == 1) {
+      sprintf("%s == %s", varname, sprintf("\"%s\"", choices))
```
gogonzo, line 48:
```suggestion
      sprintf('%s == "%s"', varname, choices)
```

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
+    } else {
+      sprintf("%s %%in%% c(%s)", varname, toString(sprintf("\"%s\"", choices)))
```
gogonzo, line 50:
```suggestion
      sprintf("%s %%in%% c(%s)", varname, toString(choices))
```

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
+  checkmate::assert_string(varname)
   checkmate::assert_numeric(range, len = 2, sorted = TRUE)
-  varname <- call_check_parse_varname(varname)
-  call(
-    "&",
-    call(">=", varname, range[1]),
-    call("<=", varname, range[2])
-  )
+
+  sprintf("%s >= %.4f & %s <= %.4f", varname, range[1], varname, range[2])
```
gogonzo, line 80: You can use index to specify which argument to take: `%<index>$<format>` . But besides, I wonder if 4 digits is fine. This condition will be produced based on the slider selection - I believe `%.4f` rounds the values so it might cause some bugs if in shiny (reactive values) we will keep the better precision. Using calls was safer in this case because original number was a part of the call. I will think about this so we can discuss.

```suggestion
  sprintf("%1$s >= %2$.4f & %1$s <= %3$.4f", varname, range[1], range[2])
```

R/call_utils.R
```diff
@@ -163,308 +123,86 @@ call_condition_logical <- function(varname, choice) {
...
-  )
-
-  call(
-    "&",
-    call(">=", varname, call("as.POSIXct", range[1], tz = timezone)),
-    call("<", varname, call("as.POSIXct", range[2], tz = timezone))
+  sprintf(
+    "%s >= as.POSIXct(\"%s\", tz = \"%s\") & %s < as.POSIXct(\"%s\", tz = \"%s\")",
+    varname, format(range[1], tz = timezone), timezone,
+    varname, format(range[2] + 1, tz = timezone), timezone
```
gogonzo, line 143: You can avoid passing the same argument to the sprintf by using indices. Feel free to use escape character or just wrapping text in `''` as in suggestion.
```suggestion
    '%1$s >= as.POSIXct("%2$s", tz = "%4$s") & %s < as.POSIXct("%3$s", tz = "%4$s")',
    varname, format(range[1], tz = timezone), format(range[2] + 1, tz = timezone), timezone
```

tests/testthat/test-FilteredDataset.R
```diff
@@ -152,10 +152,10 @@ testthat::test_that("$get_call returns the filter call of the dataset", {
   filter_call <- shiny::isolate(dataset$get_call())$filter
 
   testthat::expect_equal(
-    deparse1(filter_call),
-    paste0(
+    filter_call,
+    str2lang(paste(
```
gogonzo, line 156: `quote` here would work perfectly, you can break the lines without concatenating string.

tests/testthat/test-FilterStates.R
```diff
@@ -666,7 +666,7 @@ testthat::test_that(
         test_dataset,
         numbers >= 1 & numbers <= 3 &
           choices %in% c("a", "c") &
-          !logical &
+          logical == FALSE &
```
gogonzo, line 669: 😢 What do you think about `!isTrue(logical)`
```r
> x <- c(T, F, NA)
> x == TRUE
[1]  TRUE FALSE NA
> !is.na(x) & x == TRUE
[1]  TRUE FALSE FALSE
> isTRUE(x)
[1] FALSE

```

R/call_utils.R
gogonzo: Should allow more than `character` and `factor`. `ChoicesFilterState` can be based on more than just that.

https://github.com/insightsengineering/teal.slice/blob/1128a868f4fe4934409bf60adca0da3d68e46ec1/R/FilterState-utils.R#L128

In the above line I referenced - I'm not sure whether it should be applied to all data types actually. If we had just 5-dates in a data it doesn't make sense to give a calendar to select dates from.

R/call_utils.R
chlebowa: This one needs the quotemarks.

gogonzo: still, we don't need two `sprintf`. First can have quotemarks

R/call_utils.R
chlebowa: `isTRUE` only returns TRUE for a logical of length 1.

I don't mind `!logical` but this string construction is neater. If not for multiple types in `call_condition_choice`, all of these `sprintf` calls could be moved to the respective `$get_call` methods and these functions could be removed.

R/FilterStateEmpty.R
chlebowa: This was done for consistency with other classes.

Note that `call("is.na", private$get_varname_prefixed())` is insufficient as `private$get_varname_prefixed()` returns character now, so using `call` would require `call("is.na", as.name(private$get_varname_prefixed()))`.

R/call_utils.R
chlebowa: I think we do:
```
> sprintf("%s %%in%% c(%s)", varname, toString(sprintf("\"%s\"", choices)))
[1] "var %in% c(\"A\", \"B\")"

> sprintf("%s %%in%% c(%s)", varname, toString(choices))
[1] "var %in% c(A, B)"

> sprintf("%s %%in%% c(\"%s\")", varname, toString(choices))
[1] "var %in% c(\"A, B\")"
```

R/call_utils.R
chlebowa: Well, currently `ChoicesFilterStates` is created only for characters, factors and numerics. Do we expand to other types?

gogonzo: Yup, I think we should expand to other types as well.

R/call_utils.R
```diff
@@ -49,108 +21,96 @@ call_check_parse_varname <- function(varname) {
...
+  checkmate::assert_string(varname)
   checkmate::assert_numeric(range, len = 2, sorted = TRUE)
-  varname <- call_check_parse_varname(varname)
-  call(
-    "&",
-    call(">=", varname, range[1]),
-    call("<=", varname, range[2])
-  )
+
+  sprintf("%1$s >= %2$.4f & %1$s <= %3$.4f", varname, range[1], range[2])
```
gogonzo, line 80: I think we should change this precision to 10, to follow what we have in the slider. It is still possible that slider initializes with double having 10 digits precision
```suggestion
  sprintf("%1$s >= %2$.10f & %1$s <= %3$.10f", varname, range[1], range[2])
```

```r
library(scda)

ADSL$numeric <- seq(0, by = 1e-8, length.out = nrow(ADSL))

  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      })
        attr(data, "code")()
      })
    }
  )
}

data <- cdisc_data(
  cdisc_dataset("ADSL", ADSL),
  cdisc_dataset("ADRS", ADRS)
)

app <- init(
  data = data,
  filter = list(ADSL = list(numeric = list()))
)


runApp(app)

```

image

gogonzo: > EDIT: shall I increase precision? Or perhaps we want to define this differently, say %.ng?

I'm indifferent about `%g` and `%f`, I think `%f` is better than %g to present in the code for a lower precision numbers (most of the cases)

chlebowa: We also use `toString` for numbers, which seems to round to the 14th decimal. Shall I round numerics to the 10th before passing to `toString` as well?

chlebowa: I set all number-to-string convertions to round the 10th decimal. Seems a little better now:
image

gogonzo (commented):
I'm not sure whether changing calls to strings simplifies and improves readibility of the code. In my opinion the gain is minimal and the biggest simplification was removal `call_extract_list`, `call_extract_array`  and replace them by one-line calls (they could be replaced by simple `call` statements also).

I'm concerned the most with the double conversion of the values. Whatever happens in the future, we will have to remember that possible precision handling is performed in two places:
- number to string for example `range_char <- sprintf("min >= %.10f", min_number)`
- string to call `str2lang(range_char)`

Whatever happens in the future, we will have to remember that possible precision handling is performed in two places.


I'd like this decision to be collective as my point of view is biased.

chlebowa:
> I'm not sure whether changing calls to strings simplifies and improves readibility of the code. In my opinion the gain is minimal and the biggest simplification was removal `call_extract_list`, `call_extract_array` and replace them by one-line calls (they could be replaced by simple `call` statements also).

The main goal was not to improve readability but to work with a consistent type from data/variable names all the way up to the final call. Currently conversions between character and language happen at several stages and type has to be checked at the beginning of many calls (see call_utils for example).

Readability may actually suffer in some lines of code but in my opinion the flow of the code becomes much clearer.

> I'm concerned the most with the double conversion of the values. Whatever happens in the future, we will have to remember that possible precision handling is performed in two places:
> 
> * number to string for example `range_char <- sprintf("min >= %.10f", min_number)`
> * string to call `str2lang(range_char)`
> 
> Whatever happens in the future, we will have to remember that possible precision handling is performed in two places.

Three places, actually, `$fet_pretty_range_step` has this: ` signif(digits = 10, (max(pretty_range) - min(pretty_range)) / n_steps)`.

R/call_utils.R
```diff
@@ -49,422 +20,226 @@ call_check_parse_varname <- function(varname) {
...
+#' call_condition_choice("SEX", choices = "F")
+#' call_condition_choice("SEX", choices = c("F", "M"))
+#' call_condition_choice("x$SEX", choices = Sys.Date())
+#' call_condition_choice("x$SEX", choices = Sys.time())
+#' }
+#' @return a character string
 #' @keywords internal
-call_condition_choice <- function(varname, choices) {
-  varname <- call_check_parse_varname(varname)
+call_condition_choice <- function(varname, choices, ...) {
```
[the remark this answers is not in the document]
chlebowa: I strongly disagree. `init_filter_state` uses S3 dispatch to create a `FilterState` object of the appropriate subclass, which has a specific `$get_call` method, which calls in turn a function from `call_utils`, here `call_condition_choice`. You would have this stage use S3 dispatch again?

Also, I was further going to suggest removing `call_condition_*` functions and transferring their actions to the relevant `$get_call` methods to simplify the codebase.

Same for `call_condition_range`.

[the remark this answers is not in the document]
gogonzo: > Also, I was further going to suggest removing call_condition_* functions and transferring their actions to the relevant $get_call methods to simplify the codebase.

Good point, these call_condition_* methods won't be called anywhere else so transfering them to the class is a good idea. I don't think `ChoicesFilterState$get_call <- function() {...}` should contain current copy of the `call_condition_choice` as in my opinion `ChoicesFilterState` needs a split already (see [this](https://github.com/insightsengineering/teal.slice/blob/6c163aa112023036f53baac35a6d93059a535fbc/R/FilterStateChoices.R#L253))
It seems like this is a more relevant structure:

local-filter-panel-Page-9 drawio (2)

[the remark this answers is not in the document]
chlebowa: So `FilterState` would contain `ChiocesFilterState` and `RangeFilterState` and both of those would further contain `Numeric-`, `Date-`, `DatetimeFilterState`, with `ChoicesFilterState` also containing `Logical-` and `Character/FactorFilterState`, is this correct?

gogonzo:
> The main goal was not to improve readability but to work with a consistent type from data/variable names all the way up to the final call. Currently conversions between character and language happen at several stages and type has to be checked at the beginning of many calls (see call_utils for example).

Agree, having call utils is unnecessary if the classes could handle the calls themselves. Removing few call_extract_* and modifying call_condition_* to string followed by changes in the classes is a half-measure which needs a continuation (most likely on the feature branch).

R/call_utils.R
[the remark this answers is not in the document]
gogonzo: Yup, I think this makes sense. I don't know yet whether factor/character/logical would need a separate class but probably yes - for better encapsulation. Then all call utills (and maybe some more) could be moved to the classes.
Do you agree with this proposition?

R/call_utils.R
chlebowa: Rolled back to `!logical`.

BLAZEWIM:
One note: while testing, happen to run into some (16) warnings coming from `test-call_utils.R`
```
Warning (???): get_call returns a condition evaluating to TRUE for NA values after set_keep_na(TRUE)
unknown timezone 'WMT'
Backtrace:
 1. base::as.POSIXct("1900-01-01 01:24:02", tz = "WMT")
 2. base::as.POSIXct.default("1900-01-01 01:24:02", tz = "WMT")
 5. base::as.POSIXlt.character(x, tz, ...)
 6. base::strptime(xx, f, tz = tz)
```

It looks as if my PC (Windows) does not recognize some timezones abbreviations, WMT and PST to be precise.

What is interesting, those warnings pop out only when running `devtools::test()`, not `devtools::check()`.

R/call_utils.R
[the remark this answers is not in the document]
chlebowa: This is is a serious refactor and has been put off to a later date.

chlebowa:
> A possible easy solution is to remove `as.character` when passing `choiceValues` to `checkboxGroupInput` but it will require a bit of testing and may produce downstream effects.

:point_up_2: This will not work as `checkboxGroupInput` will always return character.

Store variable class in a private field and restore it in `$get_call`?

gogonzo:
> > A possible easy solution is to remove `as.character` when passing `choiceValues` to `checkboxGroupInput` but it will require a bit of testing and may produce downstream effects.
> 
> 👆 This will not work as `checkboxGroupInput` will always return character.
> 
> Store variable class in a private field and restore it in `$get_call`?

This is too much at this moment. I would create a new issue for this. Please create and attach example in the issue where dataset has `categorical`, `date`, `posixct`, `posixlt`, `numeric`, `logical` variables with lett than 5 unique values. Example can be based on cdisc example from https://github.com/insightsengineering/teal.slice/pull/165
Please specify exactly in what cases we have problems, and then we can address remaining issues in the next sprint.

R/FilteredDataCDISC.R
```diff
@@ -105,30 +93,44 @@ CDISCFilteredData <- R6::R6Class( # nolint
...
+            character(0)
+          }
+        parent_keys <- names(keys)
+        dataset_keys <- unname(keys)
+
+        y_arg <-
+          if (length(parent_keys) == 0L) {
+            parent_dataname
+          } else {
+            sprintf("%s[, c(%s), drop = FALSE]", parent_dataname, toString(sprintf("\"%s\"", parent_keys)))
```
gogonzo, line 110: please use dQuote or sQuote
```suggestion
            sprintf("%s[, c(%s), drop = FALSE]", parent_dataname, toString(dQuote(parent_keys)))
```

R/FilteredDataCDISC.R
```diff
@@ -310,9 +312,11 @@ CDISCFilteredData <- R6::R6Class( # nolint
 #' @keywords internal
 #'
 #' @examples
-#' teal.slice:::topological_sort(list(A = c(), B = c("A"), C = c("B"), D = c("A")))
-#' teal.slice:::topological_sort(list(D = c("A"), A = c(), B = c("A"), C = c("B")))
-#' teal.slice:::topological_sort(list(D = c("A"), B = c("A"), C = c("B"), A = c()))
+#' \dontrun{
```
gogonzo, line 315: why not to run?

R/calls_combine_by.R
```diff
@@ -0,0 +1,35 @@
...
+#'
+#' @examples
+#' \dontrun{
+#' calls <- list(
+#'   call_condition_choice("SEX", "F"),
+#'   call_condition_range("AGE", c(20, 50)),
+#'   call_condition_choice("ARM", "ARM: A"),
+#'   call_condition_logical("SURV", TRUE)
+#' )
+#' calls <- lapply(calls, str2lang)
```
gogonzo, line 21: please don't use call_condition_* as they don't exist anymore. Add dummy calls using `quote` or something

tests/testthat/test-FilterState.R
```diff
@@ -155,7 +155,7 @@ testthat::test_that(
 
     testthat::expect_identical(
       shiny::isolate(filter_state$test_add_keep_na_call()),
-      quote(!is.na(test) & TRUE)
+      "!is.na(test) & TRUE"
```
gogonzo, line 158: add_keep_na_call returns a call
```suggestion
      quote(!is.na(test) & TRUE)
```

chlebowa: Test fixes are ready, pending discussion results.

R/calls_combine_by.R
chlebowa: Documentation will be amended once all changes are settled on.

R/FilteredDataCDISC.R
chlebowa: :+1:

chlebowa:
> > > A possible easy solution is to remove `as.character` when passing `choiceValues` to `checkboxGroupInput` but it will require a bit of testing and may produce downstream effects.
> > 
> > 
> > point_up_2 This will not work as `checkboxGroupInput` will always return character.
> > Store variable class in a private field and restore it in `$get_call`?
> 
> This is too much at this moment. I would create a new issue for this. Please create and attach example in the issue where dataset has `categorical`, `date`, `posixct`, `posixlt`, `numeric`, `logical` variables with lett than 5 unique values. Example can be based on cdisc example from #165 Please specify exactly in what cases we have problems, and then we can address remaining issues in the next sprint.

Switching to checkboxes for all data classes has already been added in this PR per your request.

R/FilteredDataCDISC.R
chlebowa: In order to not use `:::` and avoid check errors.

R/FilterStateDatettime.R
```diff
@@ -170,18 +161,31 @@ DatetimeFilterState <- R6::R6Class( # nolint
     #' @description
     #' Returns reproducible condition call for current selection.
     #' For this class returned call looks like
-    #' `<varname> >= as.POSIXct(<min>, tz = <timezone>) & <varname> <= <max>, tz = <timezone>)`
+    #' `<varname> >= as.POSIXct(<min>) & <varname> <= <max>)`
     #' with optional `is.na(<varname>)`.
     get_call = function() {
```
asbates, line 166: I'm getting warnings from the time zone here.

image

From what I can gather `attr(as.POSIXlt(...), "tzone")` is different from `Sys.timezone()` which was used before. The attribute returns "PST" for me, the latter "America/Los_Angeles". When printing a POSIXlt, it prints with "PST". This makes sense because that's typically what is used. However, when trying to specify that as a timezone it gives warnings because it expects it in the "America/Los_Angeles" format.

It looks like you're checking the timezone of the data? Maybe you could get the time zone this way, via attribute. Then check if it's in `olsonNames()`. If it's in this list, use that. Otherwise use `Sys.timezone()`. I think this could resolve the immediate warnings.

More generally, and potentially beyond the scope of this issue, what if we check the timezone of the data on initialization and store that in the class as a reference. I'm not sure if we have rules/guidelines around handling time zones. But to me it makes sense to keep the time zone of the data if present.

asbates (changes requested):
Main comment is on time zones which are always tricky. We may want someone else to weigh in on this.

R/FilterStateDatettime.R
chlebowa: The thing is I am not checking timezones anywhere and I am not getting the warnings you get. This looks like an automatic tz validation step and it is odd that it fails as PST should be a valid name. It certainly is on my machine (Ubuntu 20.04 R 4.2.2) although @BLAZEWIM had some issues with it.

Here is what a comparison returns:
```
> all.equal(
    as.POSIXct("2020-02-20 12:00:00", tz = "PST"), 
    as.POSIXct("2020-02-20 12:00:00", tz = "America/Los Angeles")
  )
[1] "'tzone' attributes are inconsistent ('PST' and 'America/Los Angeles')"
```

The reason I did NOT use Sys.timezone() is because of `airDatePicker` assuming you are in UTC, so results will vary depending on your local tz. For all tests, examples, etc. I set (or attempted to set) tz manually.

R/FilteredDataCDISC.R
```diff
@@ -310,9 +312,11 @@ CDISCFilteredData <- R6::R6Class( # nolint
...
+#' topological_sort(list(A = c(), B = c("A"), C = c("B"), D = c("A")))
+#' topological_sort(list(D = c("A"), A = c(), B = c("A"), C = c("B")))
+#' topological_sort(list(D = c("A"), B = c("A"), C = c("B"), A = c()))
+#' }
 topological_sort <- function(graph) {
```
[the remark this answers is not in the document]
chlebowa: Good point but I would like @gogonzo to confirm.

Why is it even here?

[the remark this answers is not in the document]
gogonzo: It is used in CDISCFilteredData

https://github.com/insightsengineering/teal.slice/blob/abf51503ec187dde04b56bd2c340f162e48479b0/R/FilteredDataCDISC.R#L69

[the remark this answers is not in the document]
chlebowa: Indeed!

gogonzo:
> This is too much at this moment. I would create a new issue for this. Please create and attach example in the issue where dataset has `categorical`, `date`, `posixct`, `posixlt`, `numeric`, `logical` variables with lett than 5 unique values. Example can be based on cdisc example from #165 Please specify exactly in what cases we have problems, and then we can address remaining issues in the next sprint.

https://github.com/insightsengineering/teal.slice/issues/215

R/FilterStateDatettime.R
```diff
@@ -116,19 +116,10 @@ DatetimeFilterState <- R6::R6Class( # nolint
...
       self$set_selected(var_range)
 
-      if (shiny::isRunning()) {
-        session <- getDefaultReactiveDomain()
-        if (!is.null(session$userData$timezone)) {
-          private$timezone <- session$userData$timezone
-        }
-      } else if (isTRUE(attr(x, "tz") != "")) {
-        private$timezone <- attr(x, "tz")
-      }
```
gogonzo, line 130: Why to remove this one? Warnings reported by @asbates few lines below are related to this? @asbates what happens if you add these lines back?

chlebowa: Because I understand this stores the timezone of the local machine, which is redundant.

chlebowa: Sorry, I misspoke. Timezone of the data is what we want, do we not?

chlebowa: I agree.

asbates: Definitely. This makes more sense as another issue.

gogonzo (approved):
Looks good. I confirm #215 as fixed. Timezone issue not related directly with this issue/PR and can be addressed later

Issue #183, opened 2023-02-01
ChoiceFilterState improve init


nikolas-burkoff:
I think these `get_choices_label` functions are too complicated. I think we can simplify and tidy up the way how we obtain counts. In the init of `choicesFilterState` there is `private$histogram_data` which isn't used at all. Having `private$choices` where value is a factor level and name is a count is also not elegant. I think we can introduce 

```
private$choices # factor levels
private$choices_count # unfiltered count
private$f_choices_count # reactive
private$calculate_choices_count(x) 
```

That is just a proposition, I think it would be nice if we tidy up these `initialize` methods and how we preserve counts.

_Originally posted by @gogonzo in https://github.com/insightsengineering/teal.slice/issues/180#issuecomment-1411649446_

donyunardi:
### Acceptance Criteria:
* Simplify the property/attributes of the class
* Update the class' methods

PR #194, opened 2023-02-13
183 choicefs improve init

BLAZEWIM:
# Pull Request

Fixes #[183](https://github.com/insightsengineering/teal.slice/issues/183)


R/FilterStateChoices.R
```diff
@@ -45,7 +45,7 @@ ChoicesFilterState <- R6::R6Class( # nolint
     #' \item{`"matrix"`}{ `varname` in the condition call will be returned as `<dataname>[, <varname>]`}
     #' }
     initialize = function(x,
-                          x_reactive,
+                          x_reactive = NULL,
```
mhallal1, line 48: update docs to imply that the default is NULL

gogonzo: Please remove this default value, we can address this later. If you change `x_reactive` in this class it would be needed also in the others. At this moment `x_reactive` must be provided and it has to be `reactive`. It could be a `reactive(NULL)`, but never NULL

gogonzo (commented):
Please change `x_reactive` to have no defaults in init. Please fix the assert to not let `x_reactive` to be `NULL`, it has to be a reactive.
Please fix the `get_filtered_counts()` call mentioned below.
Besides, it's looking good

R/FilterStateChoices.R
mhallal1: This is still not resolved, please fix before announcing that it is resolved.

BLAZEWIM: yep, sorry, commited locally but did not push

tests/testthat/test-ChoicesFilterState.R
```diff
@@ -1,64 +1,71 @@
 testthat::test_that("The constructor accepts character or factor", {
-  testthat::expect_no_error(ChoicesFilterState$new("test", varname = "test"))
-  testthat::expect_no_error(ChoicesFilterState$new(as.factor("test"), varname = "test"))
+  testthat::expect_no_error(ChoicesFilterState$new("test", x_reactive = reactive(NULL), varname = "test"))
+  testthat::expect_no_error(ChoicesFilterState$new(as.factor("test"), reactive(NULL), varname = "test"))
 })
 
+testthat::test_that("The constructor reactive as x_reactive", {
```
mhallal1, line 6: `testthat::test_that("The constructor accepts only reactive input for x_reactive", {`

mhallal1 (approved):

PR #197, opened 2023-02-13
conflict resolution

chlebowa:
Merging `main` branch into `filter_panel_refactor`  following #189.


R/FilterState-utils.R
```diff
@@ -265,7 +265,7 @@ init_filter_state.POSIXlt <- function(x,
 #' @keywords internal
 #'
 #' @examples
-#' \dontrun{
+#' \donttest{
```
gogonzo, line 268: we want to run all examples during R CMD check and know which examples fail
```suggestion
#' \donttest{
```

Issue #174, opened 2023-01-27
checkboxInput counts causes infinitive loop


gogonzo:
Please change very quickly values of the checkboxInput and see what happens. Reason for this is that by changing a selection for the second time triggers event before the first one completes. The problem lays in the updateCheckboxInput which triggers a reactive. Perhaps we need to find the other way to update the counts and "count bar" for example through `runjs` or something .

updateCheckboxInput can be found in `ChoicesFilterState$srv_inputs`, `RangeFilterState$keep_inf_srv`, `FilterState$keep_na_srv`.

chlebowa:
This only happens when more than one filter is active:
```
teal.gallery::launch_app("basic-teal")
```
Add filter on `iris$Species`, click away and it works fine, but add another filter on `iris` and then a doubleclick on a checkbox triggers the loop. Could it possibly be not because of `update*`?

gogonzo:
> Add filter on iris$Species, click away and it works fine, but add another filter on iris and then a doubleclick on a checkbox triggers the loop. Could it possibly be not because of update*?

There is a possibility. That was my highest likelihood guess after analysing logs and seeing what followed what.

chlebowa:
So @nikolas-burkoff and I tested it in an app that is a little slower than _basic-teal_ and the loop does trigger. Apparently in some cases one simply cannot click quickly enough.

donyunardi:
Per our discussion, the acceptance criteria is to decouple the filter counts/label from the input so we can separate the two activities.

Issue #167, opened 2023-01-19
pickerInput labels (counts ) are too reactive


gogonzo:
Counts of the filter are too reactive. Selecting values instantly trigger re render of the inputs. In case of the pickerInput one is not able to select multiple values because click instantly impacts the ui-input.

https://user-images.githubusercontent.com/36482959/212970289-e8c3a460-d3cc-49e9-bbd8-78824ede4ca2.mp4

We need to find the non-invasive way to update the counts which won't retrigger reactive events. Also `pickerInput` selection should be applied only when picker is closed https://github.com/dreamRs/shinyWidgets/issues/392

donyunardi:
Could be related with #173 and #174
Per our discussion, the accepted solution is to decouple the filter counts/label from the input so we can separate the two activities.

PR #198, opened 2023-02-14
updateable count labels

### LINKED ISSUE #173: radioButton counts causes infinitive loop
Please change very quickly values of the radioButton (LogicalFilterState) and see what happens. Reason for this is that by changing a selection for the second time triggers event before the first one completes. The problem lays in the `updateRadioButton` which triggers a reactive. Perhaps we need to find the other way to update the counts and "count bar" for example through `runjs` or something .

gogonzo:
closes #173 #174 #167 
Tested on: chrome, safari and firefox with default shiny, BS3-5. No major styling/js issues. 

In this PR we've changed the way how the labels are created and updated. Basically, counts are the separate shiny/html elements which have their own `id`, which they are identified by. 
1. We introduced new shiny inputs to create and update values of the labels:
- countBar - creates n ui elements for each choice with progressbar and text label
- countBar - creates a ui html element with progressbar and text label
- make_count_text - composes the label `$choice ($filtered/$unfiltered)`

2. We've created respective update* methods to keep the labels up to date
- updateCountBars
- updateCountBar
- updateCountText

3. To update the counts without triggering reactive cycle we created two js handlers which change a bar and a label found using `id`.

4. Instead of `observeEvent(filtered_counts(), ...)` we used `renderUI(... filtered_counts())` which serves as the observer for the visible item only!

5. In the ChoicesFilterState we applied [this](https://github.com/dreamRs/shinyWidgets/issues/392) to trigger change of the selection only when the picker is closed - closing #167 

Untitled Diagram drawio (2)

You can find examples in #165 and one can use also `teal.gallery::launch_app`.


asbates:
I would consider having the progress bars CSS as the following.

- Add border radius to right side of unfiltered bar to match the left side
- Use variables where possible to maintain user theme if they have one set. (`var(--color, othercolor)` will use `--color` if available, otherwise `othercolor`.
- Instead of built in css lightblue, use Shiny's default primary color to better match theme.

```css
.progress-bar.state-count-bar-filtered {
  background-color: var(--bs-primary, rgb(51, 122, 183));
  opacity: 0.8;
  color: var(--bs-body-color, var(--dark, #333333));
  overflow: visible;
  text-align: left;
  white-space: nowrap;
}

.progress-bar.state-count-bar-unfiltered {
  background-color: var(--bs-primary, rgb(51, 122, 183));
  opacity: 0.439;
  border-top-right-radius: var(--bs-border-radius, .375rem);
  border-bottom-right-radius: var(--bs-border-radius, .375rem);
}
```

donyunardi:
Thanks for working on this @gogonzo. I think this is looking great already. The js handler solution works great on addressing our reactive cycle issue.

@lcd2yyz 
Could you please look at this PR, specifically the progress bar?

inst/css/filter-panel.css
```diff
@@ -251,3 +251,22 @@ a.remove_all:hover {
...
+  background-color: lightblue !important;
+  color: var(--bs-body-color, var(--dark, #333333));
+  overflow: visible;
+  text-align: left;
+  white-space: nowrap;
+}
+
+.progress-bar.state-count-bar-unfiltered {
+  background-color: rgba(173, 216, 230, 0.439) !important;
+}
```
gogonzo, line 272: @asbates do you know how to apply alpha/opacity to the background only? Opacity affects text and background in the same time which means that it could be barely visible,

image


```suggestion
.progress-bar.state-count-bar-filtered {
  background-color: var(--bs-primary, rgb(51, 122, 183));
  opacity: 0.8;
  color: var(--bs-body-color, var(--dark, #333333));
  overflow: visible;
  text-align: left;
  white-space: nowrap;
}

.progress-bar.state-count-bar-unfiltered {
  background-color: var(--bs-primary, rgb(51, 122, 183));
  opacity: 0.439;
  border-top-right-radius: var(--bs-border-radius, .375rem);
  border-bottom-right-radius: var(--bs-border-radius, .375rem);
}
```

asbates: I do not. Bootstrap uses hex for color variables and as of now there isn't a way to add alpha to an existing hex color. (apparently this is in the works). So setting opacity on the whole bars is probably not the way to go. The text looked light to me but I wanted to bring it up just in case it was just me.

R/count_labels.R
```diff
@@ -0,0 +1,213 @@
+#' Progress bars with labels
+#'
+#' `shiny` element showing progressbar counts. Each element can has
```
mhallal1, line 3:
```suggestion
#' `shiny` element showing progressbar counts. Each element can have a
```

R/count_labels.R
```diff
@@ -0,0 +1,213 @@
...
+#' counts <- table(choices)
+#' labels <- countBars(
+#'   inputId = "counts",
+#'   choices = c("a", "b", "c"),
+#'   countsmax = c(20, 20, 20),
+#'   countsnow = unname(counts)
+#' )
+#'
+#'
+#' shinyApp(
```
mhallal1, line 30: Screenshot from 2023-02-20 11-11-47

R/count_labels.R
```diff
@@ -0,0 +1,213 @@
...
+    message = list(
+      id = session$ns(inputId),
+      label = label
+      )
+    )
+}
+
+#' Make a count text
+#'
+#' Returns a text describing filtered counts. Text is composed in following way:
```
mhallal1, line 195:
```suggestion
#' Returns a text describing filtered counts. The text is composed in the following way:
```

mhallal1 (commented):
Issue created while reviewing this: https://github.com/insightsengineering/teal.slice/issues/203

R/count_labels.R
mhallal1: fixed, thanks

R/count_labels.R
```diff
@@ -0,0 +1,214 @@
...
+#' @param inputId (`character(1)`) `shiny` id
+#' @param choices (`vector`) determines label text.
+#' @param countsmax (`numeric`) determining maximal count of each element.
+#'  Length should be the same as `choices`.
+#' @param countsnow (`numeric`) actual counts of each element.
+#'  Length should be the same as `choices`.
+#' @param counttotal (`numeric(1)`) Determines the size of the whole progress bar.
+#'  For example, in case there are three choices with some counts, one might like
+#'  to display each progress bar with the relative size to the `totalcount`.
+#' @return list of `shiny.tag`
```
mhallal1, line 17: `counttotal` is an an argument for  `countBars`

mhallal1: neither the session

mhallal1 (approved):
Tested locally and on BEE with multiple apps and conditions.
Works as expected.

R/count_labels.R
gogonzo: session is there - countBars and updateCountBars have a shared documentation page

image

mhallal1: fair enough,

R/utils.R
```diff
@@ -104,3 +104,22 @@ get_teal_bs_theme <- function() {
...
+#' not work with `devtools`. Therefore, we redefine this method in each package
+#' as needed. Thus, we do not export this method
+#'
+#' @param pattern (`character`) pattern of files to be included, passed to `system.file`
+#' @param except (`character`) vector of basename filenames to be excluded
+#'
+#' @return HTML code that includes `JS` files
+#' @keywords internal
+include_js_files <- function(pattern = NULL, except = NULL) {
+  checkmate::assert_character(except, min.len = 1, any.missing = FALSE, null.ok = TRUE)
```
chlebowa, line 120: `min.len = 1` precludes NULL

gogonzo: I've removed exclude at all as it's not used in this package. However, `min.len = 1` and `null.ok = TRUE` can be specified together which prevents someone from specifying `character(0)`

R/count_labels.R
```diff
@@ -0,0 +1,211 @@
...
+
+    countBar(
+      inputId = ns(i),
+      label = choice,
+      countmax = countmax,
+      countnow = countnow,
+      counttotal = counttotal
+    )
+  })
+}
```
chlebowa, line 82: I suggest `mapply`:

```suggestion
countBars <- function(inputId, choices, countsmax, countsnow = NULL) {
  checkmate::assert_string(inputId)
  checkmate::assert_vector(choices)
  checkmate::assert_numeric(countsmax, len = length(choices))
  checkmate::assert_numeric(countsnow, len = length(choices), null.ok = TRUE)
  if (!is.null(countsnow)) {
    checkmate::assert_true(all(countsnow <= countsmax))
  } else {
    countsnow <- rep(0L, length(choices))
  }

  ns <- NS(inputId)

  mapply(
    countBar,
    inputId = ns(seq_along(choices)),
    label = as.character(choices),
    countmax = countsmax,
    countnow = countsnow,
    MoreArgs = list(
      counttotal = sum(countsmax)
    ),
    SIMPLIFY = FALSE, USE.NAMES = FALSE)
}
```

R/utils.R
chlebowa: But if `min.len = 1`, NULL will not pass as it has length 0. `min.len = 1` is sufficient here.

gogonzo:
```r
checkmate::assert_character(NULL, min.len = 1, any.missing = FALSE, null.ok = TRUE)
#> silent
```

chlebowa: I stand corrected.

R/count_labels.R
```diff
@@ -0,0 +1,211 @@
...
+      updateCountBar(
+        inputId = ns(i),
+        label = choice,
+        countmax = countmax,
+        countnow = countnow,
+        counttotal = counttotal
+      )
+    })
+  invisible(NULL)
+}
```
chlebowa, line 148: Also suggest `mapply` here

```suggestion
updateCountBars <-  function(session = getDefaultReactiveDomain(), inputId, choices,
                                  countsmax, countsnow = NULL) {
    checkmate::assert_string(inputId)
    checkmate::assert_vector(choices)
    checkmate::assert_numeric(countsmax, len = length(choices))
    checkmate::assert_numeric(countsnow, len = length(choices), null.ok = TRUE)
    if (!is.null(countsnow)) {
      checkmate::assert_true(all(countsnow <= countsmax))
    } else {
      countsnow <- rep(0L, length(choices))
    }

    ns <- NS(inputId)
    mapply(
      updateCountBar,
      inputId = ns(seq_along(choices)),
      label = choices,
      countmax = countsmax,
      countnow = countsnow,
      MoreArgs = list(
        counttotal = sum(countsmax)
      )
    )
  invisible(NULL)
}
```

gogonzo:
After addressing all comments from @chlebowa and @BLAZEWIM I'm using @mhallal1 approval. If you still see the room for improvement please create an issue 👍

PR #200, opened 2023-02-16
example apps for filter panel backend

### LINKED ISSUE #179: Add Helper apps into roxygen

chlebowa:
Closes #179.

Adds example apps in `ChoicesFilterState` and `DFFilterStates`.


gogonzo (commented):
Added examples are fine but acceptance criterion is: Add helper document in two classes FilterState and FilterStates. 
At this moment only DFFilterStates and ChoicesFilterState have examples while other subclasses don't, so I think there is more add to do here

donyunardi:
I think it's more efficient to have the example app in the abstract classes, even though the example app does create `ChoicesFilterState` and `DFFilterStates`.

It felt more inclusive rather than adding the example apps in each subclass.
Another idea is to create a link in the subclass roxygen to the abstract class roxygen for an example app.

chlebowa:
`FilterState` subclasses have their own example apps now. It has been suggested that they be combined in a single app in `FilterState` abstract class documentation, but I am not sure about it now as it would be a huge block of code.

This PR goes into the `main` branch. One the refactor branch is merged, each `filter_state` definition must specify `x_reactive` and the app UI must have 
```
    include_css_files(pattern = "filter-panel"),
    include_js_files(pattern = "count-bar-labels")
```
added.

chlebowa:
During a discussion with @gogonzo we decided
* adding a composite app to the `FilterState` abstract class would be too much as it is ~250 lines
* other `FilterStates` classes do not need examples.

I think `EmptyFilterState` can do without as well.

R/FilterStateChoices.R
```diff
@@ -15,6 +16,66 @@
...
 #' isolate(filter_state$set_keep_na(TRUE))
 #' isolate(filter_state$get_call())
+#'
+#' \dontrun{
+#' # working filter in an app
+#' library(shiny)
+#'
+#' data_choices <- c(letters[1:4], NA)
+#' filter_state_choices <- ChoicesFilterState$new(
+#'   x = data_choices,
```
gogonzo, line 26: set `x` as a vector with repeated levels for meaningful counts
```suggestion
#'   x = sample(data_choices, 100, replace = TRUE),
```

chlebowa: :+1:

R/FilterStateChoices.R
```diff
@@ -15,6 +16,66 @@
...
+#' # working filter in an app
+#' library(shiny)
+#'
+#' data_choices <- c(sample(letters[1:4], 100, replace = TRUE), NA)
+#' filter_state_choices <- ChoicesFilterState$new(
+#'   x = data_choices,
+#'   varname = "variable",
+#'   varlabel = "label"
+#' )
+#' filter_state_choices$set_state(list(selected = data_choices[c(1, 3)], keep_na = TRUE))
```
gogonzo, line 30: `data_choices[1]` and `data_choices[3]` can be identical - I don't know what can happen if one sets `selected = c("a", "a")`. Also `data_choices[1]` or `data_choices[3]` can be `NA` which is wrong

```suggestion
#' filter_state_choices$set_state(list(selected = data_choices[c(1, 3)], keep_na = TRUE))
```

chlebowa: Indeed, fixed.

gogonzo (approved):
👍

PR #205, opened 2023-02-22
166 filtered counts@filter panel refactor@main

### LINKED ISSUE #166: Counts of the filtered state are missleading
So the counts I think are not as helpful as they could be - when you deselect "M" you get 0/399:
image

But it's probably more informative to have:
image

So you can see there are 305 available given other filtering out of the 399 in total. And then checking/un-checking the checkbox does not change the label

Similar for keep_na and keep_inf

gogonzo:
Closes #166 

Let me describe the diagram below to fully explain what is the problem and what solution I propose. Below I'm attaching sequence diagram adapted to the structure we have. As you already know:
- `FilteredData$init` initializes and stores `FilteredDataset` (1 to many)
- `FilteredDataset$init` initializes and stores `FilterStates`. Relationship 1 to many in case of MAEFilteredDataset but in the case of data.frame it's just 1..1, and we are going to focus on this so far.
- `FilterStates` initializes `FilterState` through `set_filter_state` (unlike other classes which are created imediatelly in the init of parent)

So above explains `init` - the first and second sequence in the diagram. As you can see `init` and `set_filter_state` doesn't return anything as there is no "arrow" below to return.

WIP.....

image

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -76,6 +76,7 @@ FilterState <- R6::R6Class( # nolint
                           varlabel = character(0),
                           dataname = NULL,
                           extract_type = character(0)) {
+      checkmate::assert_class(x_reactive, "function")
       checkmate::assert_string(varname)
       checkmate::assert_character(varlabel, max.len = 1, any.missing = FALSE)
       checkmate::assert_string(dataname, null.ok = TRUE)
@@ -86,7 +87,7 @@ FilterState <- R6::R6Class( # nolint
       if (length(extract_type) == 1 && is.null(dataname)) {
         stop("if extract_type is specified, dataname must also be specified")
       }
-
+      private$id <- sample.int(.Machine$integer.max, 1)
       private$dataname <- dataname
       private$varname <- varname
       private$varlabel <- if (identical(varlabel, as.character(varname))) {
@@ -99,8 +100,8 @@ FilterState <- R6::R6Class( # nolint
       private$selected <- reactiveVal(NULL)
       private$na_count <- sum(is.na(x))
       private$keep_na <- reactiveVal(FALSE)
-      private$x_reactive <- x_reactive
-      private$filtered_na_count <- reactive(sum(is.na(x_reactive())))
+      private$x_reactive <- reactive(x_reactive(private$id))
+      private$filtered_na_count <- reactive(sum(is.na(private$x_reactive())))
       logger::log_trace(
         sprintf(
           "Instantiated %s with variable %s, dataname: %s",
@@ -110,6 +111,10 @@ FilterState <- R6::R6Class( # nolint
         )
       )
       invisible(self)
+    },
+    finalize = function() {
+      print("finalize")
+      self$destroy_observers()
     },
 
     #' @description
@@ -182,6 +187,10 @@ FilterState <- R6::R6Class( # nolint
       } else {
         character(1)
       }
+    },
+
+    get_id = function() {
+      private$id
     },
 
     #' @description
@@ -400,6 +409,7 @@ FilterState <- R6::R6Class( # nolint
 
   # private members ----
   private = list(
+    id = integer(0),
     choices = NULL, # because each class has different choices type
     dataname = character(0),
     keep_na = NULL, # reactiveVal logical()
@@ -578,13 +588,14 @@ FilterState <- R6::R6Class( # nolint
         # changed directly by the api - then it's needed to rerender UI element
         # to show relevant values
         private$observers$keep_na_api <- observeEvent(
+          eventExpr = self$get_keep_na(),
           ignoreNULL = FALSE, # nothing selected is possible for NA
           ignoreInit = TRUE, # ignoreInit: should not matter because we set the UI with the desired initial state
-          eventExpr = self$get_keep_na(),
           handlerExpr = {
             if (!setequal(self$get_keep_na(), input$value)) {
               updateCheckboxInput(
                 inputId = "value",
+                label = sprintf("Keep NA (%s/%s)", private$filtered_na_count(), private$na_count),
                 value = self$get_keep_na()
               )
             }
--- a/R/FilterStateLogical.R
+++ b/R/FilterStateLogical.R
@@ -216,6 +216,7 @@ LogicalFilterState <- R6::R6Class( # nolint
           # this observer is needed in the situation when private$selected has been
           # changed directly by the api - then it's needed to rerender UI element
           # to show relevant values
+          finite_values <- reactive(Filter(Negate(is.na), private$x_reactive()))
           output$trigger_visible <- renderUI({
             logger::log_trace(sprintf(
               "LogicalFilterState$server@1 updating count labels in variable: %s , dataname: %s",
@@ -226,7 +227,7 @@ LogicalFilterState <- R6::R6Class( # nolint
               inputId = "labels",
               choices = as.character(private$choices),
               countsmax = as.numeric(names(private$choices)),
-              countsnow = unname(table(factor(private$x_reactive(), levels = private$choices)))
+              countsnow = unname(table(factor(finite_values(), levels = private$choices)))
             )
             NULL
           })
@@ -274,6 +275,7 @@ LogicalFilterState <- R6::R6Class( # nolint
           private$keep_na_srv("keep_na")
 
           logger::log_trace("LogicalFilterState$server initialized, dataname: { private$dataname }")
+          finite_values <- reactive(Filter(is.finite, private$x_reactive()))
           NULL
         }
       )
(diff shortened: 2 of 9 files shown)


R/FilterStates.R
```diff
@@ -133,7 +133,10 @@ FilterStates <- R6::R6Class( # nolint
         simplify = FALSE,
         function(state_list) {
           items <- state_list()
-          filtered_items <- Filter(f = function(x) x$is_any_filtered(), x = items)
+          filtered_items <- Filter(
+            f = function(x) x$is_any_filtered() && !identical(x$get_id(), sid),
```
nikolas-burkoff, line 137: I wonder if we should allow sid to be a vector and have `!(x$get_id() %in% sid)` for future proofing?

R/FilterState.R
```diff
@@ -76,6 +76,7 @@ FilterState <- R6::R6Class( # nolint
                           varlabel = character(0),
                           dataname = NULL,
                           extract_type = character(0)) {
+      checkmate::assert_class(x_reactive, "function")
```
nikolas-burkoff, line 79: lets change the name as it's not reactive anymore - especially as there's private$x_reactive which is reactive inside the class - it's a shame we need both the function and the reactive of the evaluated function - would be nice not to, though suspect we do need them both

R/FilterState.R
```diff
@@ -86,7 +87,7 @@ FilterState <- R6::R6Class( # nolint
       if (length(extract_type) == 1 && is.null(dataname)) {
         stop("if extract_type is specified, dataname must also be specified")
       }
-
+      private$id <- sample.int(.Machine$integer.max, 1)
```
nikolas-burkoff, line 90: I think having the id inside the class is ok as an object can know "who it is". I think the id should be passed in from FilterStates however - and only created here if no id is passed in

R/FilterStateLogical.R
```diff
@@ -216,6 +216,7 @@ LogicalFilterState <- R6::R6Class( # nolint
           # this observer is needed in the situation when private$selected has been
           # changed directly by the api - then it's needed to rerender UI element
           # to show relevant values
+          finite_values <- reactive(Filter(Negate(is.na), private$x_reactive()))
```
nikolas-burkoff, line 219: finite_values is confusing here non_missing_values is better

R/FilterStates.R
```diff
@@ -123,7 +123,7 @@ FilterStates <- R6::R6Class( # nolint
     #'
     #' @return `call` or `NULL`
     #'
-    get_call = function() {
+    get_call = function(sid = integer(0)) {
```
nikolas-burkoff, line 126: I'd be explicit with a name with get_call: `sid -> excluded_filterstate_ids` or something like that

R/FilterState.R
```diff
@@ -76,6 +76,7 @@ FilterState <- R6::R6Class( # nolint
                           varlabel = character(0),
```
nikolas-burkoff, line 76: we should make sure the default x_reactive works as I'd love to keep being able to use filterstate outside of filterstates as it makes it easier to understand the classes

nikolas-burkoff (commented):
A few initial comments from me inline

I understand what's going on here, but it is really complicated - it would conceptually be simpler if the FilterState was given the column of data filtered by all other filters and then the application of the call was directly on the FilterState as then the data flow is much clearer - however, I suspect in practice that will be much much harder to do.

R/FilterState.R
gogonzo: Yup, this is exactly what I'm trying to do now:
- remove FilterState$id
- implement `id` in `FilterStates$set_filter_state` where new `FilterState` will have attribute attached in `state_list`

R/FilterStates.R
gogonzo: I've chosen this name as `excluded_id` or similar could be quickly changed. We have a straight way now to implement hierarchical filter. Let's assume we have 4 filters and we'd like to update counts of the filter-3:

- 1
- 2
- 3**
- 4

```
# 1. my branch here - all filters except this one
states <- private$state_list[where sid != sid]

# 2. future - hierarchical
states <- private$state_list[where sid < sid]
```

So the phrase `excluded` doesn't really have anything in common with (2). Other proposition?

R/FilterStates.R
```diff
@@ -133,7 +133,16 @@ FilterStates <- R6::R6Class( # nolint
...
-          filtered_items <- Filter(f = function(x) x$is_any_filtered(), x = items)
+          filtered_items <- Filter(
+            f = function(x) x$is_any_filtered() && !identical(attr(x, "sid"), sid),
+            x = items
+          )
+          # hierarchical
+          #filtered_items <- Filter(
+          #  f = function(x) x$is_any_filtered() && attr(x, "sid") < sid,
+          #  x = items
+          #)
```
gogonzo, line 144: Uncomment to get hierarchical!

nikolas-burkoff: @npaszty 🥳 (it'll be a long time until it's in main but thought you'd be happy)

R/FilterStates.R
gogonzo: Let's keep `sid` a scalar now. If we get to thinking about filter-group then we can decide how to address this. I think almost everything needed is here already

R/FilterState.R
gogonzo: Done - I've added default value `reactive(NULL)` to all classes

R/FilterStateLogical.R
gogonzo: done

R/FilterState.R
gogonzo: Simplified initialize so it doesn't have two places where code is evaluated. The only thing to discuss is a naming

nikolas-burkoff:
- The numeric slider from the cdisc example app isn't working:
image
(the AVAL one is so I don't know if this is a keep inf/na issue or due to the very small range) - the left end can't be moved and the right side snaps to 0

R/FilterStateLogical.R
```diff
@@ -335,6 +334,7 @@ LogicalFilterState <- R6::R6Class( # nolint
           private$keep_na_srv("keep_na")
 
           logger::log_trace("LogicalFilterState$server initialized, dataname: { private$dataname }")
+          non_missing_values <- reactive(Filter(is.finite, private$x_reactive()))
```
nikolas-burkoff, line 337: Already on line 278?

R/FilterStates-utils.R
```diff
@@ -6,10 +6,11 @@
...
 #'   the R object which `subset` function is applied on.
 #'
-#' @param data_reactive (`reactive`)\cr
-#'   should return an object constistent with the `FilterState` class or `NULL`.
+#' @param data_reactive (`function(sid)`)\cr
+#'   should return an object of the same type as `data` or `NULL`.
 #'   This object is needed for the `FilterState` counts being updated
-#'   on a change in filters. If `reactive(NULL)` then filtered counts are not shown.
+#'   on a change in filters. If function returns `NULL` then filtered counts are not shown.
+#'   Function has to have `sid` argument being a character.
```
nikolas-burkoff, line 13: I may add why it needs to have this

R/FilterStatesDF.R
```diff
@@ -137,22 +137,33 @@ DFFilterStates <- R6::R6Class( # nolint
     #' @param data (`data.frame`)\cr
     #'   the R object which `dplyr::filter` function is applied on.
     #'
-    #' @param data_reactive (`reactive`)\cr
-    #'   should a `data.frame` object or `NULL`.
+    #' @param data_reactive (`function(sid)`)\cr
+    #'   should return a `SummarizedExperiment` object or `NULL`.
```
nikolas-burkoff, line 141: Think this is wrong?

R/FilterStates.R
```diff
@@ -550,6 +555,79 @@ FilterStates <- R6::R6Class( # nolint
...
+          # objects has random and unique sid attribute
+          #  which allows a reactive from below to find a right object in the state_list
+          sid <- sprintf("%s-%s-%s", state_list_index, varname, sample.int(size = 1L, n = .Machine$integer.max))
+          fstate <- init_filter_state(
+            x = if (!is.array(data)) {
+              data[[varname]]
+            } else {
+              data[, varname]
+            },
+            x_reactive = if (!is.array(data)) {
```
nikolas-burkoff, line 614: i would add a code comment here

nikolas-burkoff (commented):
Code looks good - a few minor comment in line + making the labels work for keep NA and Inf in filterstate. I'll test apps again and then if I don't find anything I'll approve

gogonzo:
@nikolas-burkoff I'll include this in the vignette with UMLs I made if we get to some stable point.

```
I wonder if it makes sense to add some explanation in the PR description into the roxygen as what is going on here is complicated and I expect for someone new coming into this code it's going to be tough
```

nikolas-burkoff (approved):
yes this is nice 👍

PR #207, opened 2023-02-23
Disabling of a single filter state

### LINKED ISSUE #15: Enable / disable filters on a variable level (but not removing it)

gogonzo:
closes #15 
Allows for a single filter state to be disabled and enabled.

* Adds and adjusts fields and methods to `xFilterState` classes to support disabling and enabling.
* Add switch to filter card header to control disabling.
* When switch is disabled the state is cached and methods like `is_any_filtered()` reflect the new NULL state. So outside classes like `FilterStates` don't recognize the disabled `FilterState` as applying any filtering.
* When switch is enabled, the state is restored from the cache.
* Inputs are also disabled/enabled based on the switch.

You can test with the apps in #165


R/FilterStateChoices.R
```diff
@@ -92,6 +92,9 @@ ChoicesFilterState <- R6::R6Class( # nolint
     #' optional `is.na(<varname>)`.
     #' @return (`call`)
     get_call = function() {
+
+      if (private$is_disabled()) return(NULL)
```
gogonzo, line 96: as state is cached the call will return NULL anyway - this `if` is unnecessary

R/FilterState.R
```diff
@@ -389,11 +404,16 @@ FilterState <- R6::R6Class( # nolint
       ns <- NS(id)
 
       theme <- getOption("teal.bs_theme")
+      enable <- if (private$is_disabled()) {
+        FALSE
+      } else {
+        TRUE
+      }
```
gogonzo, line 411:
```suggestion
      enable <- !private$is_disabled()
```

R/FilterState.R
```diff
@@ -412,6 +432,8 @@ FilterState <- R6::R6Class( # nolint
     extract_type = logical(0),
     x_reactive = NULL, # reactive containing the filtered variable, used for updating counts and histograms
     filtered_na_count = NULL, # reactive containing the count of NA in the filtered dataset
+    disabled = FALSE, # is this filter disabled?
```
gogonzo, line 435: default should be `NULL` as we use reactiveVal here
```suggestion
    disabled = NULL, # reactiveVal returning logical
```

R/FilterStateChoices.R
asbates: So I made this change and now dates and datetimes are throwing errors. More specifically, the [type check](https://github.com/insightsengineering/teal.slice/blob/c6fec4c6bf946a5c72e9c7367268640d3de97754/R/call_utils.R#L212) in `call_condition_range_date` is failing.

There's some odd reactivity behavior going on here. When a filter state is disabled (and state set to `NULL`) **only** the date and datetime `get_call` methods are called. This is where the error occurs. For the other filter state types though, `get_call` is _not_ called on disabling. Any idea what might be causing this? Should I revert this change for dates and datetimes?

asbates: Never mind, I found it.

mhallal1 (commented):
The toggle here is not clear to me:
Screenshot from 2023-02-27 17-51-21

I would suggest something more informative like:
Screenshot from 2023-02-27 17-55-21

Screenshot from 2023-02-27 17-55-11

lcd2yyz:
> What do you think about this?
> 
> image
> 
> @lcd2yyz what do you think?

I like this one better, the check/cross mark coupled with the coloring will ensure no ambiguity.
But consistency is also important - was there any discussion when selecting the toggle button for the entire filter panel or just chosen by whoever implemented it at the time? If the latter, can we switch the button for entire panel to be the same as above? 

Also, on a more inquisitive look, does it actually make sense to place the filter panel toggle on the upper right corner? Perhaps it would it be more logical to place it inside the "Active Filter Variable" panel instead? 
image

R/FilterState.R
```diff
@@ -254,7 +255,7 @@ FilterState <- R6::R6Class( # nolint
     #' @return NULL invisibly
     #'
     set_keep_na = function(value) {
-      checkmate::assert_flag(value)
+      checkmate::assert_flag(value, null.ok = TRUE)
```
gogonzo, line 258: Hmm, why is it eased? I think we should limit to TRUE/FALSE only.

R/FilterState.R
```diff
@@ -304,10 +305,15 @@ FilterState <- R6::R6Class( # nolint
           private$dataname
         )
       )
-      value <- private$cast_and_validate(value)
-      value <- private$remove_out_of_bound_values(value)
-      private$validate_selection(value)
-      private$selected(value)
+      if (private$is_disabled()) {
+        private$selected(value)
```
gogonzo, line 309: I think we should throw warning if it's disabled and avoid setting value. Setting value of disabled filter is confusing mechanism which causes `$get_selected` to return values and `$get_call` to return condition. Enabling will overwrite these values anyway, so they are redundant and potentially could cause some bugs.

R/FilterState.R
```diff
@@ -507,6 +525,43 @@ FilterState <- R6::R6Class( # nolint
...
+    #
+    # NA counts are based on unfiltered and filtered (optional) `x`.
+    get_na_label = function() {
+      sprintf(
+        "Keep NA (%s%s)",
+        if (is.null(private$x_reactive())) "" else sprintf("%s/", private$filtered_na_count()),
+        private$na_count
+      )
+    },
+
```
gogonzo, line 538: Doesn't seem to be used
```suggestion
```

R/FilterStateChoices.R
```diff
@@ -448,11 +462,15 @@ ChoicesFilterState <- R6::R6Class( # nolint
         id = id,
         function(input, output, session) {
           output$summary <- renderUI({
-            n_selected <- length(self$get_selected())
-            tagList(
-              tags$span(sprintf("%s levels selected", n_selected)),
-              if (self$get_keep_na()) tags$span("NA") else NULL
-            )
+            if (private$is_disabled()) {
+              tags$span("Disabled")
```
gogonzo, line 466: Seems that:
- ui_summary is identical across classes and could be moved to the FilterState
- srv_summary could be "abstract" and moved to `FilterState`.
```r
output$summary <- renderUI({
  if (private$is_disabled() {
    tags$span("Disabled")
  } else {
    private$summary()
  }
}}
```


- `XyzFilterState` could have a `private$summary` which returns these summary text elements

gogonzo (commented):
I think `is_disabled` is not robust enough, we could also benefit by making `ui_sumary` and `srv_summary` more abstract.

asbates:
> I like this one better, the check/cross mark coupled with the coloring will ensure no ambiguity. But consistency is also important - was there any discussion when selecting the toggle button for the entire filter panel or just chosen by whoever implemented it at the time? If the latter, can we switch the button for entire panel to be the same as above?
> 
> Also, on a more inquisitive look, does it actually make sense to place the filter panel toggle on the upper right corner? Perhaps it would it be more logical to place it inside the "Active Filter Variable" panel instead? image

@lcd2yyz I can't say whether there was a discussion or not for the main panel switch, I don't know. As far as location, yes I think it makes more sense in the Active Filter Variables section as well. I think this is outside the scope of the current issue so I opened #217. @donyunardi can we get this added on an upcoming sprint please.

R/FilterState.R
asbates: When the filter is disabled, the state values are set to `NULL`. So we need this here to allow for that.

R/FilterState.R
asbates: I disagree. I think it's misleading if the filter state is disabled but there are  values set, which is the case if we don't set the state to `NULL`. If you run one of the apps in #165 on this branch, you'll see the main output shows that a disabled filter state has values. To me that implies the filter is active.


As far as `$get_selected` and `$get_call`, if the filter state is disabled, and the state set to `NULL`, these methods will return `NULL` so they won't apply any filters to the data. The [adjustment to `calls_combine_by`](https://github.com/insightsengineering/teal.slice/blob/990940a2d621e6b486e09c460014adbd57cfd77d/R/call_utils.R#L468) for example ignores any calls for a disabled filter.

R/FilterStateChoices.R
asbates: I think this is outside the scope of this issue. I've opened #218 with these comments. Is that OK?

asbates:
> I think `is_disabled` is not robust enough, we could also benefit by making `ui_sumary` and `srv_summary` more abstract.

Can you expand on what you mean by `is_disabled` not being robust enough?

asbates:
In testing using the switch with the check/X (above), the first switch in the cards doesn't render correctly. It is just slightly narrower than the remaining cards. This leads to some of the green still showing when the switch is off. If we remove the first card, the new first card switch renders fine. If we add a card for the original first switch variable, it renders fine. On Firefox everything renders fine. The issue is Chrome and Edge.

@lcd2yyz @donyunardi @gogonzo I can try some more to alleviate this but I think it will take quite a while. What do you think we should do?

lcd2yyz:
> In testing using the switch with the check/X (above), the first switch in the cards doesn't render correctly. It is just slightly narrower than the remaining cards. This leads to some of the green still showing when the switch is off. If we remove the first card, the new first card switch renders fine. If we add a card for the original first switch variable, it renders fine. On Firefox everything renders fine. The issue is Chrome and Edge.

Do you know what might be causing it? I would hate to spending time on fixing CSS, rather than developing functionality.
Does the problem resolve by itself if we switch to the round button toggle (the second option you suggested originally)? If so, let's go with that.

gogonzo (commented):
Looks like I forgot to submit my review yesterday ⬇️

R/FilterState.R
```diff
@@ -618,7 +684,7 @@ InteractiveFilterState <- R6::R6Class( # nolint
     #
     # @param id (`character(1)`) Id for the containing HTML element.
     # @param parent_id (`character(1)`) id of the FilterStates card container
-    ui_bs3 = function(id, parent_id) {
+    ui_bs3 = function(id, parent_id, enable) {
```
donyunardi, line 687: Add `@param enable ...` in the documentation.

R/FilterState.R
```diff
@@ -664,7 +738,7 @@ InteractiveFilterState <- R6::R6Class( # nolint
     #
     # @param id (`character(1)`) Id for the containing HTML element.
     # @param parent_id (`character(1)`) id of the FilterStates card container
-    ui_bs45 = function(id, parent_id) {
+    ui_bs45 = function(id, parent_id, enable) {
```
donyunardi, line 741: Add @param enable ... in the documentation.

R/FilterStateChoices.R
```diff
@@ -459,37 +461,32 @@ ChoicesFilterState <- R6::R6Class( # nolint
             }
           })
 
+          observeEvent(private$disabled(), {
```
donyunardi, line 464:
```suggestion
          observeEvent(private$is_disabled(), {
```

It looks like all other class (FilterStateDate, FilterStateDatetime) is using `private$is_disabled()` method rather than the `private$disabled` property. I propose we do the same for consistency.

donyunardi (commented):
I made some comment on some minor stuff.
Everything else looks good to me.

PR #208, opened 2023-02-23
fix for warnings and errors in test due to values comparison

BLAZEWIM:
# Pull Request

I noticed warnings (`Programmatic range specification on `) and errors in `teal.slice` tests on main coming from `FilterStateRange.R`, solved with @chlebowa 

It was due to the fact that we used `identical` / `setequal` for comparisons of double turned from integers with double etc..


chlebowa (approved):
:+1:

Issue #187, opened 2023-02-03
Implement a new API


gogonzo:
Initial work for the #138 

Check `?filter_state_api`. There are `set_filter_state`, `get_filter_state`, `remove_filter_state` and `get_filter_state`. They are exported functions which calls methods of the same name in the `FilteredData`.
They passing the instructions to the classes  to manage FilterState in the `filter_state_list`. Currently this communication is done by passing the list, for example
```r
list(
  ADSL = list( # goes to FilteredDataset and to FilterStates
    AGE = list( # refers to FilterState for AGE column
      selected = , # initial selected values
      keep_na = , # initial value for keep_na
      keep_inf = , # initial value for keep_inf
    )
  ),
  MAE = list( # goes to MAEFilteredDataset
    subjects = list( # goes to MAEFIlterStates
       ...
    )
  )
)

```

We need to introduce a different way of passing this informations through the classes. Currently nested list shown above is transfered through `FilteredData$set_filter_state -> FilteredDataset$set_filter_state -> FilterStates$set_filter_state -> FilterState$set_state`. While passed from class to class list get unnested and `FilterState`.

We want to introduce something like this, which can be easily extended in the future by new type of filters or by new "arguments" which can cause some specific behavour of the filter-state within state_list.
```r
filters = filter_settings(

    filter_var(dataname = "ADSL", varname = "ARM", choices = c("ARM A", "ARM B", "ARM C", NA)), # list
    
    filter_var(dataname = "ADSL", varname = "AGE", label = "AGE 18-60", choices = c(18, 60)),

    mae_parient_var(dataname = "MAE", varname = "gender"),

    mae_gene_subset(dataname = "MAE", experiment_name = "e1", varname = "gene.id"),

    mae_gene_select(),

   filterable = ...

  )
```

Comparing to the current API, in the new one we should be able also to specify "choices" which will limit number of choices in the initialized filter-state.

------

Arguments of the filter_var

- @param dataname (`character(1))` name of the dataset
- @param varname (`character(1))` name of the column in dataset
- @param choices (`vector`) values of the column which should define range/set of the choices
- @param selected (`vector`) values of the choices being a default/initial selection
- @param label (`character(1)`) text describing the filter

Arguments of the filter_settings
- @param ... (`filter_var`, `mae_patient_var`, `mae_gene_subset`, `mae_gene_select`)
- @param filterable (`list`) named after datasets, containing character vectors containing columns which can be chosen to filter. Same as current implementation on main.

[...]

PR #222, opened 2023-03-06
187 new filter panel api

chlebowa:
This PR contains work towards a new filter panel API as outlined in #187.

**TESTING class methods**:

1. preparations
```
# get FilteredData object on demand
get_fd <- function() {
  init_filtered_data(
    x = list(
      iris = list(dataset = iris),
      mtcars = list(dataset = mtcars),
      mae = list(dataset = miniACC)
    )
  )
}


# specify filter states (old way)
fss <- list(
  iris = list(
    "Species" = list(selected = "setosa"),
    "Sepal.Length" = list(selected = c(5, 6))
  ),
  mtcars = list(
    "disp" = list(selected = c(0, 280)),
    "cyl" = list(selected = 6)
  ),
  mae = list(
    subjects = list(
      years_to_birth = list(selected = c(30, 50), keep_na = TRUE, keep_inf = FALSE),
      vital_status = list(selected = "1", keep_na = FALSE),
      gender = list(selected = "female", keep_na = TRUE)
    ),
    RPPAArray = list(
      subset = list(
        ARRAY_TYPE = list(selected = "", keep_na = TRUE)
      )
    )
  )
)

# specify filter states (new way)
tss <- filter_settings(
  filter_var("iris", "Species", selected = "setosa"),
  filter_var("iris", "Sepal.Length", selected = c(5, 6)),
  filter_var("mtcars", "disp", selected = c(0, 280)),
  filter_var("mtcars", "cyl", selected = 6),
  filter_var("mae", "years_to_birth", selected = c(30, 50), keep_na = TRUE, keep_inf = FALSE, datalabel = "subjects", target = "y"),
  filter_var("mae", "vital_status", selected = "1", keep_na = FALSE, datalabel = "subjects", target = "y"),
  filter_var("mae", "gender", selected = "female", keep_na = TRUE, datalabel = "subjects", target = "y"),
  filter_var("mae", "ARRAY_TYPE", selected = "", keep_na = TRUE, datalabel = "RPPAArray", target = "subset")
)
```

2. convert old states specification to new one 
```
fss_new <- as.teal_slices(fss)
identical(fss, tss)
identical(fss_new, tss)
```
:point_up: this happens in `FilteredData$set_filter_state` with a warning 
Note `as.teal_slices` does not perform any validation, so a list-like filter state that specifies filters on columns of `MAE@colData` that is not wrapped as `list(MAE = list(...))` but is only `list(var = list(...))` will be interpreted as a `data.frame` filter.

3. set states as list
```
# create FilteredData
fd <- get_fd()
# apply filter states
fd$set_filter_state(fss)
# see calls
fd$get_call("iris") %>% isolate
fd$get_call("mtcars") %>% isolate
fd$get_call("mae") %>% isolate
#recover filter states
fd$get_filter_state() %>% isolate
```

4. set states as `teal_slices`
```
# create FilteredData
fd <- get_fd()
# apply filter states
fd$set_filter_state(tss)
# see calls
fd$get_call("iris") %>% isolate
fd$get_call("mtcars") %>% isolate
fd$get_call("mae") %>% isolate
#recover filter states
fd$get_filter_state() %>% isolate
```
Note that calls are not generated. This is these filters are instantiated and constructors don't know how to handle choices yet, so by default they are created with everything selected, hence no calls.

5. modify states as `teal_slices`
```
fd$set_filter_state(tss)
# see calls
fd$get_call("iris") %>% isolate
fd$get_call("mtcars") %>% isolate
fd$get_call("mae") %>% isolate
#recover filter states
fd$get_filter_state() %>% isolate
```

**TESTING wrapper functions**:
```
  x = list(
    iris = list(dataset = iris),
    mae = list(dataset = miniACC)
  )
)
fs <- filter_settings(
  filter_var("iris", "Species", selected = c("setosa", "versicolor")),
  filter_var("iris", "Sepal.Length", selected = c(5.1, 6.4)),
  filter_var("mae", "years_to_birth", selected = c(30, 50),
             keep_na = TRUE, keep_inf = FALSE, datalabel = "subjects", target = "y"),
  filter_var("mae", "vital_status", selected = "1", keep_na = FALSE, datalabel = "subjects", target = "y"),
  filter_var("mae", "gender", selected = "female", keep_na = TRUE, datalabel = "subjects", target = "y"),
  filter_var("mae", "ARRAY_TYPE", selected = "", keep_na = TRUE, datalabel = "RPPAArray", target = "subset")
)

# set initial filter state
set_filter_state(datasets, filter = fs)

fd$.__enclos_env__$private$get_filter_count() %>% isolate

# get filter state
get_filter_state(datasets)

fd$.__enclos_env__$private$get_filter_count() %>% isolate

# modify filter state
set_filter_state(
  datasets,
  filter_settings(
    filter_var("iris", "Species", selected = "setosa", keep_na = TRUE)
  )
)

fd$.__enclos_env__$private$get_filter_count() %>% isolate

# remove specific filters
remove_filter_state(
  datasets,
  filter_settings(
    filter_var("iris", "Species"),
    filter_var("mae", "years_to_birth"),
    filter_var("mae", "vital_status")
  )
)

fd$.__enclos_env__$private$get_filter_count() %>% isolate

# remove all states
clear_filter_states(datasets)

fd$.__enclos_env__$private$get_filter_count() %>% isolate

```


R/filter_state_api_NEW.R
```diff
@@ -0,0 +1,158 @@
...
+  checkmate::assert_class(tss, "teal_slices")
+  Filter(function(x) x[[feature]] == value, tss)
+}
+extract_by_feat(all_filters, "dataname", "dataname1")
+
+# get feature from all slices
+extract_feat <- function(tss, feature) {
+  checkmate::assert_class(tss, "teal_slices")
+  lapply(tss, function(x) x[[feature]])
+}
```
gogonzo, line 157: First impression is that these are not needed. Please use Filter and lapply instead in the classes - assert will be redundant as set_filter_state would have asserts already.

gogonzo: I'm fine now. I will get back to it once we have a full view on everything

R/FilterState-utils.R
```diff
@@ -78,213 +88,294 @@ init_filter_state <- function(x,
...
 #' @keywords internal
 #' @export
 init_filter_state.logical <- function(x,
                                       x_reactive = reactive(NULL),
-                                      varname,
-                                      varlabel = attr(x, "label"),
-                                      dataname = NULL,
+                                      dataname,
+                                      varname = attr(x, "label"),
+                                      choices = NULL,
```
gogonzo, line 148: What do you think?
```suggestion
                                      choices = unique(na.omit(x)),
```

R/FilterState-utils.R
```diff
@@ -78,213 +88,294 @@ init_filter_state <- function(x,
...
 #' @keywords internal
 #' @export
 init_filter_state.numeric <- function(x,
                                       x_reactive = reactive(NULL),
-                                      varname,
-                                      varlabel = attr(x, "label"),
-                                      dataname = NULL,
+                                      dataname,
+                                      varname = attr(x, "label"),
+                                      choices = NULL,
```
gogonzo, line 180: what do you think?
```suggestion
                                      choices = range(x, finite = TRUE),
```

R/FilterState-utils.R
```diff
@@ -78,213 +88,294 @@ init_filter_state <- function(x,
...
+                                      selected = NULL,
```
gogonzo, line 181:
```suggestion
                                      choices = range(x, finite = TRUE),
                                      selected = choices,
```

R/FilterState-utils.R
```diff
@@ -78,213 +88,294 @@ init_filter_state <- function(x,
...
+                                      varlabel = character(0),
```
gogonzo, line 182:
```suggestion
                                      varlabel = attr(x, "label"),
```

R/FilterState-utils.R
```diff
@@ -78,213 +88,294 @@ init_filter_state <- function(x,
...
+                                      keep_na = NULL,
```
gogonzo, line 183:
```suggestion
                                      keep_na = any(is.na(x)),
```

R/FilterState-utils.R
```diff
@@ -78,213 +88,294 @@ init_filter_state <- function(x,
...
 
 #' @keywords internal
 #' @export
 init_filter_state.factor <- function(x,
                                      x_reactive = reactive(NULL),
-                                     varname,
-                                     varlabel = attr(x, "label"),
-                                     dataname = NULL,
+                                     dataname,
+                                     varname = attr(x, "label"),
```
gogonzo, line 215: label is not a varname, probably `deparse(substitute(x))` is more accurate
```suggestion
                                     varname,
```

R/filter_state_api_NEW.R
```diff
@@ -0,0 +1,216 @@
...
+#'     "dataname1" = "varname2"
+#'   )
+#' )
+#'
+#' @export
+#' @rdname new_api
+#'
+filter_settings <- function(
+    ...,
+    slices,
```
gogonzo, line 121: I see the point but I think we can make it simpler. I don't like the idea to give user more opportunities
```suggestion
```

gogonzo: I have a proposition for you
```
arg_list <- slices
arg_list <- c(arg_list, list(exclude = excluded_varnames, count_type = private$count_type)

do.call(filter_settings, arg_list)
```

chlebowa: :+1:

R/FilterState-utils.R
chlebowa: will do :+1:

chlebowa: Does this go for `FilterState` constructors as well?

R/FilteredData.R
```diff
@@ -511,39 +504,46 @@ FilteredData <- R6::R6Class( # nolint
...
+        call. = FALSE)
+        checkmate::assert_subset(names(state), self$datanames())
+        lapply(names(state), function(dataname) {
+          logger::log_trace(
+            "FilteredData$set_filter_state initializing, dataname: { paste(names(state), collapse = ' ') }"
+          )
+
+          fdataset <- private$get_filtered_dataset(dataname = dataname)
+          dataset_state <- state[[dataname]]
+          fdataset$set_filter_state(state = dataset_state)
```
gogonzo, line 541: In case of passing old-nested-list down to FilterStates we should just convert them to `teal_slices` object. You can create utils function which convert `list->teal_slices`. Once we fully deprecate it will require less changes.

```suggestion
```

chlebowa: That would mean the conversion only happens at the top level (`FilteredData`) and everything downstream works on `teal_slices`. Is this your intent?

gogonzo: Yes. FilteredData is the only concern. Other classes are not intended to be used directly

chlebowa: All right, will do :+1:

R/FilterState-utils.R
```diff
@@ -78,213 +90,294 @@ init_filter_state <- function(x,
...
+    selected = selected,
     varlabel = varlabel,
-    dataname = dataname,
+    keep_na = keep_na,
+    keep_inf = keep_inf,
+    fixed = fixed,
     extract_type = extract_type
   )
+
+  do.call(InteractiveFilterState$new, args)
```
nikolas-burkoff, line 141: there's lots of do.call's in the PR - is there a reason to use it here? If so I'd add a code comment

chlebowa: The reason is two-fold: 1) to allow the `if`s to use the same argument list rather than typing out the full call twice, but to a different class at the beginning; 2) consistency.

nikolas-burkoff: shame you can't just do:


init_filter_state.default <- function(x, ...) {
  InteractiveFilterState$new(x, ...)
}

and not type them out at all...

chlebowa: I wasn't sure you couldn't but I have a feeling that would get us into trouble one way or another.

I'm willing to try in a different issue, though.

chlebowa: There are mode `do.call` calls in the PR, those mostly deal with variable numbers of arguments.

R/FilterStateDate.R
```diff
@@ -92,35 +97,71 @@ DateFilterState <- R6::R6Class( # nolint
...
+                          choices = NULL,
+                          selected = NULL,
+                          keep_na = NULL,
+                          keep_inf = NULL,
+                          fixed = FALSE,
+                          extras = NULL,
+                          dataname_prefixed = character(0),
                           varlabel = character(0),
-                          dataname = NULL,
                           extract_type = character(0)) {
```
gogonzo, line 147: This constructor is getting too complex in my opinion. I still think `dataname_prefixed` and `extract_type`
 can be removed. This would mean that onlu `x`, `x_reactive` and `varlabel` would be only fields not covered in the API.
 
I would change `get_call`, `get_varname_prefixed` to have `dataname` and `extract_type` and change `add_keep_na_call(filter_call)` to `get_keep_na_call(dataname, extract_type)` (and the same with `add_keep_inf_call`).
 Thanks to this `get_keep_na_call` would have the same arguments as `get_call` and would be easier to understand.
 
```suggestion
                          varlabel = character(0)) {
```

R/FilterState.R
```diff
@@ -227,17 +253,23 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
+        varname = private$varname,
+        choices = private$choices,
+        selected = private$selected(),
+        keep_na = if (!is.null(private$keep_na)) private$keep_na() else NULL,
+        keep_inf = if (!is.null(private$keep_inf)) private$keep_inf() else NULL,
+        fixed = private$fixed
       )
+      args <- append(args, private$extras)
+      args <- Filter(Negate(is.null), args)
+      do.call(filter_var, args)
```
gogonzo, line 272: I think `get_state` is exactly what we need 👍

R/FilterState.R
```diff
@@ -336,42 +368,49 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
+      if (inherits(state, "teal_slice")) {
+        if (!is.null(state$selected)) {
+          self$set_selected(state$selected)
+        }
+        if (!is.null(state$keep_na)) {
+          self$set_keep_na(state$keep_na)
+        }
+        if (!is.null(state$keep_inf)) {
+          self$set_keep_inf(state$keep_inf)
+        }
```
gogonzo, line 387: This method is not available in the abstract class I suggest to remove
```suggestion
```

and make in RangeFilterState
```r
      set_state = function(state) {
         super$set_state(state)
         if (!is.null(state$keep_inf)) {
          self$set_keep_inf(state$keep_inf)
        }
      }
```

R/filter_state_api_NEW.R
```diff
@@ -0,0 +1,350 @@
...
+#   if (!missing(master)) {
+#     slices <- lapply(slices, function(x) {
+#       x$dataname <- master
+#       x
+#     })
+#   }
+#
+#   do.call(filter_settings, as.list(slices, list(exclude = list(), count_types = "none")))
+# }
+as.teal_slices <- function(x) {
```
gogonzo, line 283: very good idea to simplify the code in the `FilteredData` class and make it easy to test separately!

gogonzo (commented):
Looks good. I'm still not convinced by the size of the constructor in the `FilterState`. I think you can remove `else` (`!is.teal_slices`) everywhere now

R/FilterStateDate.R
```diff
@@ -92,35 +97,71 @@ DateFilterState <- R6::R6Class( # nolint
...
     initialize = function(x,
                           x_reactive = reactive(NULL),
+                          dataname,
                           varname,
+                          choices = NULL,
+                          selected = NULL,
+                          keep_na = NULL,
+                          keep_inf = NULL,
+                          fixed = FALSE,
+                          extras = NULL,
```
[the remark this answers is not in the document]
chlebowa: Since `extras` are stored in one private filed I think it's more trouble than it's worth. Everything that is done in `filter_var` to handle `...` would have to be duplicated in the `FilterState` constructor and then they would be wrapped together anyway.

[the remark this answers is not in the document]
chlebowa: Done.

chlebowa:
Added testing code in PR description.

chlebowa:
Opening for review.

The file `filter_state_api_NEW.R` contains a number of utility functions, some of which can be removed.
Names of some functions/methods/arguments/fields could use tweaking.
Also, I believe all of `resolve_state` can be removed now.

There are two points I am not convinced of:
  1. How filterable variables are determined, now that `FilterStates` receives `exclude` in `teal_slices`. I am not sure whether checking for filterable variable types is still applicable (though I suppose it guards against filtering on list columns).
2. The `remove...` flow is written so that one only has to pass a `teal_slice` containing `dataname` and `varname` as these are the only two mandatory elements. This causes `FilteredDatasetMAE$remove_filter_state` to be a little ugly and convoluted, and it is vulnerable to one variable name existing in many state lists within one data set. The last one seems unlikely but still. This can be remedied by requiring that `teal_slice$datalabel` also be provided for this class but I would rather keep requirements i) consistent and ii) simple.

I would love to hear your thoughts on those issues in particular, the whole PR notwithstanding.

P.S. If you find yourself asking "why?" a lot, a talk may be helpful.

R/FilterState.R
```diff
@@ -277,6 +303,32 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
+    #'  Value(s) which come from the filter selection. Value is set in `server`
+    #'  modules after selecting check-box-input in the shiny interface. Values are set to
+    #'  `private$keep_inf` which is reactive.
+    #'
+    set_keep_inf = function(value) {
+      if (shiny::isolate(private$is_disabled())) {
+        warning("This filter state is disabled. Can not change keep Inf.")
+      } else {
+        checkmate::assert_flag(value)
+        private$keep_inf(value)
```
gogonzo, line 319: `private$keep_inf` is missing in the abstract class

R/FilterStateChoices.R
```diff
@@ -89,40 +94,74 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
+            x = x,
+            x_reactive = x_reactive,
+            dataname = dataname,
+            varname = varname,
+            choices = choices,
+            selected = selected,
+            keep_na = keep_na,
+            keep_inf = keep_inf,
+            fixed = fixed,
+            extract_type = extract_type),
```
gogonzo, line 161: styler didn't shout?
```suggestion
            extract_type = extract_type
          ),
```

R/filter_state_api_NEW.R
```diff
@@ -0,0 +1,442 @@
...
+      v <- paste(substr(v, 1, 16), "...")
+    }
+    v
+  }
+
+  ind <- intersect(names(x), names(formals(filter_var)))
+  xx <- x[ind]
+  cat("teal_slice", "\n")
+  for (i in seq_along(xx)) {
+    element_name <- format(names(xx)[i], width = name_width)
```
gogonzo, line 208: I wouldn't print if something was `NULL`
```suggestion
    if (is.null(xx[[i]])) next
    element_name <- format(names(xx)[i], width = name_width)
```

R/filter_state_api_NEW.R
```diff
@@ -0,0 +1,442 @@
...
+#' extract_fun(all_filters, dataname == "dataname2")
+#' extract_fun_s(all_filters, 'dataname == "dataname2"')
+#' x <- "dataname2"
+#' extract_fun_s(all_filters, sprintf('dataname == "%s"', x))
+#'
+#'NULL
+
+
+#' @export
+#' @rdname new_api
```
gogonzo, line 85: Also `@name` tag is missing in the docs of the `NULL` above
```suggestion
#' @rdname teal_slice_api
```

R/filter_state_api_NEW.R
```diff
@@ -0,0 +1,442 @@
...
+    }
+  }
+  cat(sprintf("\ncount type: %s", ct), "\n")
+}
+
+
+# get slices where feature is value
+#' @export
+#' @rdname new_api
+#' @keywords internal
```
gogonzo, line 382: so the export or internal ;] please remove `export` as noted in the comment about NAMESPACE

R/filter_state_api_NEW.R
gogonzo: I was thinking to export this function but It will be only temporary. Please remove `@export` tag

R/filter_panel_api.R
```diff
@@ -135,10 +69,9 @@ NULL
 #' @export
 set_filter_state <- function(datasets, filter) {
   checkmate::assert_multi_class(datasets, c("FilteredData", "FilterPanelAPI"))
-  checkmate::assert_list(filter, min.len = 0, null.ok = TRUE)
-  if (length(filter) > 0) {
-    datasets$set_filter_state(filter)
-  }
+  checkmate::assert_class(filter, "teal_slices")
```
gogonzo, line 72: we still want to support the old way
```suggestion
  checkmate::assert(
    checkmate::check_class(filter, "teal_slices"),
    checkmate::check_list(filter, min.len = 0),
    checkmate::check_null(filter)
  )
```

R/FilterStateDate.R
```diff
@@ -170,16 +208,19 @@ DateFilterState <- R6::R6Class( # nolint
...
     #' `<varname> >= <min value> & <varname> <= <max value>` with
     #' optional `is.na(<varname>)`.
+    #' @param dataname name of data set; defaults to `private$dataname`
     #' @return (`call`)
-    get_call = function() {
-      choices <- as.character(self$get_selected())
+    #'
+    get_call = function(dataname) {
+      if (missing(dataname)) dataname <- private$dataname
+      choices <- as.character(self$get_selected(dataname))
```
gogonzo, line 216: I guess this `dataname` suppose to go to `get_varname_prefixed`
```suggestion
      choices <- as.character(self$get_selected())
```

gogonzo (changes requested):
Please test with `187_new_api@filter_panel_refactor@main` in teal. Please try testing examples from [here](https://github.com/insightsengineering/teal.slice/pull/165#issue-1548941880), they are failing at this moment.

`srv_add` still creates a list which makes app fail. 

https://github.com/insightsengineering/teal.slice/blob/68f7a554821eda720864d9e535368318295e9e0b/R/FilterStatesDF.R#L480

https://github.com/insightsengineering/teal.slice/blob/68f7a554821eda720864d9e535368318295e9e0b/R/FilterStatesMAE.R#L326

https://github.com/insightsengineering/teal.slice/blob/68f7a554821eda720864d9e535368318295e9e0b/R/FilterStatesMatrix.R#L288

https://github.com/insightsengineering/teal.slice/blob/68f7a554821eda720864d9e535368318295e9e0b/R/FilterStatesSE.R#L446

https://github.com/insightsengineering/teal.slice/blob/68f7a554821eda720864d9e535368318295e9e0b/R/FilterStatesSE.R#L469

NAMESPACE
```diff
@@ -29,12 +34,22 @@ S3method(variable_types,data.frame)
...
+export(extract_fun_s)
+export(filter_settings)
+export(filter_var)
 export(get_filter_expr)
 export(get_filter_state)
 export(init_filter_states)
 export(init_filtered_data)
 export(init_filtered_dataset)
+export(is.teal_slice)
+export(is.teal_slices)
```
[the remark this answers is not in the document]
chlebowa: I know we don't expect them to be used outside of the filter panel, but aren't `is.<class>` and `as.<class>` functions usually exported?

R/filter_state_api_NEW.R
chlebowa: Adding `show_all = FALSE` argument to `print` methods just in case.

NAMESPACE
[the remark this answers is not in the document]
gogonzo: `is_teal_slices` solves the problem

R/filter_state_api_NEW.R
chlebowa: How about just `teal_slice`? Too confusing with package name?

R/filter_state_api_NEW.R
chlebowa: `@keywords internal` merely remove help pages from the index, I don't think S3 methods for generics need to appear there.

R/filter_state_api_NEW.R
gogonzo: teal_slice is 👍

R/filter_state_api_NEW.R
gogonzo: 👍

NAMESPACE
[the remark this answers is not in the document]
chlebowa: removed for `is.*` and `as.*`

chlebowa:
> Please test with `187_new_api@filter_panel_refactor@main` in teal. Please try testing examples from [here](https://github.com/insightsengineering/teal.slice/pull/165#issue-1548941880), they are failing at this moment.
> 
> `srv_add` still creates a list which makes app fail.

resolved

R/FilterStatesMAE.R
```diff
@@ -125,96 +131,88 @@ MAEFilterStates <- R6::R6Class( # nolint
...
+      logger::log_trace("{ class(self)[1] }$set_filter_state initializing, dataname: { private$dataname }")
+
+      # Drop teal_slices that refer to excluded variables.
+      varnames <- unique(unlist(extract_feat(state, "varname")))
+      filterable <- get_supported_filter_varnames(SummarizedExperiment::colData(private$data))
+      if (!all(varnames %in% filterable)) {
+        excluded_variables <- toString(dQuote(setdiff(varnames, filterable), q = FALSE))
+        state <- extract_fun_s(
+          state,
+          sprintf("!varname %%in%% c(%s)", )
```
gogonzo, line 168: sprintf is missing argument
```suggestion
          sprintf("!varname %%in%% c(%s)", )
```

R/FilterStatesDF.R
```diff
@@ -277,83 +286,70 @@ DFFilterStates <- R6::R6Class( # nolint
...
+      filterable <- private$filterable_varnames
+      if (!all(varnames %in% filterable)) {
+        excluded_variables <- toString(dQuote(setdiff(varnames, filterable), q = FALSE))
+        state <- extract_fun_s(
+          state,
+          sprintf("!varname %%in%% c(%s)", )
         )
-        logger::log_warn("Columns filters { excluded_varnames_str } were excluded from { private$dataname }")
-        state <- state[state_varnames %in% filterable_varnames]
+        logger::log_warn("filters for columns: { excluded_varnames } excluded from { private$dataname }")
```
gogonzo, line 312:
```suggestion
        logger::log_warn("filters for columns: { excluded_variables } excluded from { private$dataname }")
```

R/FilterStatesMAE.R
```diff
@@ -125,96 +131,88 @@ MAEFilterStates <- R6::R6Class( # nolint
...
+      filterable <- get_supported_filter_varnames(SummarizedExperiment::colData(private$data))
+      if (!all(varnames %in% filterable)) {
+        excluded_variables <- toString(dQuote(setdiff(varnames, filterable), q = FALSE))
+        state <- extract_fun_s(
+          state,
+          sprintf("!varname %%in%% c(%s)", excluded_variables)
         )
-        logger::log_warn("Columns filters { excluded_varnames_str } were excluded from { private$dataname }")
-        state <- state[state_varnames %in% filterable_varnames]
+        logger::log_warn("filters for columns: { excluded_varnames } excluded from { private$dataname }")
```
gogonzo, line 170:
```suggestion
        logger::log_warn("filters for columns: { excluded_variables } excluded from { private$dataname }")
```

R/FilterStatesMAE.R
```diff
@@ -324,8 +316,10 @@ MAEFilterStates <- R6::R6Class( # nolint
                   private$dataname
                 )
               )
+
               varname <- input$var_to_add
-              self$set_filter_state(setNames(list(list()), varname))
+              self$set_filter_state(filter_settings(filter_var(private$dataname, varname)))
```
gogonzo, line 321: this line throws error
```
Error in eval(expr, x) : object 'target' not found
```

R/FilterStatesMAE.R
```diff
@@ -259,21 +257,15 @@ MAEFilterStates <- R6::R6Class( # nolint
...
           # available choices to display
           avail_column_choices <- reactive({
-            choices <- setdiff(get_supported_filter_varnames(data = data), active_filter_vars())
+            active_filter_vars <- unique(unlist(extract_feat(self$get_filter_state(), "varname")))
+            choices <- setdiff(get_supported_filter_varnames(data = data), active_filter_vars)
             varlabels <- vapply(
               colnames(data),
               FUN = function(x) {
-                label <- attr(data[[x]], "label")
+                label <- attr(data[[x]], "label") # TODO: can this ever be NULL or just character(0)
```
gogonzo, line 268: If attribute isn't specified then it returns NULL. If you want this to be robust then line below we should have `if (!checkmate::test_string(label))`.

```suggestion
                label <- attr(data[[x]], "label")
```

gogonzo (commented):
get_call returns `MAE <- `to filter SE instead of `MAE[[experiment_name]] <- `. The call should use `datalabel` here

R/FilterStatesMAE.R
gogonzo: Unresolve as it's still here - `excluded_varnames` passed to logger doesn't exist as binding was changed to `excluded_variables` few lines above.

gogonzo (commented):
minor

R/FilterStatesMAE.R
chlebowa: Oops, sorry.

R/FilterState.R
```diff
@@ -90,16 +105,22 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
-        varlabel
-      }
-      private$extract_type <- extract_type
       private$selected <- reactiveVal(NULL)
+      private$keep_na <- reactiveVal(keep_na)
+      private$keep_inf <- reactiveVal(keep_inf)
+      private$extras <- list(...)
+      private$fixed <- fixed
+      # Establish varlabel.
+      varlabel <- attr(x, "label")
```
asbates, line 114: If there is no label this would be NULL I think. Would an empty string be more appropriate?

R/teal_slice.R
```diff
@@ -0,0 +1,431 @@
...
+#' Functions for passing filter state information between objects.
+#'
+#' These functions create and manage filter state specifications.
+#' A single filter state can be fully described by a `teal_slice` object and such
+#' objects will be used to create, modify, and delete filter state.
+#'
+#' A `teal_slice` contains a number of common elements (all named arguments of `filter_var`)
+#' but only `dataname` and `varname` are mandatory, while the others have default values.
+#' Setting any of the other values to NULL means that these parameters will not be modified
+#' (when setting an existing state) or that they will be determined by data (when creating new a new one).
```
asbates, line 13: I like this methodology and the explanation here.

R/teal_slice.R
```diff
@@ -0,0 +1,431 @@
...
+  }
+  cat(sprintf("\ncount type: %s", ct), "\n")
+}
+
+
+# get slices where feature is value
+#' @rdname teal_slice
+#' @keywords internal
+#'
+extract_by_feat <- function(tss, feature, value) {
```
asbates, line 376: What about 'get' instead of 'extract'? I'm also thinking something similar to purrr for the names of these functions.

- `extract_feat` -> `get_feature`
- `extract_by_feat` -> `get_feature_val`
- `extract_fun` -> `get_feature_lgl`
- `extract_fun_s` -> `get_feature_lgl_str`

R/teal_slice.R
```diff
@@ -0,0 +1,431 @@
...
+#' @param keep_na `logical(0-1)` optional logical flag specifying whether to keep missing values
+#' @param keep_inf `logical(0-1)` optional logical flag specifying whether to keep infinite values
+#' @param fixed `logical(1)` logical flag specifying whether to fix this filter state (i.e. forbid setting state)
+#' @param exclude `named list` of `character` vectors where list names match names of data sets
+#'                 and vector elements match variable names in respective data sets;
+#'                 specifies which variables are not allowed to be filtered
+#' @param count_type `character(1)` string specifying how observations are tallied by these filter states
+#' @param show_all `logical(1)` specifying whether NULL elements should also be printed
+#' @param tss `teal_slices`
+#' @param feature `character(1)` name of `teal_slice` element
```
asbates, line 54: I'm not sure about 'feature.' It feels a bit like feature in the 'variable'/'column' sense. That being said I don't really have an alternative.

R/teal_slice.R
```diff
@@ -0,0 +1,431 @@
...
+  Filter(function(x) isTRUE(eval(expr, x)), tss)
+}
+
+
+# name teal_slices according to value of respective feature in slices
+# possibly useful in set_filter_state_impl but needs safeguards
+#' @rdname teal_slice
+#' @keywords internal
+#'
+name_slices <- function(tss, feature) {
```
asbates, line 421: I don't see this used anywhere. If this is a have just in case, I'd encourage removing it. In my experience this just creates confusion down the road.

R/teal_slice.R
```diff
@@ -0,0 +1,431 @@
...
+#' teal.slice:::extract_fun_s(all_filters, sprintf('dataname == "%s"', x))
+#'
+#' @name teal_slice
+NULL
+
+
+#' @export
+#' @rdname teal_slice
+#'
+filter_var <- function(
```
asbates, line 94: I get needing 'dataname' and 'varname' from the outside perspective (initializing an app). But it feels extra from an inside perspective (e.g. in FilterStates).

What if `filter_settings` took named `teal_slice`s? Kind of like now with the lists but the right hand side is a `filter_var` instead of a list. This might need another function though to handle all filters for a single data set.

I'm thinking something more like this.
```r
    ADSL = filter_data( # filter_group maybe?
        filter_var(AGE, selected = c(0, 100)),
        filter_var(RACE)
    ),
    ADTTE = filter_data(
        filter_var(AVAL, keep_na = FALSE)
    )
)
```

Once FilteredData, FilterStates are initialized, an individual `filter_var` shouldn't care what the dataname or varname is. That part can be handled by the class via e.g. a named list where names correspond to the variable.

Otherwise if I want to set a single FilterState say AGE in ADSL, I have to call `fs$set_state(ADSL, AGE, selected = c(0, 10))`? The filter state `fs` knows what data and variable it belongs to though so it shouldn't have to specified here. I could be misunderstanding how this works though

R/FilterState.R
```diff
@@ -336,42 +398,36 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
-      logger::log_trace(sprintf(
-        "%s$set_state, dataname: %s setting state of variable %s to: selected=%s, keep_na=%s",
-        class(self)[1],
-        private$dataname,
-        private$varname,
-        paste(state$selected, collapse = " "),
-        state$keep_na
-      ))
-      stopifnot(is.list(state) && all(names(state) %in% c("selected", "keep_na")))
+      checkmate::assert_class(state, "teal_slice")
```
gogonzo, line 408: please add assert on `state$varname` to be the same as `private$varname`. Same for a `dataname`

R/teal_slice.R
chlebowa: The names of these functions are absolutely up for duscussion.

R/teal_slice.R
chlebowa: This name is also up for discussion.

R/teal_slice.R
chlebowa: Yup, some of these helpers will go away as soon as the core of the PR settles. I will go through the code and try to use as few as possible. Specifically, I don't think `extract_fun` is needed anymore as we don't have nested lists in a `teal_slice` so we don't have to use e.g. `extras$experiment == "exp_name"`.

R/FilterState.R
chlebowa: Adding assertion on `dataname` for higher levels as well.

R/teal_slice.R
chlebowa: I disagree. 

1. To my understanding the goal of the new api is to pass complete information on a single `FilterState` through all classes of the filter panel.
2. Builtding a `teal_slice` in layers would be no different from the nested lists this is meant to replace. Why stop at `dataname`? There are other elements one could be adding along the way, e.g. what is currently called `varlabel` or `target`. (Note: `exclude` and `count_type` are departures from this, as is `extract_type`).
3. Nobody is going to type these specifications other than for the sake of examples or during app development.
4. API-driven calls to `FilterState` objects are called from the top level anyway.

In the context of (4), the `FilterState`s don't care which dataset they belong to, but in the unlikely event that the same variable (`varname`) name exists in more than one dataset (`dataname`), both are needed.

Ultimately, there are no solutions, only trade-offs. I should probably frame this on a wall somewhere as keeps coming up lately. To me having to type more is a reasonable trade-off for a `teal_slice` being an atom of information. (I suppose `dataname` and `varname` sould be the nucleus.) 

Furthermore, a `FilterState` not knowing or caring what `FilterStates` it belongs to keeps it simple, which we are in dire need of.

R/FilterState-utils.R
```diff
@@ -75,216 +84,280 @@ init_filter_state <- function(x,
...
+    choices = choices,
+    selected = selected,
+    keep_na = keep_na,
+    keep_inf = keep_inf,
+    fixed = fixed,
+    extract_type = extract_type
+  )
+  args <- append(args, list(...))
+
   if (length(unique(x[!is.na(x)])) < getOption("teal.threshold_slider_vs_checkboxgroup")) {
```
chlebowa, line 195: Since we drop NAs here anyway, is there a point to having a default `choices = unique(na.omit(x))`?

R/teal_slice.R
gogonzo: @asbates @chlebowa Let's divide discussion on three parts which have been put on question (all can be independent). Point of discussion discussion is that all understand each other and understand the technical consequences of doing something in another way. If during discussion you figure out that it's okey let's merge it and continue discussion in another issue.

1. Public functions to interact with filter-panel. @asbates put recent implementation in question as `filter_data` might be a good addition.

2. Apart from (1), `teal_slices` might be divided per dataset list to avoid holding and specifying `dataname` information multiple times. This can be independent of (1) as returned `teal_slices` might be built in the same way as implemented already.

3. `FIlterState` isn't strictly tied up with `teal_slice` as initialize uses arguments specified separately and `set_state` requires `teal_slice`. Both can be viewed as inconsistent.

------

I'd like to hear others opinion about `initialize` and `set/get_state` in `FilterState` (3). Should `FilterState` return `teal_slice` or a `list` which is then turned into `teal_slice` in `FilterState`. How it's implemented on this branch is up for discussion:
- Is called on resolved `teal_slice` and `FilterState$initialize` receives all attributes in the arguments (not via teal_slice).
- `FIlterState$initialize` sets fixed features (dataname, varname, choices, fixed, extract_type, datalabel, target) and reactive features (selected, keep_na, keep_inf)
- `FilterState$set_state` accept `teal_slice` object and sets reactive features and ignores static features.
- `FIlterState$get_state` returns `teal_slice` as it holds all necessary informations. 

Alternatively:
- we can remove all `dataset` specific arguments from `init_filter_state` so that it will initialize with `varname, varlabel, choices, selected, keep_na, keep_inf`
- we can change `FilterState$set_state(state)` to accept a state=list containing `selected, keep_na, keep_inf`
- `FilterState$get_state` returns a list equivalent to `state` in `set_state`
- `FIlterStates$get_filter_state` would create `teal_slice` based on the list returned from `get_state` and on the dataset attributes (dataname, target).


-----------

My personal view - I'm fine with current implementation. There is one important reason (for me) why `FilterState$get_state` returns more informations (also irrelevant for itself). We can try to remove some of the unnecessary steps - for example `filter_state_list` holding two lists for summarized experiments.

chlebowa: > I'd like to hear others opinion about `initialize` and `set/get_state` in `FilterState` (3). Should `FilterState` return `teal_slice` or a `list` which is then turned into `teal_slice` in `FilterState`. How it's implemented on this branch is up for discussion:
> 
> * Is called on resolved `teal_slice` and `FilterState$initialize` receives all attributes in the arguments (not via teal_slice).
> * `FIlterState$initialize` sets fixed features (dataname, varname, choices, fixed, extract_type, datalabel, target) and reactive features (selected, keep_na, keep_inf)

I don't understand what you mean by this.

Full disclosure: I am biased against reversing changes I myself made in this PR.

I don't really see what is gained by dropping `dataname` from a `teal_slice` other than a little convenience. On the other hand, here are the gains I see by not dropping it:
- unambiguous clarity of `FilterState` definition
- consistency of defining state across the hierarchy
- ease of restructure the hierarchy (which has not been mentioned in this PR specifically, but was implied in others)

R/teal_slice.R
chlebowa: removed

R/teal_slice.R
chlebowa: Some history:
`extract_by_feat` was created as a Filter wrapper for the simple purpose of `get slices where this field has this value"
`extract_fun` was later added because information was stored in nested lists, so `extract_by_feat` could not directly refer to a field and things like `extras$target == "y"` had to be evaluated in a `teal_slice`
`extract_fun_s` was then added because programmatic construction of an expression is not trivial, to I went to building a string and converting that to a call within the `extract*` function

Even though there are no nested fields at the moment, the latter still holds true. Therefore, I think subsetting `teal_slices` with a string forming a logical predicate can cover all our needs and the other subsetting functions can be dropped.

How about this:
1. remove `extract_fun` and use the `extract_fun_s` in all cases
2. rename `extract_fun_s` to `slices_which` or `slices_where` (I prefer the former)
3. remove `extract_by_feat` and use `slices_which` in all cases
4. rename `feature` to `field` or `element` `extract_feat` to `slices_field`/`slices_element`

R/teal_slice.R
chlebowa: discussed in another comment

R/teal_slice.R
gogonzo: local-filter-panel-api comment drawio


Sorry for a stream of consciousness. Problematic is what I drawed above. We initialize FS by explicitly passing arguments and set_choices, set_selected directly in init. `teal_slice` object is unwrapped in `FilterStates`. While `set_state` uses `teal_slice` directly. I think we can optimize this, but it doesn't mean we need to do this in this PR. Here are some propositions:

#### 1 Initialize creates "empty" FS 

Everything is set through set_state as a single point of truth. `teal_slice` is utilized and reproduced in the FilterState


local-filter-panel-api 2 drawio (1)




#### 2 compose/decompose teal_slice in FilterStates

`teal_slice` is utilized and reproduced in `FilterStates`. Thanks to this, `FilterState` doesn't store/return anything unnecessary 

local-filter-panel-api 3 drawio


Let's discuss this tomorrow

R/FilterState.R
```diff
@@ -160,151 +128,251 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
+    ui = function(id, parent_id = "cards") {
+      ns <- NS(id)
+
+      theme <- getOption("teal.bs_theme")
+
+      if (is.null(theme)) {
+        private$ui_bs3(id, parent_id)
+      } else {
+        private$ui_bs45(id, parent_id)
+      }
```
gogonzo, line 283: ui got simplied on the filter_panel_refactor@main - please keep the ui from refactor branch as distinction `ui_bs3` and `ui_bs45` was removed. Please ask @asbates for help if needed

BLAZEWIM:
to discuss:
- how to initialize disabled
- defaults for varname, dataname, fixed in set_state: impact on set_* (additional check needed?)
- do we want to remove everything custom from init? example: private$set_choices_counts
- modifying private$x by set_choices: move to another method?
- set extras warning
- FilterStateRange to revisit
- I commented the test not to forget what was tested, but most of them uses now private methods

BLAZEWIM:
This code:

 ``` 
testfs <- R6::R6Class(
    classname = "testfs",
    inherit = DateFilterState,
    public = list(
      disable = function() private$disable(),
      enable = function() private$enable()
    )
  )
  date_seq <- seq(Sys.Date() - 2, Sys.Date(), 1)
  fs <- testfs$new(date_seq, varname = "x", dataname = "data")
  fs$set_state(filter_var(selected = date_seq[1:2], keep_na = TRUE, varname = "x", dataname = "data"))
  fs$disable()
  fs$enable()
```

produces interesting error:
```
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
Error during wrapup: evaluation nested too deeply: infinite recursion / options(expressions=)?
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
```

Apparently this `if` cannot be exited:
```
      if (!is.null(private$cache)) {
        private$disabled(FALSE)
        self$set_state(private$cache)
        private$cache <- NULL
      }
```

**Comment by BLAZEWIM** [comment]
This code:

 ``` 
testfs <- R6::R6Class(
    classname = "testfs",
    inherit = DateFilterState,
    public = list(
      disable = function() private$disable(),
      enable = function() private$enable()
    )
  )
  date_seq <- seq(Sys.Date() - 2, Sys.Date(), 1)
  fs <- testfs$new(date_seq, varname = "x", dataname = "data")
  fs$set_state(filter_var(selected = date_seq[1:2], keep_na = TRUE, varname = "x", dataname = "data"))
  fs$disable()
  fs$enable()
```

produces interesting error:
```
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
Error during wrapup: evaluation nested too deeply: infinite recursion / options(expressions=)?
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
```

Apparently this `if` cannot be exited:
```
      if (!is.null(private$cache)) {
        private$disabled(FALSE)
        self$set_state(private$cache)
        private$cache <- NULL
      }
```

chlebowa:
> This code:
> (...)
> produces interesting error:

Solved.
This occurred because `enable` called `set_state` and `set_state` called enable.
To alleviate this, I modified `enable` so that after the state is retrieved from cache, the `disabled` field in the `teal_slice` is set to NULL and thus `set_state` no longer triggers `enable` od `disable` as these are only called if `teal_slice$disabled` is either TRUE or FALSE.
See [here](https://github.com/insightsengineering/teal.slice/blob/187_new_api%40filter_panel_refactor%40main/R/FilterState.R#L632).

chlebowa:
Should constructors have default values on `keep_na` and `keep_inf`? If so, what?

R/FilterState.R
```diff
@@ -250,129 +187,79 @@ InteractiveFilterState <- R6::R6Class( # nolint
...
-            "%s$set_keep_na set for variable %s to %s.",
-            class(self)[1],
-            private$varname,
-            value
+        # Allow for enabling a filter state before altering state.
+        if (isTRUE(state$disabled)) private$disable()
+        if (isFALSE(state$disabled)) private$enable()
+
+        if (private$is_disabled()) {
+          logger::log_trace("{ this filter state is disabled: { private$dataname } { private$varname }")
```
asbates, line 206: I would expect a warning here. At the very least a `log_warn`. Attempting to set state on a disabled filter suggests something is wrong and as an app/module developer, I'd like to know about that.

tests/testthat/test-FilterState.R
```diff
@@ -1,196 +1,180 @@
 # initialize ----
-testthat::test_that("The constructor accepts character as varname", {
-  testthat::expect_no_error(InteractiveFilterState$new(c(7), varname = "test"))
-  testthat::expect_error(InteractiveFilterState$new(c(7), varname = quote(pi)))
-  testthat::expect_error(InteractiveFilterState$new(c(7), varname = call("test")))
+testthat::test_that("constructor checks arguments", {
```
asbates, line 2: I'm thinking these types of tests may not be needed. I feel like they are

1. testing `checkmate` which is already tested
2. potentially brittle

By 'brittle' I mean: let's say we allow one of the arguments to be `NA` in the future that we don't allow now. The associated test would then fail. But I think that's testing details of implementation vs. intended behavior.

In other words, I feel tests like these aren't really testing the behavior of the classes which is what I think tests should focus on.

Curious what others think.

chlebowa: Agree with (1), disagree with (2). I think testing assertions is redundant, but I was explicitly asked to do so in another project, so I thought I would go ahead, especially since input types were tested previously. And if we do test for allowed input, I _would_ want a test to fail if another type of input is allowed. This only means the tests did not permit a forbidden input in the first place.

R/FilterState.R
chlebowa: will do :+1:

chlebowa:
@asbates [this](https://github.com/insightsengineering/teal.slice/blob/187_new_api%40filter_panel_refactor%40main/tests/testthat/test-FilteredData.R#L737) stopped working after `filter_panel_refactor@main` was merged into this branch, could you have a look?

donyunardi:
Can’t we call the `disable/enable` method directly? 
This works for me:

```r
testthat::test_that("switching disable/enable button caches and restores state", {
  filtered_data <- FilteredData$new(
    list(
      iris = list(dataset = iris),
      mtcars = list(dataset = mtcars)
    )
  )
  fs <- filter_settings(
    filter_var(dataname = "iris", varname = "Sepal.Length", selected = c(5.1, 6.4), keep_na = TRUE, keep_inf = FALSE),
    filter_var(dataname = "iris", varname = "Species", selected = c("setosa", "versicolor"), keep_na = FALSE),
    filter_var(dataname = "mtcars", varname = "cyl", selected = c(4, 6), keep_na = FALSE, keep_inf = FALSE),
    filter_var(dataname = "mtcars", varname = "disp", keep_na = FALSE, keep_inf = FALSE)
  )
  filtered_data$set_filter_state(fs)
  shiny::testServer(
    filtered_data$srv_filter_panel,
    expr = {
      cached <- filtered_data$get_filter_state()
      testthat::expect_true(filtered_data$get_filter_panel_active())
      session$setInputs(filtered_data$filter_panel_disable())
      testthat::expect_null(filtered_data$get_filter_state())
      testthat::expect_false(filtered_data$get_filter_panel_active())
      testthat::expect_warning(session$setInputs(filtered_data$filter_panel_enable()), "Choices adjusted")
      testthat::expect_identical(filtered_data$get_filter_state(), cached)
      testthat::expect_true(filtered_data$get_filter_panel_active())
    }
  )
})
```

image

chlebowa:
> Can we call the `disable/enable()` method instead of accessing private field `filter_panel_active`? This works for me:

That is done in another test. It is very suspect that `setInput` did the job previously but stopped after the merge, might be a sign of a hidden flaw.

chlebowa:
So `setInputs` doesn't trigger the [relevant observer](https://github.com/insightsengineering/teal.slice/blob/7a660d725836126a8874d5ffaf81963539386389/R/FilteredData.R#L709) when `session$setInputs(filter_panel_active = FALSE)` is called. Odd, the switch works when clicked by hand.

I tried, I don't get it.

chlebowa:
The test failure was caused by the fact that the button `filter_panel_enable` was move from the module `srv_filter_panel` to `srv_active`. Even though the latter is called by the former, `setInputs` failed to find the id "filter_panel_active" in `input` and thus could not assign another value.

We consider this a bug ar at least a limitation on `setInputs`.

The test was changed to run in `srv_active` directly.

asbates (approved):

PR #223, opened 2023-03-06
Expand card click are

### LINKED ISSUE #203: Make the card reactive on a wider area
The card does not open if i press anywhere on it, it only reacts to the space in green where as the space in blue does not trigger it, it would be great to modify it such that the card presses anywhere i press on it:
IMG_0229

asbates:
Expands the clickable area of filter cards so they expand/collapse when clicking anywhere in the card header. (except remove icon)

Closes #203

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -626,21 +626,18 @@ InteractiveFilterState <- R6::R6Class( # nolint
         class = "panel panel-default",
         include_js_files("count-bar-labels.js"),
         tags$div(
-          class = "panel-heading",
+          class = "panel-heading accordion-toggle",
+          `data-toggle` = "collapse",
+          `data-parent` = paste0("#", parent_id),
+          href = paste0("#", ns("body")),
           tags$div(
             class = "panel-title",
-            tags$a(
-              class = "accordion-toggle",
-              `data-toggle` = "collapse",
-              `data-parent` = paste0("#", parent_id),
-              href = paste0("#", ns("body")),
               tags$span(tags$strong(self$get_varname())),
               if (length(self$get_varlabel())) {
                 tags$span(self$get_varlabel(), class = "filter-card-varlabel")
               } else {
                 NULL
-              }
-            ),
+              },
             actionLink(
               inputId = ns("remove"),
               label = icon("circle-xmark", lib = "font-awesome"),
@@ -672,21 +669,18 @@ InteractiveFilterState <- R6::R6Class( # nolint
         class = "card",
         include_js_files("count-bar-labels.js"),
         tags$div(
-          class = "card-header",
+          class = "card-header accordion-toggle",
+          `data-toggle` = "collapse",
+          `data-bs-toggle` = "collapse",
+          href = paste0("#", ns("body")),
           tags$div(
             class = "card-title",
-            tags$a(
-              class = "accordion-toggle",
-              `data-toggle` = "collapse",
-              `data-bs-toggle` = "collapse",
-              href = paste0("#", ns("body")),
               tags$span(tags$strong(self$get_varname())),
               if (length(self$get_varlabel())) {
                 tags$span(self$get_varlabel(), class = "filter-card-varlabel")
               } else {
                 NULL
-              }
-            ),
+              },
             actionLink(
               inputId = ns("remove"),
               label = icon("circle-xmark", lib = "font-awesome"),
--- a/inst/css/filter-panel.css
+++ b/inst/css/filter-panel.css
@@ -222,12 +222,12 @@ a.remove_all:hover {
 .panel-title,
 .card-title {
   display: flex;
-  justify-content: space-between;
 }
 
 .panel-title a,
 .card-title a {
   text-decoration: none;
+  margin-left: auto;
 }
 
 .card-title a {
@@ -240,6 +240,7 @@ a.remove_all:hover {
 
 .filter-card-varlabel {
  color: var(--bs-gray-500, var(--gray, darkgray));
+ margin-left: 0.5em;
 }
 
 .filter-card-remove,


gogonzo:
Looks good. Please merge `filter_panel_refactor@main`. I'm curious how it would work with the disable switch - if clicking disable wouldn't expand window in the same time.

BLAZEWIM:
I noticed that when you click on X to delete filter, filter expands for a moment before it is closed. While it is not a big deal in test case, it might cause some unnecessary reactivity in the whole app.

asbates:
> I noticed that when you click on X to delete filter, filter expands for a moment before it is closed. While it is not a big deal in test case, it might cause some unnecessary reactivity in the whole app.

Yes I see what you mean. That's not so great. Looking into this.

asbates:
@gogonzo @BLAZEWIM Clicking on either disable switch or remove icon toggles the panel. I think we may have to go to JavaScript to handle this though. If disable switch or remove icon are clicked, I think we need to use `event.preventDefault()` to prevent toggling and then `Shiny.setInputValue`. Of course I could be overthinking this.

asbates:
Update: there's something going on with the updateCountText JS handler for keep_na. When I click the disable switch, I get the following error.

image

The label starts out in a <span> tag with an id. When clicking disable switch, it loses the <span> tag and id. So trying to find the element to update the text is failing because the element with that id doesn't exist.

The relevance here is that I'm not able to hit the event listeners to handle the cases when the card is disabled or removed. Namely, prevent the card from collapsing/expanding.

chlebowa:
Since I now next to nothing about ront end, let me see if I understand correctly.
FilterCard buttons
* one div (white) contains the whole filter card
* one div is always visible (green)
* one div is toggled (yellow)
* the green div contains two spans (blue) and two buttons (red)
Toggling the yellow div is controlled by clicking on the green one. The problem is events emitted by the red buttons are caught by the green div and cause unwanted toggling.

Having consulted a front-ender friend of mine, I see solutions.

One is to add `event.stopPropagation` somewhere in the button handlers. This would prevent the events emitted from the buttons from propagating to their parent elements. I am not sure we can do this, since we handle the events on the R side.

The other is to remove the toggling behavior from the green div and add it to the two blue ones. It may not be elegant, but it is easy. Realistically, users would not click on buttons in order to toggle.

asbates:
Almost there. Bootstrap 3, 4 handle accordions differently than bootstrap 5. At least, the events keep firing in a different order for me. Collapse always fires first on BS 5. So we have to handle that a little differently.

Gah! As I type I'm realizing the JS [`Event.explicitOriginalTarget1`](https://developer.mozilla.org/en-US/docs/Web/API/Event/explicitOriginalTarget) isn't supported outside Firefox. I'm hoping we can check an other event property though. The issue with BS5 is we have to listen for the BS event 'collapse' on the card _body_. Otherwise the card will toggle before the e.g. disable switch listener triggers.

donyunardi:
I played around with this a bit:

https://user-images.githubusercontent.com/8597300/225145247-fe081a83-cd31-420a-8cd8-b12f2b72a682.mov

I used different input, `actionButton`, so I can add an `onClick` attribute:
```r
...
              actionButton(
                inputId = "test",
                label = NULL,
                icon = icon("star"),
                onClick = "foo(event)"
              ),
...
```

And in JS, I added this line:
```javascript
window.foo = function(e){
  e.stopPropagation()
  # code 
}
```

This solution means we need to:
1. Use different input that can take `onClick` property so we can target the element for click event. If we do this, I have a feeling we don't need to worry about the different BS version.
2. Move the disable functionality to javascript and pass the value (disable/enable) back to the app.

While it's possible, it felt a lot just to achieve our initial goal, which is to expand the click area.

If we don't expand the area, maybe we can think of another route to make it visible to the user on which area to click to expand the card (i.e. change in background color or arrow icon on the label area, etc.). @lcd2yyz what do you think of this?

lcd2yyz:
> The other is to remove the toggling behavior from the green div and add it to the two blue ones. It may not be elegant, but it is easy. Realistically, users would not click on buttons in order to toggle.

@chlebowa suggestion sounds more pragmatic to me. The schematic really explained the overlapping elements we have going on in the card. The total blue area isn't much smaller than green minus red area, and no one would consciously move the mouse to the region surrounding the red area in order to expand/collapse card. Is this simpler to implement compared to the solution @donyunardi tested? If so, let's go with this.

lcd2yyz:
Thanks @asbates for the more detailed explanation. I was comparing it to the original issue #203, which just seemed like we needed to add the second row of the blue area to reactivity. But I understand now the underlying issue is more than that.

The delete button is essentially an action button that can leverage @donyunardi's solution already. 
I would request to spend sometime to find some action button UI that can work like toggle switch. I feel being able to click anywhere on the blue area to toggle the accordion is the ideal behavior.

asbates:
Unfortunately the delete button doesn't work on Bootstrap 5 either. And I couldn't get the action button to work for me. :(

lcd2yyz:
Discussed with @donyunardi this afternoon, he explained that Bootstrap 5 isn't working when we have overlapping layers of divs. 
So can we somehow rearrange the elements so that they won't overlap?
image
The target is still to make the whole blue area clickable to get the accordion behavior. Hopefully by completely separating blue and green in nice non-overlapping rectangles can make things easier. 
Note the disable button is shown vertically in order to save space - it would be great if we can find some vertical switch (that moves up and down) or rotate our current switch 90 degrees? But not critical - we can stick with the left-right switch if nothing suitable.

asbates:
I think this is working now.

chlebowa:
I can now display multiple cards, I'm not sure if that is the current intent.
image

Disabling and closing works well :+1:

asbates:
@chlebowa can you check again please. The issue was with Bootstrap 3. I've updated and checked Bootstrap 3,4,5 on Firefox and Chrome.

chlebowa:
I can now unfold one card per data set:
image

gogonzo (approved):
👍

PR #225, opened 2023-03-07
Limited choices@187 new api@filter panel refactor@main

BLAZEWIM:
Should be reviewed together /  with regard to https://github.com/insightsengineering/teal.slice/pull/222

Closes https://github.com/insightsengineering/teal.slice/issues/232

Problems to solve:
- when initializing `FilterState` using FilterPanel API with choices, when parameter choices limits the number of possible values to select from, `is_any_filtered` should be set to TRUE to notice that the filter is applied always, therefore it must appear in the call.
- test could not run due to different errors, mostly `argument "dataname" missing`
- remove class `InteractiveFilterState` and `FilterState-abstract`, as they are not needed anymore (fixed filter will be set using `fixed` parameter)
```
state <- ChoicesFilterState$new(
  letters[c(1, 2)],
  x_reactive = reactive(NULL),
  "test",
  "var",
  choices = c("a", "b"),
  selected = c("a", "b"),
)
shiny::isolate(state$get_state())
state$.__enclos_env__$private$is_choice_limited

state <- ChoicesFilterState$new(
  letters,
  x_reactive = reactive(NULL),
  "test",
  "var",
  choices = c("a", "b"),
  selected = c("a", "b"),
)
shiny::isolate(state$get_state())
state$.__enclos_env__$private$is_choice_limited
```

# Pull Request

Part of https://github.com/insightsengineering/teal.slice/issues/187


R/FilterState.R
```diff
@@ -464,6 +463,7 @@ InteractiveFilterState <- R6::R6Class( # nolint
     dataname = character(0),
     varname = character(0),
     choices = NULL, # because each class has different choices type
+    choices_limited = FALSE, # flag whether number of possible choices was limited when specifying filter
```
gogonzo, line 466: `choices_limited` suggest that this field contains choices

```suggestion
    is_choices_limited = FALSE, # flag whether number of possible choices was limited when specifying filter
```

R/FilterStateChoices.R
```diff
@@ -168,10 +168,11 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
       x <- droplevels(x)
-      choices <- table(x)
-      private$set_choices(names(choices))
-      self$set_selected(names(choices))
-      private$set_choices_counts(unname(choices))
+      choices_table <- table(x)
+      private$set_choices(names(choices_table))
+      self$set_selected(names(choices_table))
+      private$set_choices_counts(unname(choices_table))
+      private$set_choices_limited(x, choices)
```
gogonzo: Looks like `choices` are not set anywhere. `private$choices` and `private$choices_count` are  set on values taken from `x`. As a result `FilterState` still have all choices but with the flag `choices_limited`.

What I would suggest is to filter `x` and `x_reactive` in the beginning constructor and set flag if choices are not complete. 

Also - since `x` is a vector, `choices` argument could have a default value `choices = unique(na.omit(x))` in this class.

R/FilterStateDatettime.R
```diff
@@ -162,6 +162,12 @@ DatetimeFilterState <- R6::R6Class( # nolint
       var_range <- as.POSIXct(trunc(range(x, na.rm = TRUE), units = "secs"))
       private$set_choices(var_range)
       self$set_selected(var_range)
+      if (!is.null(choices)) {
```
gogonzo, line 165: I think `choices = NULL`  shouldn't be allowed.

R/FilterStateChoices.R
```diff
@@ -168,10 +168,11 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
       x <- droplevels(x)
-      choices <- table(x)
-      private$set_choices(names(choices))
-      self$set_selected(names(choices))
-      private$set_choices_counts(unname(choices))
+      choices_table <- table(x)
+      private$set_choices(names(choices_table))
+      self$set_selected(names(choices_table))
+      private$set_choices_counts(unname(choices_table))
+      private$set_is_choice_limited(x, choices)
```
gogonzo, line 175: `choices` and `selected` argument are ignored. Constructor sets is_choice_limited without limiting anything

```r
state <- ChoicesFilterState$new(
  letters,
  x_reactive = reactive(NULL),
  "test",
  "var",
  choices = c("a", "b"),
  selected = c("a", "b"),
)

shiny::isolate(state$get_state())
# TRUE
state$.__enclos_env__$private$is_choice_limited
# returns all letters and everything is selected
```

gogonzo (changes requested):
Choices are not set from the argument and selected is not limited by this. 
Once done please edit initial PR comment to include instructions what has been done and some examples.

R/FilterStateChoices.R
```diff
@@ -163,15 +163,18 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
         x <- factor(as.character(x), levels = as.character(sort(unique(x))))
       }
       x <- droplevels(x)
-      choices <- table(x)
-      private$set_choices(names(choices))
-      self$set_selected(names(choices))
-      private$set_choices_counts(unname(choices))
+      choices_table <- table(x)
+      private$set_choices(names(choices_table))
+      self$set_selected(selected)
```
gogonzo, line 176: Since super$initialize gets `choices` and `selected`, these calls doesn't need to be here as set_choices and set_selected is done twice. This means that `super$initialize` should at the end this init.

BLAZEWIM: I see that choices are not yet used in `super$initialize`, and selected is set as `private$selected <- reactiveVal(NULL)`. So we can skip this line, adjust choices and call `super$initialize` at the end, right? And same goes for other types?

gogonzo: If `FilterState$initialize` has argument `choices` and `selected` but doesn't use them then it's wrong

gogonzo: `set_selected` in all subclasses calls `super$set_selected`. It means you can easily remove them all. Then it might be clear that `set_selected` is an abstract process and probably should be moved to the abstract class.

https://github.com/insightsengineering/teal.slice/blob/3ee6bdd60cac3b8db5c497c621127226ec1acd4c/R/FilterStateRange.R#L323

R/FilterStateChoices.R
```diff
@@ -171,15 +171,18 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
+      private$set_choices_counts(unname(choices_table))
```
gogonzo, line 185: looks like `private$set_choices_count` is the only "setter" which suppose to be called in ChoicesFilterState

gogonzo (commented):
This is the list of the allowed operations in the constructors of the child classes:
- adjusting choices to the relevant values for this class
- adjusting default selected (if not specified) for a given choices
- passing adjusted values to the super$initialize
- setting only these private fields which are specific for this class:
   - filtered_counts (ChoicesFilterState)

```
private$set_is_choice_limited(choices, x)
private$set_choices(choices)
private$set_selected(selected)
```

R/FilterStateDate.R
```diff
@@ -134,14 +134,26 @@ DateFilterState <- R6::R6Class( # nolint
...
+      if (is.null(choices)) {
+        choices <- range(x, na.rm = TRUE)
+      } else {
+        private$set_is_choice_limited(x, choices)
+        x <- x[x >= choices[1L] & x <= choices[2L]]
+        choices <- c(max(choices[1L], min(x)) , min(choices[2L], max(x)))
+      }
+
+      if (is.null(selected)) selected <- choices
+      selected <- c(max(selected[1], min(choices)) , min(selected[2], max(choices)))
```
gogonzo, line 155: handled by set_selected
```suggestion
```

R/FilterStateDatettime.R
```diff
@@ -137,14 +137,32 @@ DatetimeFilterState <- R6::R6Class( # nolint
...
+          as.POSIXct(trunc(x, units = "secs")) >= choices[1L] &
+            as.POSIXct(trunc(x, units = "secs")) <= choices[2L]
+        ]
+        choices <- c(
+          max(choices[1L], min(as.POSIXct(x), na.rm = TRUE)),
+          min(choices[2L], max(as.POSIXct(x), na.rm = TRUE))
+        )
+      }
+      if (is.null(selected)) selected <- choices
+      selected <- c(max(selected[1], min(choices)) , min(selected[2], max(choices)))
```
gogonzo, line 164:
```suggestion
```

R/FilterStateRange.R
```diff
@@ -141,8 +140,7 @@ RangeFilterState <- R6::R6Class( # nolint
                           ...) {
       checkmate::assert_numeric(x, all.missing = FALSE)
       if (!any(is.finite(x))) stop("\"x\" contains no finite values")
-
-      # validation on x_reactive here
+      checkmate::assert_class(x_reactive, 'reactive')
 
       do.call(
```
gogonzo, line 145: `super$initialize` is still in the beginning of the constructor body. It doesn't correspond with #222 where super class sets `keep_inf` and I see this to be potential error as `private$keep_inf` is prepared too late (see the line 164 below). I guess you need to synchronize your actions with @chlebowa

gogonzo (commented):
RangeFilterState failing 

```
filter_state <- teal.slice:::RangeFilterState$new(
  x = c(seq(1:10), NA, Inf),
  varname = "x",
  dataname = "data",
  keep_inf = TRUE
)
# Error in private$cast_and_validate(value) : 
#  The array of set values must have length two.
```

R/FilterStateRange.R
BLAZEWIM: `RangeFilterState` was not pushed yet, working on it now.

R/FilterStateDate.R
```diff
@@ -209,47 +218,30 @@ DateFilterState <- R6::R6Class( # nolint
     #' For this class returned call looks like
     #' `<varname> >= <min value> & <varname> <= <max value>` with
     #' optional `is.na(<varname>)`.
-    #' @param dataname name of data set; defaults to `private$dataname`
     #' @return (`call`)
     #'
-    get_call = function(dataname) {
-      if (missing(dataname)) dataname <- private$dataname
+    get_call = function() {
```
gogonzo, line 223:
```suggestion
    get_call = function(dataname) {
```

R/FilterStateDate.R
```diff
@@ -109,7 +109,7 @@ DateFilterState <- R6::R6Class( # nolint
     #'   if `extract_type` argument is not empty.
     #' @param varname (`character(1)`)\cr
     #'   name of the variable.
-    #' @param choices (`atomic`, `NULL`)\cr
+    #' @param choices (`vector`, c(min(x, na.rm = TRUE), max(x, na.rm = TRUE)))\cr
```
chlebowa, line 112: I don't think I understand this specification here. The help should only list allowed types of input, not values, should it? If this is to be a default value, it should be put in the definition, not the help. Also, `range`.

```suggestion
    #' @param choices (`vector`, `range(x, na.rm = TRUE)`)\cr
```

R/FilterStateDate.R
```diff
@@ -134,14 +134,25 @@ DateFilterState <- R6::R6Class( # nolint
...
                           selected = NULL,
-                          keep_na = NULL,
-                          keep_inf = NULL,
+                          keep_na = FALSE,
+                          keep_inf = FALSE,
                           fixed = FALSE,
                           extract_type = character(0),
                           ...) {
-      checkmate::assert_date(x)
+      stopifnot(is(x, "Date"))
```
chlebowa, line 142: what's wrong with `assert_date`?

R/FilterStateDatettime.R
```diff
@@ -112,7 +112,7 @@ DatetimeFilterState <- R6::R6Class( # nolint
     #'   if `extract_type` argument is not empty.
     #' @param varname (`character(1)`)\cr
     #'   name of the variable.
-    #' @param choices (`atomic`, `NULL`)\cr
+    #' @param choices (`vector`, c(min(x, na.rm = TRUE), max(x, na.rm = TRUE)))\cr
```
chlebowa, line 115: same as `DateFilterState`

R/FilterStateRange.R
```diff
@@ -134,55 +133,38 @@ RangeFilterState <- R6::R6Class( # nolint
...
-            keep_inf = keep_inf,
-            fixed = fixed,
-            extract_type = extract_type
-          ),
-          list(...)
-        )
-      )
+      checkmate::assert_numeric(choices, null.ok = TRUE)
+      if (!any(is.finite(x))) stop("\"x\" contains no finite values")
+      checkmate::assert_class(x_reactive, 'reactive')
```
chlebowa, line 144: Please keep assertions in the same order as formals.

R/FilterStateRange.R
```diff
@@ -134,55 +133,38 @@ RangeFilterState <- R6::R6Class( # nolint
...
 
       private$is_integer <- checkmate::test_integerish(x)
       private$keep_inf <- reactiveVal(keep_inf)
       private$inf_filtered_count <- reactive(
         if (!is.null(private$x_reactive())) sum(is.infinite(private$x_reactive()))
       )
       private$inf_count <- sum(is.infinite(x))
-
+      if (is.null(choices)) choices <- c(min(na.omit(x[is.finite(x)])), max(na.omit(x[is.finite(x)])))
```
chlebowa, line 152: use `range` function

R/FilterStateEmpty.R
```diff
@@ -135,32 +134,6 @@ EmptyFilterState <- R6::R6Class( # nolint
...
-    set_state = function(state) {
-      if (!is.null(state$selected)) {
-        stop(
-          sprintf(
-            "All values in variable '%s' are `NA`. Unable to apply filter values \n  %s",
-            private$varname,
-            paste(state$selected, collapse = ", ")
-          )
-        )
-      }
```
chlebowa, line 159: This method gives a very specific error message, are we sure we want to get rid of it?

chlebowa (commented):
(I only checked R files.)

I think default values in `FilterState$initialize` should be in line with `filter_var` defaults. The constructors are called by `FilterStates$set_filter_state_impl` by taking values directly from a `teal_slice`.

Is it all right for constructors of `FilterState` subclasses to first set private fields and then call `super$initialize`? Will the fields not be reset?

R/FilterStateEmpty.R
chlebowa: Oh, that makes sense :+1:

R/FilterStateDate.R
```diff
@@ -134,14 +134,25 @@ DateFilterState <- R6::R6Class( # nolint
...
       checkmate::assert_date(x)
+      checkmate::assert_class(x_reactive, 'reactive')
+      checkmate::assert_multi_class(choices, c("Date"), null.ok = TRUE)
+
+      if (is.null(choices)) {
+        choices <- range(x, na.rm = TRUE)
+      } else {
+        private$set_is_choice_limited(x, choices)
+        x <- x[x >= choices[1L] & x <= choices[2L]]
+        choices <- c(max(choices[1L], min(x)) , min(choices[2L], max(x)))
```
gogonzo, line 151: please throw warning when choices are changed to available values.

R/FilterStateLogical.R
```diff
@@ -153,28 +170,6 @@ LogicalFilterState <- R6::R6Class( # nolint
...
-        if (levels(df) %in% c(TRUE, FALSE)) {
-          choices_not_included <- c(TRUE, FALSE)[!c(TRUE, FALSE) %in% levels(df)]
-          levels(df) <- c(levels(df), choices_not_included)
-        }
-      }
-      tbl <- table(df)
-      choices <- as.logical(names(tbl))
-
-      private$set_choices(choices)
-      private$set_choices_counts(as.integer(tbl))
```
gogonzo, line 167: set_choices_counts is missing and it probably causes app crash

Issue #206, opened 2023-02-22
When disabling filter panel, don't remove UI


asbates:
When the whole filter panel is disabled, inputs should be disabled instead of removing the entire card UI. When #15 is implemented the mechanism will exist to disable at the individual card level. So it shouldn't be difficult to adapt the current process.

@gogonzo update:
- When disabling `FIlterState` we should avoid caching and restoring. Instead `is_any_filtered` should return `FALSE` when `disabled`. We don't need caching/restoring and emptying `state` which contains `disable` information in the new API - so we can distinguish which `state` are enabled and which are disabled

donyunardi:
### Acceptance Criteria:
* Not delete the inputs when global panel is disabled
* Instead, with the new API, we can send signal to disable each cards
* May need to modify class

chlebowa:
A question has come up as to what exactly is the function the global disable switch. We see two possibilities:
1) when set to `disable`, disable all filter cards, and when set to `enable`, enable all filter cards
2) when set to `disable`, disable all filter cards, and when set to `enable`, restore all filter cards to their previous state, including disabled status

@lcd2yyz, can you comment?

lcd2yyz:
> 2\. when set to `disable`, disable all filter cards, and when set to `enable`, restore all filter cards to their previous state, including disabled status

Option 2 makes more sense.

lcd2yyz:
> The functionality of the global and individual switches would be different in case 2 so I think it would be good to distinguish them visually.

Can you explain more about the difference that you see about the global vs individual card switch?
To me, the toggle would apply the same behavior of activate filter (enable) ->inactivate filter (disable) -> restoring previous state before disabling (re-enable). The only difference would be the level which it controls, so potentially the global switch will need to "know" many more filter states before it was inactivated.

From UX perspective, having a visual consistency of the UI component is important. So unless they are fundamentally different, I would prefer to keep the same look for the toggle switch

asbates:
Yes I agree visual consistency is important. But I think if things are visually the same, they should have the same functionality.

The difference with the global enable/disable is that it's not enabling/disabling. For example if I have on individual card disabled, click the global disable switch, then global enable, the individual filter card is still disabled. So I clicked enable, but a filter is disabled.

On the individual filter perspective, the switch truly enables and disables. Disable means that input is disabled and the filter is not applied to the data. When enabled, the input is enabled and the filter is applied to the data.

chlebowa:
Here's a new hurdle:
[...]

PR #228, opened 2023-03-08
Move filter panel switch to active vars

### LINKED ISSUE #219: Move the global filter enable/disable button to Active Filter Variables card
> I like this one better, the check/cross mark coupled with the coloring will ensure no ambiguity. But consistency is also important - was there any discussion when selecting the toggle button for the entire filter panel or just chosen by whoever implemented it at the time? If the latter, can we switch the button for entire panel to be the same as above?
> 
> Also, on a more inquisitive look, does it actually make sense to place the filter panel toggle on the upper right corner? Perhaps it would it be more logical to place it inside the "Active Filter Variable" panel instead? image

@lcd2yyz I can't say whether there was a discussion or not for the main panel switch, I don't know. As far as location, yes I think it makes more sense in the Active Filter Variables section as well. I think this is outside the scope of the current issue so I opened #217. @donyunardi can we get this added on an upcoming sprint please.

_Originally posted by @asbates in https://github.com/insightsengineering/teal.slice/issues/207#issuecomment-1448452690_

asbates:
Moves the filter panel disable/enable switch into the active variables section.

Closes #219


nikolas-burkoff:
Once you've disabled the filter panel you can still add a new filter which then appears - which is confusing

image

and then when you re-enable the filter panel the other filter that was there has vanished?

asbates:
> Once you've disabled the filter panel you can still add a new filter which then appears - which is confusing
> 
> image
> 
> and then when you re-enable the filter panel the other filter that was there has vanished?
> 
> Edit: With the switch in its old location disabling the filter panel also disabled the "Add filter" panel as well

This should be fixed now.

nikolas-burkoff (commented):
2 points:

- The collapse icons can be incorrect if you disable the filter panel when one of the dataset filters is collapsed:

image

- If you have disabled and enabled filters:

image

And then disable and enable the filter panel it doesn't remember the state of the filter panel when it was disabled:

image

Assuming this is the desired behavio(u)r then this is fine.

asbates:
>     * The collapse icons can be incorrect if you disable the filter panel when one of the dataset filters is collapsed:
> 
 I have no idea why because I didn't change anything around these as far as I know. I'll look into it.

>     * If you have disabled and enabled filters: 
> And then disable and enable the filter panel it doesn't remember the state of the filter panel when it was disabled:
> 
> Assuming this is the desired behavio(u)r then this is fine.

This is not desired behavior, at least to me. Actually, I think we shouldn't remove the filter cards at all (#206) which is the root of the problem. However, I don't think it's within the scope of this PR.

asbates:
>     * The collapse icons can be incorrect if you disable the filter panel when one of the dataset filters is collapsed:

This happens on filter panel refactor branch as well so I opened #239

nikolas-burkoff (approved):
Given that we have #206 and #239 I think this can go in 👍

PR #243, opened 2023-04-11
Review teal.slice documentation before release

BLAZEWIM:
Fixes #[238](https://github.com/insightsengineering/teal.slice/issues/238)

### DIFF
--- a/R/calls_combine_by.R
+++ b/R/calls_combine_by.R
@@ -14,7 +14,7 @@
 #' \dontrun{
 #' calls <- list(
 #'   quote(SEX == "F"), # subsetting on factor
-#'   quote(AGE >= 20 & AGE <= 50), # subsetting on range
+#'   quote(AGE >= 30 & AGE <= 50), # subsetting on range
 #'   quote(!SURV) # subsetting on logical
 #' )
 #' calls_combine_by(calls, "&")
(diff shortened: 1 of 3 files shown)


R/calls_combine_by.R
```diff
@@ -12,13 +12,21 @@
...
-#'   call_condition_choice("SEX", "F"),
-#'   call_condition_range("AGE", c(20, 50)),
-#'   call_condition_choice("ARM", "ARM: A"),
-#'   call_condition_logical("SURV", TRUE)
+#'   shiny::isolate(fs_ch$get_call()),
+#'   shiny::isolate(fs_rng$get_call()),
+#'   shiny::isolate(fs_log$get_call())
 #' )
-#' calls <- lapply(calls, str2lang)
+#'
```
chlebowa, line 29: How about something simpler?
```
calls <- list(
  quote(SEX == "F),               # subsetting on factor
  quote(AGE >= 20 & AGE <= 50),   # subsetting on range
  quote(!SURV)                    # subsetting on logical
)
```

It's more abstract but is enough to understand this particular function.

chlebowa (approved):

R/calls_combine_by.R
chlebowa: Would also save us the trouble of changing this example again when #187 is merged.

BLAZEWIM: Last argument is decisive

Issue #220, opened 2023-03-03
Filter card summary design


asbates:
Review and improve the design of the filter card summary.

Some comments from @nikolas-burkoff 

Arrow would be better for displaying selected range of numeric columns.

![image](https://user-images.githubusercontent.com/36482959/222779533-5413ec6e-e734-4f38-98e3-96411893193c.png)

If there aren't any NA/Inf maybe we shouldn't show NA/Inf

![image](https://user-images.githubusercontent.com/36482959/222779605-ffbdf370-7b19-40c3-bc92-1a32e1c2430d.png)

lcd2yyz:
Also would like to add some conditional handling in the summary for categorical variables (checkbox groups UI)
- If all available levels are selected (basically no filtering applied), then display "All levels selected". 
- If only 1-2 levels are selected (for variable with more than 2 levels available), display the full selected value (eg. `M, F` for variable SEX)
- If >=3 or more levels selected, then keep display as-is

![image](https://user-images.githubusercontent.com/78242757/224216741-cdd8a307-103b-49d2-acca-ff239dbbcac2.png)

asbates:
For NA/Inf, I don't think displaying just "NA" or "Inf" conveys much information. What I mean is I don't think "NA" conveys to the user that missing values are kept. We should rethink not just the alignment here but how the message is conveyed.

lcd2yyz:
Yes agree, I thought it was something still WIP, so I didn't comment. 
So can we go back to this? Originally posted to #181 
![image](https://user-images.githubusercontent.com/78242757/225787479-6d1f669b-9775-4216-afc9-93bfde93ef26.png)

We can right align the text to show `Includes: NA, Inf` with a check or cross mark behind the NA/Inf. If there are no NA/Inf values in the data, then don't show in "Includes"

donyunardi:
###  Levels selected
> Also would like to add some conditional handling in the summary for categorical variables (checkbox groups UI)
> 
> * If all available levels are selected (basically no filtering applied), then display "All levels selected".
> * If only 1-2 levels are selected (for variable with more than 2 levels available), display the full selected value (eg. `M, F` for variable SEX)
> * If >=3 or more levels selected, then keep display as-is

[...]

PR #244, opened 2023-04-17
141 implement fixed filter state

### LINKED ISSUE #141: [Feature Request]: Implement fixed filter feature
blocked by #187 

### Feature description

We need to modify FIlterState class to have a `fixed` feature (set via api) which makes this filter non-interactive.
When `FIlterState` has property `fixed = TRUE` then it doesn't display interactive inputs but only the fixed state.

1. Property can be set via `filter_var(..., fixed = TRUE)`
2. `get_call` always return the same call
2. `ui` of such `FilterState` should display only `varname`, `label`, `dataname` and a resulting condition. 

Note: label of the fixed filter isn't the same as varlabel - it is a specific name describing the condition. For example if condition is `SEX == "F"` then label assigned would be `"Females"`. Discuss with @donyunardi and @lcd2yyz how labels describing filters should be displayed. In the current design this `label` probably should be replace what's printed in the summary.

Note: when for conditions refering to multiple we will use FilterGroup which hasn't been implemented yet.

chlebowa:
Closes #141.

`FilterState` has a `private$fixed` field that holds a logical flag. In this PR:
* a fixed `FilterState` has a _lock_ icon added to the UI
* a fixed `FilterState` does not display the filter summary part of the filter card
* a fixed `FilterState` displays a simplified version of the subsetting expression; this is a temporary measure, in the future this will display the metadata used to create the fixed filter state
* `FilterState$enable` is rewritten to accommodate for the need to disable/enable a fixed filter state


R/FilterState.R
```diff
@@ -602,7 +614,9 @@ FilterState <- R6::R6Class( # nolint
       if (!is.null(private$cache)) {
         state <- private$cache
         state$disabled <- NULL
+        private$fixed <- FALSE
```
asbates, line 617: I'm confused by this. Why do we set `fixed = FALSE` and then `fixed = TRUE`? I thought fixed was supposed to remain constant.

asbates (changes requested):
I might be misunderstanding what "fixed" means in terms of filters. I thought it meant "this value will never change". So if I am an app author I want to set say `SAFFL == 'Y'` and the user should not be allowed to change this. If my understanding is correct, then we need an adjustment to disabling.


Using the running CDISC example application, I set a filter `filter_var("ADSL", "SAFFL", selected = "Y", fixed = TRUE)`. If I then click the disable switch for the card, `selected` changes to `NULL`. Clicking again sets `selected = 'Y'`.

If this behavior is correct I think we need to tweak the UI. Because the "input" UI, in this case the progress bar, doesn't indicate to a user the filter is not applied, if the disable button is off. In other words when we do have inputs we disable them. I think in the fixed filter case we should do something similar to indicate the status.

asbates:
If fixed does mean "user can never change in any way" then what if we don't display the disable/enable switch and remove icon? I think this would be more clear to the user vs. having a switch that doesn't work like the others. And I think implementation would be simpler because `disable/enable` would never be called if fixed?

donyunardi:
To my understanding, `fixed`  means:
* User can't remove the filter interactively (by clicking the 'x' button on the filter card) during the session
* User can't change the value related to the filter state
  * either shiny input won't be available for the filter card
  * or we gray out the inputs, the same way when we disable a filter card.
* User can still enable/disable the filter card if needed by clicking the filter card toggle switch, however, the initial state should be enable.

@lcd2yyz Could you please review this and give us your thought?

lcd2yyz:
Thanks for the summary @donyunardi , let me go through the items one-by-one

> * User can't remove the filter interactively (by clicking the 'x' button on the filter card) during the session

This one is dependent on the level of "" I think there needs to be a few different levels of "locking" in terms of removal interactivity.
1) For normal teal app (what we have and what we are developing for, right now), agree to above - user should still be able to remove the filter by clicking on 'x' button. The fixed filter is only referring to fixing the selected value with the attached label (meaning). But users can decide to apply or not apply, enable or disable as they explore data through the app.
2) But for future development, thinking about creating regulatory-facing app and chevron-to-teal approach, we need to also have a way to completely lock all interactivity (ie. cannot change, cannot disable, cannot remove). We do not need to implement this right now, but should take into consideration for design and have a "placeholder" (I thought this was what the initial proposal for a separate fixed filter class was trying to enable, but maybe I was wrong?).

> * User can't change the value related to the filter state
>   * either shiny input won't be available for the filter card
>   * or we gray out the inputs, the same way when we disable a filter card.

Since the fixed filter can only be defined/created via API code, I think it has to be the former. 
There is no need to show the progress bar or counts, as we don't need the "exploration" stage. We only need to display the filter condition, like suggested in https://github.com/insightsengineering/teal.slice/issues/141#issuecomment-1452309177
image

> * User can still enable/disable the filter card if needed by clicking the filter card toggle switch, however, the initial state should be enable.

Considering for now just the case 1 (as described above) - yes agree.

R/FilterState.R
```diff
@@ -283,6 +286,11 @@ FilterState <- R6::R6Class( # nolint
             `data-toggle` = "collapse",
             `data-bs-toggle` = "collapse",
             href = paste0("#", ns("body")),
+            if (private$fixed) {
+              actionButton("dummy", label = NULL, icon = icon("lock"), class = "dummy-button")
```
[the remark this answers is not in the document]
chlebowa: Oh, that's possible? :sweat_smile: Sure, that will be better.

R/FilterState.R
chlebowa: This is so that `set_state` can proceed. Normally it aborts if `fixed = TRUE`.

I agree this it not very neat and I am open to discussion.

chlebowa:
My understanding was a disabled state can still be disabled or removed. Both of these can be changed with not much effort, if only we agree on the goal.


This is not a good idea as the _condition_ need not be as short and sweet as this one.

I am not against removing the card body if `fixed = TRUE`. What are others' opinions?

chlebowa:
> If this behavior is correct I think we need to tweak the UI. Because the "input" UI, in this case the progress bar, doesn't indicate to a user the filter is not applied, if the disable button is off. In other words when we do have inputs we disable them. I think in the fixed filter case we should do something similar to indicate the status.

Is the "Disabled" in the header not sufficient?

On this note, I am a little worried that if checkboxes are not there the user cannot tell which choices are selected by the progress bars alone.

asbates:
> > If this behavior is correct I think we need to tweak the UI. Because the "input" UI, in this case the progress bar, doesn't indicate to a user the filter is not applied, if the disable button is off. In other words when we do have inputs we disable them. I think in the fixed filter case we should do something similar to indicate the status.
> 
> Is the "Disabled" in the header not sufficient?
> 
> On this note, I am a little worried that if checkboxes are not there the user cannot tell which choices are selected by the progress bars alone.

I was thinking something more along the lines of what Chendi suggested above. I agree that with a lot of levels selected, long level names, ..., it can be unwieldy.

I very much like the idea of `$format()`. We could put that into a `<p>` tag and it should format nicely. And it would be consistent across filter/variable types which would be nice.

asbates:
> Thanks for the summary @donyunardi , let me go through the items one-by-one
> 
> > * User can't remove the filter interactively (by clicking the 'x' button on the filter card) during the session
> 
> This one is dependent on the level of "" I think there needs to be a few different levels of "locking" in terms of removal interactivity.
> 
>     1. For normal teal app (what we have and what we are developing for, right now), agree to above - user should still be able to remove the filter by clicking on 'x' button. The fixed filter is only referring to fixing the selected value with the attached label (meaning). But users can decide to apply or not apply, enable or disable as they explore data through the app.
> 
>     2. But for future development, thinking about creating regulatory-facing app and chevron-to-teal approach, we need to also have a way to completely lock all interactivity (ie. cannot change, cannot disable, cannot remove). We do not need to implement this right now, but should take into consideration for design and have a "placeholder" (I thought this was what the initial proposal for a separate fixed filter class was trying to enable, but maybe I was wrong?).
> 
> 
> > * User can't change the value related to the filter state
> >   
> >   * either shiny input won't be available for the filter card
> >   * or we gray out the inputs, the same way when we disable a filter card.
> 
> Since the fixed filter can only be defined/created via API code, I think it has to be the former. There is no need to show the progress bar or counts, as we don't need the "exploration" stage. We only need to display the filter condition, like suggested in [#141 (comment)](https://github.com/insightsengineering/teal.slice/issues/141#issuecomment-1452309177) image
> 
> > * User can still enable/disable the filter card if needed by clicking the filter card toggle switch, however, the initial state should be enable.
> 
> Considering for now just the case 1 (as described above) - yes agree.

I think the lock is a bit misleading here. To me it implies nothing can change about this filter which is what I thought fixed means. Perhaps for the future item 2. we can call it a "locked" filter? But if for a "fixed" filter we can in fact change some things, I wouldn't consider it locked. 

I don't have a specific icon suggestion yet but I'll look for a few. I do think it would be good to visually distinguish regular vs. fixed filters. But a lock feels to strong for me.

chlebowa:
Reworked the the filter card according to @lcd2yyz design in #220.

Since we don't currently pass `citril` specifications to filter states, existing features will be used as placeholders. This is reflected in `private$get_citril_*` methods.

lcd2yyz:
> I do think it would be good to visually distinguish regular vs. fixed filters. But a lock feels to strong for me.

I feel the current design already able to distinguish the different usage quite well
- regular interactive filters = no symbol, have switch, have X remove button
- fixed filter (normal case 1 usage, cannot change value, but can disable/remove) = padlock symbol, have switch, have X remove button
- completely "locked" filter (case 2 usage, cannot change value, cannot disable/remove) = padlock symbol, no switch, no removal button (something to implement in the future)

Happy to hear any other suggestions if you find some more suitable symbol though.

BLAZEWIM:
My opinion on the X button of the fixed filter: the problem with it is that after clicking X, the filter is destroyed and it cannot be restored. Also, it is destroyed when you "X" all filters, which can be a source of other problems, and limits the significance of fixed filter. If you can save and load filters, this is fine (fixed filter will always be available in the load section), but the question remains: why you might need to "destroy" the filter that was initially created for you for some reason?

Consider the case when fixed filter is removed by accident by the user with "X" all button, which will happen often. What you do now? You need to restart the app.

chlebowa:
> My opinion on the X button of the fixed filter: the problem with it is that after clicking X, the filter is destroyed and it cannot be restored. Also, it is destroyed when you "X" all filters, which can be a source of other problems, and limits the significance of fixed filter. If you can save and load filters, this is fine (fixed filter will always be available in the load section), but the question remains: why you might need to "destroy" the filter that was initially created for you for some reason?
> 
> Consider the case when fixed filter is removed by accident by the user with "X" all button, which will happen often. What you do now? You need to restart the app.

I wholeheartedly agree.

chlebowa:
Following design discussion, I changed "citril" to "metadata" - in my mind this will describe a filter group in the future but we'll cross that bridge when we get to it. Note that metadata will only be displayed on fixed filter cards. See non-cdisc example app.

R/FilterState.R
```diff
@@ -530,6 +569,23 @@ FilterState <- R6::R6Class( # nolint
         }
       str2lang(ans)
     },
+    get_metadata_id = function() {
+      private$metadata_id
+    },
+    get_metadata_title = function() {
+      if (identical(private$metadata_title, character(0))) {
+        private$get_varname()
```
donyunardi, line 577: Label is missing


This update will fix it:
```suggestion
        private$get_varlabel()
```

R/FilterState-utils.R
```diff
@@ -75,6 +78,9 @@ init_filter_state <- function(x,
                               fixed = FALSE,
                               disabled = FALSE,
                               extract_type = character(0),
+                              metadata_id = character(0),
+                              metadata_title = character(0),
```
donyunardi, line 82: I need to make mental note that `metadata_id` is the variable name and `metadata_title` is the variable label.
I sometime thinks that `metadata_title` is the variable name.

Should we name this `metadata_varname` and `metadata_varlabel` instead to make it consistent with `varname` and `varlabel` variables that's already being used in other places?

Or is there any better naming convention?

donyunardi (changes requested):

donyunardi:
We also probably need to add `metadata_id`, `metadata_title`, and `metadata_condition` as parameters here?
https://github.com/insightsengineering/teal.slice/blob/6710cdf2a1edc4704a694ce95ed721519058632e/R/teal_slice.R#L107-L116

R/FilterState.R
```diff
@@ -530,6 +569,23 @@ FilterState <- R6::R6Class( # nolint
         }
       str2lang(ans)
     },
+    get_metadata_id = function() {
+      private$metadata_id
```
donyunardi, line 573: To make it consistent with `get_metadata_title`:

```suggestion
      if (identical(private$metadata_id, character(0))) {
        private$get_varname()
      } else {
        private$metadata_id
      }
```

R/FilterState-utils.R
chlebowa: On second thought, the metadata items are not actually the varname and varlabel. Since we are only using varname and varlabel as fall-back values, I think it's better to have their names better reflect what they are.

lcd2yyz:
> after clicking X, the filter is destroyed and it cannot be restored

Ideal behavior is clicking on X does not "destroy" the fixed filter, but removed from the filter panel of the active module. It should still sit in a queue at backend, and ready to be restored (or loaded to other modules) if desired. I think this would be needed for future steps of enabling module-specific filters.

> If you can save and load filters, this is fine (fixed filter will always be available in the load section)

Yes this is the idea, it should be easily accessed by app users, without going through variable selections again.

chlebowa:
> > after clicking X, the filter is destroyed and it cannot be restored
> 
> Ideal behavior is clicking on X does not "destroy" the fixed filter, but removed from the filter panel of the active module. It should still sit in a queue at backend, and ready to be restored (or loaded to other modules) if desired. I think this would be needed for future steps of enabling module-specific filters.

FYI, there is currently no mechanism and no UI element to restore a filter state that has been removed (by clicking on the x icon). This remove-restore behavior sounds like what we have implemented as disable/enable.

I know in the future it is planned that the filter panel holds many filter states defined by the app dev or the user that can be applied at will to modules, but currently the x icon does mean destroy.

chlebowa:
In order to avoid mission creep, we decided to limit this PR to introducing and handling fixed filter states. Metadata properties will be removed and pushed off to a separate issue.

R/FilterState-utils.R
donyunardi: resolving because we'll handle the metadata stuff on other issue.

donyunardi (changes requested):
Small changes about the method name.

Also, .Rd file should be updated for these classes since we're updating the default values.
* DateFilterState
* DatetimeFilterState
* RangeFilterState

Or, we can let the CICD picks this up. =)

Other than that, everything else looks good to me.

R/FilterStateEmpty.R
```diff
@@ -154,10 +154,16 @@ EmptyFilterState <- R6::R6Class( # nolint
         )
       )
     },
+
     set_choices = function(choices) {
       private$choices <- choices
       invisible(NULL)
     },
+
+    get_metadata_condition = function() {
```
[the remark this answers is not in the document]
chlebowa: indeed!

asbates (approved):

chlebowa:
All discussed review comments have been applied. There is one recurring error in unit tests that I cannot resolve, help will be appreciated. (It can be found by searching for "todo" or by running `devtools::test()`).

Otherwise ready to merge.

R/FilterStateDatettime.R
```diff
@@ -289,6 +292,14 @@ DatetimeFilterState <- R6::R6Class( # nolint
       private$is_choice_limited <- (any(xl < choices[1L], na.rm = TRUE) | any(xl > choices[2L], na.rm = TRUE))
       invisible(NULL)
     },
+
+    get_metadata_expression = function(dataname) {
+      if (missing(dataname)) dataname <- private$dataname
+      varname <- private$get_varname_prefixed(dataname)
+      selected <- private$get_selected()
+      sprintf("%1$s => %2$s & %1$s <= %3$s", deparse1(varname), selected[1], selected[2])
+    },
```
gogonzo, line 301: what is the difference between this and `get_call`?

R/FilterState.R
```diff
@@ -321,7 +323,11 @@ FilterState <- R6::R6Class( # nolint
           `data-bs-parent` = paste0("#", parent_id),
           tags$div(
             class = "filter-card-body",
-            private$ui_inputs(ns("inputs"))
+            if (private$fixed) {
+              tags$span(private$get_metadata_expression())
```
gogonzo, line 327:
```suggestion
              tags$span(deparse1(isolate(self$get_call()))))
```

asbates:
> All discussed review comments have been applied. There is one recurring error in unit tests that I cannot resolve, help will be appreciated. (It can be found by searching for "todo" or by running `devtools::test()`).
> 
> Otherwise ready to merge.

For `test-MAEFilterStates` -> "get_filter_state returns `teal_slices` with features identical those used to set state" at least, the problem is `exclude` is a named list in one case and unnamed in the other. 

image


`names(attributes(fs_out)[['exclude']])` is `characte(0)` but it still technically has a name.

R/FilterStateDatettime.R
chlebowa: no Inf and NA information

chlebowa:
Oh, so I misinterpreted the error message and I thought it referred to item 1 of attributes, not the names attribute. Right, all clear, many thanks Andrew.

Since Dawid is reworking the `exclude` argument in another PR, I will skip these tests for the time being.

asbates (approved):

donyunardi (approved):

R/FilterState.R
```diff
@@ -582,32 +597,85 @@ FilterState <- R6::R6Class( # nolint
...
+    #
+    # Filter state properties are packed into a `teal_slice` and saved in cache (private field).
+    # State properties that refer to selection are set to NULL.
+    #
     # @return `NULL` invisibly
     disable = function() {
-      private$cache <- self$get_state()
+      logger::log_trace("{ class(self)[1] }$set_state disabling fiter state of variable: { private$varname }")
+
+      private$cache_state()
```
gogonzo, line 609: It seems like caching and restoring doesn't really make sense. Instead of all these methods we can just add condition in `get_call`. So:

- when disabled returned call is NULL so the data won't be filtered
- when disabled returned state will be returned with all fields but with flag `disabled = TRUE`

R/FilterStateChoices.R
```diff
@@ -345,6 +349,17 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
+      if (missing(dataname)) dataname <- private$dataname
+      varname <- private$get_varname_prefixed(dataname)
+      selected <- private$get_selected()
+      if (length(selected) == 1L) {
+        sprintf("%s == %s", deparse1(varname), selected)
+      } else {
+        sprintf("%s %%in%% c(%s)", deparse1(varname), toString(selected))
+      }
+    },
+
```
gogonzo, line 362:
```suggestion
```

gogonzo (commented):
Looks stunning! 
- Don't agree with default keep_na and keep_inf. 
- I think caching/restoring not needed

R/FilterState.R
chlebowa: I will overhaul disabling/enabling in #206, this is getting out of scope.

donyunardi: #206 felt like very UI driven to me which would still be out of scope since the changes is more around back-end functionality. I suggest we create a new issue instead if the suggested update is outside of the scope.

chlebowa: Oh, #206 is most certainly back-end driven. In the current implementation `filter_state_disable` caches and deletes all filter states, and `filter_state_enable` re-instantiates filter states based on the cached properties.

PR #247, opened 2023-04-17
Initialize FilterState based on count type

### LINKED ISSUE #188: Implement `reactive_counts` option to the `filter_settings`
Could be done after https://github.com/insightsengineering/teal.slice/issues/187 

In the `filter_settings` we want the option `reactive_counts = character(1)` to enable/disable recalculation of the count labels. In cases when `datasets` are big, we would like to avoid the situation where all active filters are recomputed based on the filtered-values. When `reactive_counts = FALSE` the count bars will show only raw-data-counts as we used to present before.

Make sure that when `reactive_counts = "none"` initializes `FilterState` with the `x_reactive = reactive(NULL)`

Note: `reactive_counts` should be a character describing type of counts. IN the near future we will think more about "hierarchical" reactive counts which are straightforward now to implement (on the server side).

asbates:
If the `count_type` of a `teal_slices` object (`filter_settings`) is "none", then initialize `FilterState` objects with `x_reactive = reactive(NULL)`.

I feel like there's more that needs to be done but I don't know what. Am I missing something here?

@chlebowa most of the work for this you did with the API change so I think this is all that is needed here?

Fixes #188

### DIFF
--- a/R/FilterStates.R
+++ b/R/FilterStates.R
@@ -71,7 +71,7 @@ FilterStates <- R6::R6Class( # nolint
                           dataname,
                           datalabel = character(0),
                           excluded_varnames = character(0),
-                          count_type = c("none", "all", "hierarchical")) {
+                          count_type = c("all", "none")) {
       checkmate::assert_function(data_reactive, args = "sid")
       checkmate::assert_string(dataname)
       checkmate::assert_character(datalabel, max.len = 1, any.missing = FALSE)
--- a/R/FilterStates.R
+++ b/R/FilterStates.R
@@ -622,7 +622,11 @@ FilterStates <- R6::R6Class( # nolint
           # and this no longer needs to be a function to pass sid. reactive in the FilterState
           # is also beneficial as it can be cached and retriger filter counts only if
           # returned vector is different.
-          x_reactive = reactive(data_reactive(sid)[, x$varname, drop = TRUE]),
+          x_reactive = if (attr(state, "count_type") == "none") {
+            reactive(NULL)
+          } else {
+            reactive(data_reactive(sid)[, x$varname, drop = TRUE])
+          },
           extract_type = extract_type
         )
         arg_list <- append(arg_list, x)
--- a/tests/testthat/test-teal_slice.R
+++ b/tests/testthat/test-teal_slice.R
@@ -109,7 +109,10 @@ testthat::test_that("filter_settings checks arguments", {
   testthat::expect_error(filter_settings(fs1, fs2, exclude = "fs1"), "Assertion on 'exclude' failed")
 
   testthat::expect_error(filter_settings(fs1, fs2, count_type = "fs1"), "'arg' should be")
-  testthat::expect_error(filter_settings(fs1, fs2, count_type = c("all", "none")), "'arg' must be of length 1")
+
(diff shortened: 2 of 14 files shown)


chlebowa:
The changes you made do ensure that `x_reactive = reactive(NULL)` but so far every `teal_slices` was created with the default `filter_settings(..., count_type = "none")`.

It seems this change alone removes the reactive counts implemented by Dawid previously. Perhaps the reactive counts are supposed to be covered by another value of `count_type`.

I have to admit I don't quite understand.

asbates:
Marked as blocked until we get more input from Dawid, back next week.

R/FilterStates.R
```diff
@@ -622,7 +622,11 @@ FilterStates <- R6::R6Class( # nolint
           # and this no longer needs to be a function to pass sid. reactive in the FilterState
           # is also beneficial as it can be cached and retriger filter counts only if
           # returned vector is different.
-          x_reactive = reactive(data_reactive(sid)[, x$varname, drop = TRUE]),
+          x_reactive = if (attr(state, "count_type") == "none") {
+            reactive(NULL)
+          } else {
+            reactive(data_reactive(sid)[, x$varname, drop = TRUE])
+          },
```
chlebowa, line 629:
```suggestion
          x_reactive = 
          if (attr(state, "count_type") == "none") {
            reactive(NULL)
          } else if (attr(state, "count_type") == "all") {
            reactive(data_reactive(sid)[, x$varname, drop = TRUE])
          } else {
            stop("hierarchical filters are not supported yet")
          },
```

chlebowa: Adding handling of remaining cases, after consulting with @gogonzo.
We can now regulate the default behavior by changing the default value of `count_type` in `filter_settings`.

chlebowa:
@gogonzo is working out a bug in constructing label text.

gogonzo:
I've pushed changes. See the cdisc example in #165 and test `count_type = "none"`.

gogonzo:
W

asbates:
> I've pushed changes. See the cdisc example in #165 and test.
> 
>     * When `count_type = "none"` is set then counts look like this
> 
> 
> image
> 
>     * By default `count_type` is `"all"`.
> 
>     * I've removed option to set `"hierarchical"` as I don't think it's possible to implement this very soon.

Looks good!

tests/testthat/test-teal_slice.R
```diff
@@ -109,7 +109,10 @@ testthat::test_that("filter_settings checks arguments", {
   testthat::expect_error(filter_settings(fs1, fs2, exclude = "fs1"), "Assertion on 'exclude' failed")
 
   testthat::expect_error(filter_settings(fs1, fs2, count_type = "fs1"), "'arg' should be")
-  testthat::expect_error(filter_settings(fs1, fs2, count_type = c("all", "none")), "'arg' must be of length 1")
+
+  testthat::expect_error(filter_settings(fs1, fs2, count_type = c("a", "b")), "'arg' must be of length 1")
+
+  testthat::expect_no_error(filter_settings(fs1, fs2, count_type = c("all", "none")))
```
chlebowa, line 115: this line can be dropped as it is equivalent to 104 above

chlebowa (approved):
:+1:

PR #253, opened 2023-04-25
remove slices_which

gogonzo:
Removes unnecessary `slices_which` function

### DIFF
--- a/R/FilterStates.R
+++ b/R/FilterStates.R
@@ -597,18 +597,18 @@ FilterStates <- R6::R6Class( # nolint
 
       # Modify existing filter states.
       state_list <- shiny::isolate(private$state_list_get(state_list_index))
-      slices_for_update <- slices_which(
-        state,
-        sprintf("varname %%in%% c(%s)", toString(dQuote(names(state_list), q = FALSE)))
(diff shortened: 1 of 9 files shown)


gogonzo:
@asbates I'd say that `slices_which` was unnecessary added. `slices_which` is a wrapper around `Filter` with confusing way of specifying logical condition (through string). Having `Filter` simplifies the code in my opinion, here is the example:

image

chlebowa:
`slices_which` was added as syntactic sugar, but the need to accommodate programmaticly defined logical predicates caused the usage to go out of hand somewhat. This wouldn't be the case if we only ever needed `slices_which(state, varname == "thisVar")`, then the function would be able to use NSE, but as it is, it is not possible.

I agree, with the current needs of the code `Filter` is the simpler way to go. However, **a** function like `slices_which` would neatly round up the API. Is it possible to modify it so that the use is less laborious rather than remove it outright? If not, I agree to remove, albeit with some reluctance.

Unit tests need removing as well.

gogonzo:
I'm closing this up as branch got outdated and I got numbers of errors here. Let's tackle this later.

PR #261, opened 2023-05-01
Fix bug in ChoicesFilterState count labels

### LINKED ISSUE #260: bug in count labels crashes app
When running the CDISC example app on `filter_panel_refactor@main`, I get this error when adding a filter on RACE from ADSL:

image

asbates:
Fixes a bug in the count labels for `ChoicesFilterState`. Uses the x with any unused levels removed to set choices counts.

Fixes #260

### DIFF
--- a/R/FilterStateChoices.R
+++ b/R/FilterStateChoices.R
@@ -177,7 +177,7 @@ ChoicesFilterState <- R6::R6Class( # nolint
         private$tzone <- Find(function(x) x != "", attr(as.POSIXlt(x), "tzone"))
       }
 
-      private$set_choices_counts(unname(table(x)))
+      private$set_choices_counts(unname(table(x_factor)))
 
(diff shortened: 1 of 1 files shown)


donyunardi (approved):
Just to give context, the error comes from this line:
https://github.com/insightsengineering/teal.slice/blob/c52af31f020df6d7d954f8fca2b31d38bdef9912/R/FilterStateChoices.R#L426-L435

`countsmax` is longer than `countsnow`:
image

`labels` length is longer than `private$choices`:
image

Looking at class `ChoicesFilterState`, the length of `private$choices_count` is greater than `private$choices`
image

I also looked at the data and it does look like ADSL has unused levels, which is probably why droplevels was done in `choices`:
image

The fix is to assign the right `choices_count` to match the `choices` with droplevels.
LGTM.

PR #262, opened 2023-05-02
add filter_expr

gogonzo:
1. introduced `filter_expr` which creates an object  of two classes `teal_slice_expr` and `teal_slice`. One can specify new filter_expr as follows:
```r
filter_expr(id = "AD", title = "ADULTS", dataname = "ADSL", expr = quote(SEX %in% c("F", "M") & AGE >= 18L))
```
2. introduced `FilterStateExpr`
3. introduced new way to handle the `id` attribute. It was needed as currently `FilterStates` kept in `state_list` are distinguished by `varname` and `teal_slice_expr` has no `varname` property. I decide to have an `id` as a name of the elements in the `state_list`. This `id` is a hash of fields (`dataname`, `varname`, `datalabel`, `arg`, `id`). This also partially fixed problem of flagging filters by id https://github.com/insightsengineering/teal.slice/issues/137 
4. removed `state_list` having two elements in the `SEFilterStates` as they were divided by `datalabel` and in the same time `datalabel` was passed to the `FilterState`.
5. Thanks to (4) I removed several methods from subclases as they were duplicated or almost duplicated.
6. instead of having method `get_prefixed_dataname` i moved this logic to the constructor which sets `private$dataname_prefixed`.
7. `expr` passed to `filter_expr` is checked. I think this check will never be robust enough as there might be `any.pkg::any_func(x)` which returns logical condition. I would remove `assert_logical_expression`. Please comment what you think.

### cdisc example 
```r
library(shiny)
library(scda)
library(scda.2022)
library(teal.data)
library(teal.transform)
library(teal.slice)
library(teal)
library(teal.modules.general)

  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      output$filter_states <-  renderPrint({
        filter_panel_api %>% get_filter_state()
      })
        attr(data, "code")()
      })
    }
  )
}

ADSL$empty <- NA
ADSL$logical1 <- FALSE
ADSL$numeric <- rnorm(nrow(ADSL))
ADSL$categorical2 <- sample(letters[1:10], size = nrow(ADSL), replace = TRUE)
ADSL$date2 <- rep(Sys.Date() + 1:3, length.out = nrow(ADSL))
ADSL$datetime2 <- rep(Sys.time() + 1:3 * 43200, length.out = nrow(ADSL))

ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- Inf
ADSL$logical[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$date[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$datetime[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$categorical2[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$categorical[sample(1:nrow(ADSL), size = 10)] <- NA

ADTTE$numeric <- rnorm(nrow(ADTTE))
ADTTE$numeric[sample(1:nrow(ADTTE), size = 10,)] <- NA

  filter_var(dataname = "ADSL", varname = "categorical", selected = c("a", "b"), disabled = TRUE),
  filter_var(dataname = "ADSL", varname = "categorical2", selected = c("a", "b"), disabled = TRUE),
  filter_var(dataname = "ADSL", varname = "numeric", selected = c(0, 140), keep_na = TRUE, keep_inf = TRUE, disabled = TRUE),
  filter_var(dataname = "ADSL", varname = "logical", selected = c(T), keep_na = TRUE, keep_inf = TRUE, disabled = TRUE),
  filter_var(dataname = "ADSL", varname = "datetime", disabled = TRUE),
  filter_var(dataname = "ADSL", varname = "date2", disabled = TRUE),
  filter_expr(id = "AF", title = "ADULT FEMALE", dataname = "ADSL", expr = "SEX %in% 'F' & AGE >= 18L", disabled = TRUE),
  filter_expr(id = "SE", title = "Safety-Evaluable", dataname = "ADSL", expr = "SAFFL == 'Y'", disabled = TRUE),
  filter_var(dataname = "ADSL", varname = "COUNTRY", selected = c("USA", "CAN", "JPN"), fixed = T, disabled = TRUE),  
  count_type = "all",
  # include_varnames = list(ADSL = c("SEX", "categorical2", "numeric", "logical", "date", "datetime", "date2", "datetime2", "COUNTRY")),
  
  
  exclude_varnames = list(
    ADTTE = intersect(colnames(ADSL), colnames(ADTTE)),
    ADRS = colnames(ADSL)
  )
)

data <- cdisc_data(
  cdisc_dataset("ADSL", ADSL),
  cdisc_dataset("ADRS", ADRS)
)

app <- init(
  data = data,
  modules = list(
    tm_data_table(
      variables_selected = list(ADSL = c("STUDYID", "USUBJID", "SUBJID", "SITEID", "AGE", "SEX")),
      dt_args = list(caption = "ADSL Table Caption")
    ),
    funny_module()
  ),
  filter = default_filters
)

shinyApp(app$ui, app$server)
```

### mae example
```r
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
        filter_panel_api |> get_filter_state() |> format()
      })
        attr(data, "code")()
      })
    }
  )
}
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
#options("teal.bs_theme" = bslib::bs_theme(version = 5))
#options(shiny.trace = TRUE)
devtools::load_all("teal")
library(scda)
library(SummarizedExperiment)

  filter_expr(id = "F", title = "females", dataname = "MAE", datalabel = "subjects", expr = "MAE$gender == 'female'"),
  filter_var(dataname = "MAE", datalabel = "subjects", varname = "race"),
  filter_var(dataname = "MAE", datalabel = "RPPAArray", arg = "subset", varname = "Protein", selected = TRUE),
  filter_var(dataname = "MAE", datalabel = "RPPAArray", arg = "subset", varname = "Genes"),
  include_varnames = list(MAE = c("race", "gender")),
  count_type = "none"
)

MAE <- miniACC
data <- teal.data::teal_data(teal.data::dataset("MAE", MAE))

app <- init(
  data = data,
  filter = default_filters
)

runApp(app)

```


R/FilterStateExpr.R
```diff
@@ -0,0 +1,350 @@
...
+FilterStateExpr <- R6::R6Class( # nolint
+  classname = "FilterStateExpr",
+  # public methods ----
+  public = list(
+    #' @description
+    #' Initialize a `FilterStateExpr` object
+    #' @param id (`character(1)`)\cr
+    #'   identifier of the filter
+    #' @param title (`reactive`)\cr
+    #'   title of the filter
```
chlebowa, line 25: Seems like this should be `character` but perhaps it should be stored in a `reactiveVal`?

R/FilterStateExpr.R
```diff
@@ -0,0 +1,350 @@
...
+    #' @return `FilterStateExpr`
+    initialize = function(id, title, dataname, expr, disabled = FALSE, ...) {
+      checkmate::assert_string(id)
+      checkmate::assert_string(title)
+      checkmate::assert_string(dataname)
+      assert_logical_expr(expr)
+      checkmate::assert_flag(disabled)
+
+      private$id <- id
+      private$title <- title
```
chlebowa, line 90: `private$title <- reactiveVal(title)`?

R/FilterStateExpr.R
```diff
@@ -0,0 +1,350 @@
...
+    #'
+    #' @return A `teal_slice` object.
+    #'
+    get_state = function() {
+      states <- append(
+        list(
+          id = private$id,
+          title = private$title,
+          dataname = private$dataname,
+          expr = quote(private$expr),
```
chlebowa, line 136: Since `private$expr` stores `expr` and `expr` must be an expression, is `quote` necessary?

R/FilterStateExpr.R
```diff
@@ -0,0 +1,350 @@
...
+      invisible(NULL)
+    },
+
+    # Enables `FilterState`
+    # Cached `state` is reset again and cache is cleared.
+    # @return `NULL` invisibly
+    enable = function() {
+      private$disabled(FALSE)
+      invisible(NULL)
+    },
```
chlebowa, line 309: Are these one-liner methods necessary if used only by `set_state`? I guess the same question can be asked for `FilterState`.

R/FilterStateExpr.R
gogonzo: good point, I'll fix it

R/FilterStateExpr.R
gogonzo: It must be "quotes" as `expr` is stored as language, and `do.call` would try evaluate this if not quoted.

R/FilterStates.R
```diff
@@ -65,55 +47,74 @@ FilterStates <- R6::R6Class( # nolint
...
                           data_reactive = function(sid = "") NULL,
                           dataname,
-                          datalabel = character(0)) {
-      checkmate::assert_function(data_reactive, args = "sid")
+                          datalabel = NULL) {
       checkmate::assert_string(dataname)
-      checkmate::assert_character(datalabel, max.len = 1, any.missing = FALSE)
+      logger::log_trace("Instantiating { class(self)[1] }, dataname: { dataname }")
+      checkmate::assert_function(data_reactive, args = "sid")
+      checkmate::assert_string(datalabel, na.ok = FALSE, null.ok = TRUE)
```
chlebowa, line 54: `na.ok = FALSE` is default

R/FilterStates.R
```diff
@@ -65,55 +47,74 @@ FilterStates <- R6::R6Class( # nolint
...
     #'
-    #' Builds \emph{subset expression} from condition calls stored in `FilterState`
-    #' objects selection. The `lhs` of the expression is `private$dataname`.
-    #' The `rhs` is a call to `self$get_fun()` with `private$dataname`
-    #' as argument and a list of condition calls from `FilterState` objects
-    #' stored in `private$state_list`.
-    #' If no filters are applied, `NULL` is returned to avoid no-op calls such as `x <- x`.
+    #' Builds \emph{subset expression} from condition calls generated by `FilterState`.
+    #' The `lhs` of the expression is a `dataname_prefixed`, where word prefixed refers to
+    #' situation when call is using some slots of the original data, for example `dataname[[slot]]`.
```
chlebowa, line 89: These are not "slots" in OOP sense, please avoid misleading verbiage. Slots of an MAE are: ExperimentList, colData, sampleMap, drops, and metadata.

R/FilterStateExpr.R
```diff
@@ -0,0 +1,350 @@
...
+    #' @description
+    #' Answers the question of whether the current settings and values selected actually filters out any values.
+    #' `FALSE` only when disabled.
+    #' @return logical scalar
+    is_any_filtered = function() {
+      if (isTRUE(private$is_disabled())) {
+        FALSE
+      } else {
+        TRUE
+      }
```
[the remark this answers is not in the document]
gogonzo: Yup, we will not even try to determine this. We can consider this when some users depend it.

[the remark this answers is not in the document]
gogonzo: Yup, we will not even try to determine this. We can consider this when some users depend it.

R/teal_slice.R
```diff
@@ -149,6 +158,22 @@ filter_var <- function(
   ans
 }
 
+#' @export
+#' @rdname teal_slice
+filter_expr <- function(id, title, dataname, expr, disabled = FALSE, ...) {
```
chlebowa, line 163: reorder formals so that `dataname` is always first?

R/teal_slice.R
```diff
@@ -218,11 +243,12 @@ c.teal_slice <- function(...) {
 format.teal_slice <- function(x, show_all = FALSE, ...) {
   name_width <- max(nchar(names(x)))
   format_value <- function(v) {
-    if (is.null(v)) {
+    v <- if (is.null(v)) {
```
chlebowa, line 246: assigning a `return` call?

R/utils.R
```diff
@@ -125,3 +125,25 @@ include_js_files <- function(pattern) {
...
 }
+
+
+#' Assert logical expression
+#'
+#' Assert logical expression. If object is not a logical expression then error is thrown.
+#' @param x anything
+#' @return NULL or error
+#' @keywords internal
+assert_logical_expr <-  function(x) {
```
chlebowa, line 136: Perhaps add a return value of `invisible(x) to conform to `checkmate::assert_*` functions?

asbates (changes requested):
It doesn't look like filter expressions are applied to the data. If you add a module `teal.modules.general::tm_data_table(variables_selected = list("ADSL" = c("SEX", "AGE")))` the table doesn't change when disabling and enabling the filter expression card.



For the MAE example I'm getting a warning and error on launch.

image


 For  the warning (`varname = "Protein"`), `FilterStateChoices$remove_out_of_bounds_values()` is getting a value of `"TRUE"` which is not one of `private$choices`. MAE is a mystery to me so this is as far as I could get.

asbates:
This may have been done in #206 and it's perhaps tangential to this PR so we can move to an issue if needed.

When a filter state is disabled, it does not reflect that. What I mean is if I disable a filter state, I can run `get_call` or `get_selected` for example and they will return values the same as a non-disabled filter state. This is confusing because we can't tell whether that filter is actually being applied or not.

Actually it looks to me like a filter is still being applied even if it's disabled. If we take the CDISC example above and put a browser statement in say `get_format`. If I disable `categorical2` and then check in the debugger, `private$x` and `private$x_reactive` are not the same. `private$x_reactive` is shorter which tells me that the filter is being applied.

Now, this may all be handled at the `FilterStates`, `FilteredDataset`, or `FilteredData` level in the sense that a data table module say, will display the correct data. But this is confusing. Maybe this discrepancy isn't exposed to module/app authors which I think is the case. But it's still confusing for developers. I honestly don't understand right now if filters are actually being applied correctly and where that happens. By 'correct' here I mean whether the data frame in a module accurately reflects the state of a `FilterState`. If I see a column is being filtered in the corresponding `FilterState` but it's `FilteredDataset` doesn't have that filter applied, that is confusing and hard to chase down unless you already know why and where the difference exists.

gogonzo:
@asbates 

> This may have been done in #206 and it's perhaps tangential to this PR so we can move to an issue if needed.
> 
> When a filter state is disabled, it does not reflect that. What I mean is if I disable a filter state, I can run `get_call` or `get_selected` for example and they will return values the same as a non-disabled filter state. This is confusing because we can't tell whether that filter is actually being applied or not.
> 
> Actually it looks to me like a filter is still being applied even if it's disabled. If we take the CDISC example above and put a browser statement in say `get_format`. If I disable `categorical2` and then check in the debugger, `private$x` and `private$x_reactive` are not the same. `private$x_reactive` is shorter which tells me that the filter is being applied.

`x_reactive` is a `x` after applying other filters. Other filters are independent on "this" filter, so disabling "this" filter has no effect on `x_reactive


> Now, this may all be handled at the `FilterStates`, `FilteredDataset`, or `FilteredData` level in the sense that a data table module say, will display the correct data. But this is confusing. Maybe this discrepancy isn't exposed to module/app authors which I think is the case. But it's still confusing for developers. I honestly don't understand right now if filters are actually being applied correctly and where that happens. By 'correct' here I mean whether the data frame in a module accurately reflects the state of a `FilterState`. If I see a column is being filtered in the corresponding `FilterState` but it's `FilteredDataset` doesn't have that filter applied, that is confusing and hard to chase down unless you already know why and where the difference exists.

`FilterState` and `FilterStateExpr` return call always regardless of `disabled`. There is method `is_any_filtered` which tells if `call` should be applied or not - Responsibility is on [FilterStates](https://github.com/insightsengineering/teal.slice/blob/6392b3c5067168ef47120a77ade90f92a22c7285/R/FilterStates.R#L126) which loops through all `FilterState` and decide which should be returned in combined call. I agree this is misleading mechanism and I'm happy to change `FilterState$get_call` to return `NULL` when `is_any_filtered` - I'll change this in next commit @chlebowa

chlebowa:
> `FilterState` and `FilterStateExpr` return call always regardless of `disabled`. There is method `is_any_filtered` which tells if `call` should be applied or not - Responsibility is on [FilterStates](https://github.com/insightsengineering/teal.slice/blob/6392b3c5067168ef47120a77ade90f92a22c7285/R/FilterStates.R#L126) which loops through all `FilterState` and decide which should be returned in combined call. I agree this is misleading mechanism and I'm happy to change `FilterState$get_call` to return `NULL` when `is_any_filtered` - I'll change this in next commit @chlebowa

I would prefer `$get_call` check for `$is_disabled`, cut out the middle man.

Since we're talking about `$is_any_filtered`. `FilterState` stores `x` and `x_reactive`, so we can determine the extent of the data in full and limited by other filter states, as well as limited by the present filter state (e.g. for a numeric `x`, the currently set limits). With that information, could we not determine " whether any observations are rejected" by evaluating the call? We could then have `$get_call` return NULL if no filtering occurs in this `FilterState` and we would not need `$is_any_filtered` at all. (This can be a separate issue or a poor idea.)

gogonzo:
> I would prefer `$get_call` check for `$is_disabled`, cut out the middle man.

If I would add `if (is_disabled) return(NULL)` in `get_call` why not to go further and just use `if (!is_any_filtered) return(NULL)`. We could turn `is_any_filtered` into private and not bother in FIlterStates if something is redundant or not?

> Since we're talking about $is_any_filtered. FilterState stores x and x_reactive, so we can determine the extent of the data in full and limited by other filter states, as well as limited by the present filter state (e.g. for a numeric x, the currently set limits). With that information, could we not determine " whether any observations are rejected" by evaluating the call? We could then have $get_call return NULL if no filtering occurs in this FilterState and we would not need $is_any_filtered at all. (This can be a separate issue or a poor idea.)

Evaluating the call on `x` and/or `x_reactive` to drop redundant filter expressions is ineffective way of obtaining simple thing. Analysing `x_reactive` with would be an unpredictable mechanism (as x_reactive is failrly complicated). Current `is_any_filtered` gives us this functionality almost for free.

chlebowa:
The MAE example breaks on startup and starts greyed-out.

```
[TRACE] 2023-05-09 12:40:11.0040 pid:73803 token:[8e05cac6] teal.slice FilteredData$set_filter_state initializing
[TRACE] 2023-05-09 12:40:11.0130 pid:73803 token:[8e05cac6] teal.slice MAEFilteredDataset$set_filter_state initializing, dataname: MAE
[TRACE] 2023-05-09 12:40:11.0233 pid:73803 token:[8e05cac6] teal.slice MAEFilterStates$set_filter_state initializing, dataname: MAE
Warning: Error in eval: object 'datalabel' not found
  1: runApp
```

gogonzo:
@chlebowa @asbates I'll fix MAE tomorrow, please don't bother yet. If I don't figure out something smart I will have to replace `expr` argument with `string` instead of `call`. Using `do.call` on a `teal_slice_expr` is problematic if `expr` is a call -> causes do.call to evaluate this.

chlebowa:
I would also lean towards supporting passing (and storing) `filter_var_expr`'s `expr` argument as a string, like you said in the deleted comment.

gogonzo:
> I would also lean towards supporting passing (and storing) `filter_var_expr`'s `expr` argument as a string, like you said in the deleted comment.

Yup, I'll fix this tomorrow. I think string is safer especially with risk of `do.call`

asbates (approved):

PR #265, opened 2023-05-05
206 global disable

### LINKED ISSUE #206: When disabling filter panel, don't remove UI

chlebowa:
Closes #206.

Changes the bahavior of `FilteredData$filter_panel_disable/enable`. Filter states are no longer removed and re-instantiated. Rather, their state is changed to disabled/enabled by sending a filter panel API command down the class structure.

* the global disable/enable method is modified appropriately
* caching individual filter states is removed; caching states at the global level is necessary as some states should be restored disabled
* an observer is added to `FilterState$server` to update the disable switch accordingly


R/FilterState.R
```diff
@@ -251,16 +251,28 @@ FilterState <- R6::R6Class( # nolint
...
-            {
-              if (isTRUE(input$enable)) {
-                private$enable()
-              } else {
-                private$disable()
-              }
-            },
-            ignoreInit = TRUE
+          # Disable/enable this filter state in response to switch flip.
+          observeEvent(input$enable, {
```
gogonzo, line 255: destroy_observers should remove all observers created by the FilterState class. It means that all observers need to be stored in this list.
```suggestion
          private$observers$enable <- observeEvent(input$enable, {
```

R/FilterState.R
```diff
@@ -251,16 +251,28 @@ FilterState <- R6::R6Class( # nolint
...
+            } else {
+              private$disable()
+            }
+          },
+          ignoreInit = TRUE
           )
+
+          # Update disable switch according to disabled state.
+          # This is necessary to react to the global disable action.
+          observeEvent(private$is_disabled(), {
```
gogonzo, line 267:
```suggestion
          private$observers$is_disabled <- observeEvent(private$is_disabled(), {
```

R/teal_slice.R
```diff
@@ -486,3 +489,52 @@ slices_which <- function(tss, expr) {
   expr <- str2lang(expr)
   Filter(function(x) isTRUE(eval(expr, x)), tss)
 }
+
+
+# sets selected fields to NULL in all `teal_slice`s in a `teal_slices`
+#' @rdname teal_slice
+#' @keywords internal
+#'
+slices_drop <- function(tss, drop = NULL) {
```
gogonzo, line 498: drop argument isn't used in the code base
```suggestion
slices_drop <- function(tss) {
```

R/teal_slice.R
```diff
@@ -486,3 +489,52 @@ slices_which <- function(tss, expr) {
...
+      unique(unlist(lapply(tss, names))),
+      c("dataname", "varname", "varlabel", "target")
+    )
+  }
+  attrs <- Filter(Negate(is.null), attributes(tss)[c("include_varnames", "exclude_varnames", "count_type")])
+
+  for (i in seq_along(drop)) {
+    tss <- lapply(tss, function(x) {
+      x <- unclass(x)
+      x[[drop[i]]] <- NULL
```
gogonzo, line 513:
```suggestion
      x[[drop[i]] <- NULL
```

R/FilteredData.R
```diff
@@ -627,8 +627,12 @@ FilteredData <- R6::R6Class( # nolint
       private$filter_panel_active <- FALSE
       fp_id <- self$get_filter_panel_ui_id()
       shinyjs::disable(paste0(fp_id, "-add"), asis = TRUE)
-      private$cached_states <- self$get_filter_state()
-      self$clear_filter_states()
+      slices <- self$get_filter_state()
+      private$cached_states <- slices
+      slices <- slices_drop(slices)
+      slices <- slices_set(slices, list(disabled = TRUE))
```
gogonzo, line 633: this task can be achieved by simple loop without introducing new 52 lines in teal_slice and 162 lines to test them. 
```suggestion
      for (i in seq_along(slices)) {
        slices[[i]]$disabled <- TRUE
      }
```

R/teal_slice.R
chlebowa: It is in unit tests.

R/FilteredData.R
chlebowa: I have a feeling we will have a need to set more fields sooner or later.

gogonzo: So let's introduce these functions later

chlebowa:
api extension removed

gogonzo (commented):
We have two public method `filter_panel_enable` and `filter_panel_disable` which are used only in `srv_active`. I think these methods should be private and maybe merged into one if it's more optimal.

chlebowa:
> We have two public method `filter_panel_enable` and `filter_panel_disable` which are used only in `srv_active`. I think these methods should be private and maybe merged into one if it's more optimal.

I'm ok with moving them to private, but I think they are better left separate, there would be too many `if`s, were they to be combined.

asbates (approved):
Beautiful!

I did notice something odd but I think it mostly likely pertains to individual cards vs. all cards.

1. flip global disable to off so all filters are disabled
2. remove a single filter card via the X icon
3. flip global disable to on

The card removed in step 2 will be added. I'm not sure if this is 'correct' behavior but I don't think it's directly related to this issue either so I opened #269

gogonzo:
lets address #269 later and move on with this one. Please remove tests which directly test disable/enable as disable/enable is a private method and we should rather test private through public only.
Tests of `srv_filter_panel` where disable/enable button is tested should be fixed.

gogonzo (approved):
👍

PR #268, opened 2023-05-05
133 take 2@filter panel refactor@main

### LINKED ISSUE #133: Add option to zoom in sliderInput

asbates:
Adds a numeric range input to range filter state cards so range can be set manually. This is a redo of #226

Fixes #133


R/FilterStateRange.R
```diff
@@ -493,6 +526,74 @@ RangeFilterState <- R6::R6Class( # nolint
...
+                shinyWidgets::updateNumericRangeInput(
+                  session = session,
+                  inputId = "selection_manual",
+                  value = self$get_selected()
+                )
+                return(NULL)
+              }
+              # all.equal not enough here b/c tolerance
+              # all.equal(0.000000001, 0) is TRUE
+              out_of_range <- isFALSE(identical(
```
chlebowa, line 569: `all.equal` was there exactly to allow the tolerance

chlebowa: Also, with `identical` `isFALSE` is redundant but with `all.equal` `isFALSE` will fail and `!isTRUE` is necessary.

chlebowa (changes requested):
Invalid inputs cannot crash the whole app. There has to be some way to block that error. `shinyvalidate` is one option but I don't know whether we want to add it to `teal.slice`.

gogonzo:
> Invalid inputs cannot crash the whole app. There has to be some way to block that error. shinyvalidate is one option but I don't know whether we want to add it to teal.slice.

We can't use `shinyvalidate` as FilterPanel must not return any shiny-validate errors - this would have a serious consequences downstream in `teal` and `teal_module`. So, if user sets the wrong value it should be automatically set to the latest selected with notification

R/FilterStateRange.R
```diff
@@ -493,6 +526,74 @@ RangeFilterState <- R6::R6Class( # nolint
...
+              #  expects it's input to be sorted
+              if (any(is.na(input$selection_manual))) {
+                showNotification(
+                  "Numeric range values must be numbers.",
+                  type = "warning"
+                )
+                shinyWidgets::updateNumericRangeInput(
+                  session = session,
+                  inputId = "selection_manual",
+                  value = self$get_selected()
```
gogonzo, line 551: this is why it fails
```suggestion
                  value = private$get_selected()
```

gogonzo: please replace all `self$get_selected` with `private$get_selected` and then module should be robust enough

chlebowa (changes requested):
Numeric selection states on app start do not follow api specification:

```
  filter_var("iris", "Sepal.Length", sel = c(5.1, 6.4), fixed = FALSE),
  filter_var("iris", "Sepal.Width", sel = c(2.5, 4.0), fixed = FALSE, disabled = TRUE),
  filter_var("iris", "Petal.Length", sel = c(4.5, 5.1), fixed = TRUE),
  filter_var("iris", "Petal.Width", sel = c(0.3, 1.8), fixed = TRUE, disabled = TRUE),
  filter_var("iris", "Species", sel = c("versicolor", "virginica"), fixed = FALSE),
  filter_var("iris", "Species2", sel = c("versicolor", "virginica"), fixed = TRUE)
)
```

image

Also, `choices` become limited.

chlebowa:
I think `RangeFilterState$set_choices` is broken but that's a different issue. I will look into it a little more.

chlebowa:
> Numeric selection states on app start do not follow api specification:
(...)

My mistake, the numbers in the app are the actual data ranges.

asbates:
> Numeric selection states on app start do not follow api specification:
> 
> ```
> default_filters <- filter_settings(
>   filter_var("iris", "Sepal.Length", sel = c(5.1, 6.4), fixed = FALSE),
>   filter_var("iris", "Sepal.Width", sel = c(2.5, 4.0), fixed = FALSE, disabled = TRUE),
>   filter_var("iris", "Petal.Length", sel = c(4.5, 5.1), fixed = TRUE),
>   filter_var("iris", "Petal.Width", sel = c(0.3, 1.8), fixed = TRUE, disabled = TRUE),
>   filter_var("iris", "Species", sel = c("versicolor", "virginica"), fixed = FALSE),
>   filter_var("iris", "Species2", sel = c("versicolor", "virginica"), fixed = TRUE)
> )
> ```
> 
> image
> 
> Also, `choices` become limited.

Huh, it's working for me. This is what I get on launch.

image

And after enabling

image

chlebowa:
> Huh, it's working for me. This is what I get on launch.
> 
> image

uh-oh

donyunardi:
I have the same result with @asbates.
This is what I see upon launch:

chlebowa:
All right, my session was littered and another object got pulled into the app. Works fine for me as well.

PR #272, opened 2023-05-09
capture label and add after droplevels

### LINKED ISSUE #271: FilterStateChoices - label is not present
The variable label is not showing for COUNTRY column:
image

The label for COUNTRY is `Country`.

Upon investigation, this line is dropping the `label` attribute from the column:
https://github.com/insightsengineering/teal.slice/blob/5d5d649e7ae26681d70a41bc709c72509c3a9f1e/R/FilterStateChoices.R#L156

This is what it looks like (before and after) when I filter the ADSL dataset using COUNTRY column:
image

donyunardi:
Fixes #271

### Summary
I captured the `label` attribute in FilterStateChoices before the factor or levels activity.
Then, before passing to parent initialization, added the `label` attribute back to the `x_factor` object.

This seems to fix the problem with the `label` in FilterStateChoices:

### DIFF
--- a/R/FilterStateChoices.R
+++ b/R/FilterStateChoices.R
@@ -148,12 +148,16 @@ ChoicesFilterState <- R6::R6Class( # nolint
         combine = "or"
       )
 
+      label <- attr(x, "label")
+
       x_factor <- if (!is.factor(x)) {
         factor(as.character(x), levels = as.character(sort(unique(x))))
       } else {
         x
       }
+
       x_factor <- droplevels(x_factor)
+      attr(x_factor, "label") <- label
 
       args <- list(
         x = x_factor,


R/FilterStateChoices.R
```diff
@@ -148,12 +148,16 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
+      label <- attr(x, "label")
+
       x_factor <- if (!is.factor(x)) {
         factor(as.character(x), levels = as.character(sort(unique(x))))
       } else {
         x
       }
+
       x_factor <- droplevels(x_factor)
+      attr(x_factor, "label") <- label
```
chlebowa, line 160:
```suggestion
      x_factor <- if (!is.factor(x)) {
        structure(
          factor(as.character(x), levels = as.character(sort(unique(x)))),
          label = attr(x, "label") 
        )
      } else {
        x
      }
```

chlebowa (approved):
:+1:

PR #277, opened 2023-05-10
273 global disable

### LINKED ISSUE #273: disable filter states `disable` switches when filter panel disabled
When the filter panel is disabled with the global filter, individual filters can be re-enabled with their own switches. Make those switches disabled.

chlebowa:
Closes #273.

Modifies `FilteredData$filter_panel_disable/enable` to also affect the `filter_active_vars_contents` div.

### DIFF
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -1096,15 +1096,17 @@ FilteredData <- R6::R6Class( # nolint
     },
 
     # @description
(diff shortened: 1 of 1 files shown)


gogonzo (approved):
Nice and very short solution

follow up: https://github.com/insightsengineering/teal.slice/issues/278

PR #281, opened 2023-05-15
Remove get_formatted_filter_state and change format method to generic print

### LINKED ISSUE #267: Remove get_formatted_filter_state and simplify format methods
Format methods in all the classes should be an abstract method which should look like following:

```r
FilterState
format = function() {
  sprintf(
   "%s:\n%s",
    class(self)[1],
    format(self$get_filter_state())
  )
}
```

`get_filter_state` in `FilteredData` shouldn't return attr("formatted") as returned `teal_slices` has its own format method.

m7pr:
Fixes #267

### DIFF
--- a/R/FilterStateChoices.R
+++ b/R/FilterStateChoices.R
@@ -208,45 +208,6 @@ ChoicesFilterState <- R6::R6Class( # nolint
     },
 
     #' @description
-    #' Returns a formatted string representing this `FilterState`.
-    #'
-    #' @param indent (`numeric(1)`)
-    #'   number of spaces before after each new line character of the formatted string;
-    #'   defaults to 0
-    #' @param wrap_width (`numeric(1)`)
-    #'   number of characters to wrap lines at in the printed output;
-    #'   allowed range is 30 to 120; defaults to 76
-    #'
-    #' @return `character(1)` the formatted string
-    #'
-    format = function(indent = 2L, wrap_width = 76L) {
-      checkmate::assert_number(indent, finite = TRUE, lower = 0L)
-      checkmate::assert_number(wrap_width, finite = TRUE, lower = 30L, upper = 120L)
-
-      # List all selected values separated by commas.
-      values <- paste(format(private$get_selected(), nsmall = 3L, justify = "none"), collapse = ", ")
-      paste(c(
-        strwrap(
-          sprintf("Filtering on: %s", private$varname),
-          width = wrap_width,
-          indent = indent
-        ),
-        # Add wrapping and progressive indent to values enumeration as it is likely to be long.
-        strwrap(
-          sprintf("Selected values: %s", values),
-          width = wrap_width,
-          indent = indent * 2L,
-          exdent = indent * 3L
-        ),
-        strwrap(
-          sprintf("Include missing values: %s", private$get_keep_na()),
-          width = wrap_width,
-          indent = indent * 2L
-        )
-      ), collapse = "\n")
-    },
-
-    #' @description
     #' Returns reproducible condition call for current selection.
     #' For this class returned call looks like
     #' `<varname> %in%  c(<values selected>)` with
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -137,19 +137,27 @@ FilterState <- R6::R6Class( # nolint
     },
 
     #' @description
-    #' Virtual method for printing FilterState.
+    #' Returns a formatted string representing this `FilterState` object.
     #'
-    format = function() {
-      stop("this is a virtual method")
+    #' @param show_all `logical(1)` passed to `format.teal_slice`
+    #'
+    #' @return `character(1)` the formatted string
+    #'
+    format = function(show_all = FALSE) {
+      sprintf(
+        "%s:\n%s",
+        class(self)[1],
+        format(self$get_state(), show_all = show_all)
+      )
     },
 
     #' @description
     #' Prints this `FilterState` object.
     #'
-    #' @param ... additional arguments to this method
+    #' @param ... additional arguments
     #'
     print = function(...) {
-      cat(shiny::isolate(self$format()), "\n")
+      cat(shiny::isolate(self$format(...)), "\n")
     },
 
     #' @description
--- a/R/FilterStateDate.R
+++ b/R/FilterStateDate.R
@@ -181,29 +181,6 @@ DateFilterState <- R6::R6Class( # nolint
     },
 
     #' @description
-    #' Returns a formatted string representing this `DateFilterState`.
-    #'
-    #' @param indent (`numeric(1)`) the number of spaces before after each new line character of the formatted string.
-    #' Default: 0
-    #' @return `character(1)` the formatted string
-    #'
-    format = function(indent = 2) {
-      checkmate::assert_number(indent, finite = TRUE, lower = 0)
-
-      vals <- private$get_selected()
-      sprintf(
-        "%sFiltering on: %s\n%sSelected range: %s - %s\n%sInclude missing values: %s",
-        format("", width = indent),
-        private$varname,
-        format("", width = indent * 2),
-        format(vals[1], nsmall = 3),
-        format(vals[2], nsmall = 3),
-        format("", width = indent * 2),
-        format(private$get_keep_na())
-      )
-    },
-
-    #' @description
     #' Returns reproducible condition call for current selection.
     #' For this class returned call looks like
     #' `<varname> >= <min value> & <varname> <= <max value>` with
--- a/R/FilterStateDatettime.R
+++ b/R/FilterStateDatettime.R
@@ -191,29 +191,6 @@ DatetimeFilterState <- R6::R6Class( # nolint
     },
 
     #' @description
-    #' Returns a formatted string representing this `DatetimeFilterState`.
-    #'
-    #' @param indent (`numeric(1)`) the number of spaces before after each new line character of the formatted string.
-    #' Default: 0
-    #' @return `character(1)` the formatted string
-    #'
-    format = function(indent = 2) {
-      checkmate::assert_number(indent, finite = TRUE, lower = 0)
-
-      vals <- private$get_selected()
-      sprintf(
-        "%sFiltering on: %s\n%sSelected range: %s - %s\n%sInclude missing values: %s",
-        format("", width = indent),
-        private$varname,
-        format("", width = indent * 2),
-        format(vals[1], nsmall = 3),
-        format(vals[2], nsmall = 3),
-        format("", width = indent * 2),
-        format(private$get_keep_na())
-      )
-    },
-
-    #' @description
     #' Returns reproducible condition call for current selection.
     #' For this class returned call looks like
     #' `<varname> >= as.POSIXct(<min>) & <varname> <= <max>)`
--- a/R/FilterStateEmpty.R
+++ b/R/FilterStateEmpty.R
@@ -109,26 +109,8 @@ EmptyFilterState <- R6::R6Class( # nolint
       } else {
         substitute(!is.na(varname), list(varname = private$get_varname_prefixed(dataname)))
       }
-    },
-
-    #' @description
-    #' Returns the filtering state.
-    #'
-    #' @return `list` containing values taken from the reactive fields:
-    #' * `keep_na` (`logical(1)`) whether `NA` should be kept.
-    #'
-    get_state = function() {
-      list(
-        dataname = private$get_dataname(),
-        varname = private$get_varname(),
-        choices = private$choices,
-        selected = private$get_selected(),
-        keep_na = private$get_keep_na(),
-        keep_inf = private$get_keep_inf(),
-        fixed = private$fixed,
-        disabled = private$is_disabled()
-      )
     }
+
   ),
 
   # private members ----
--- a/R/FilterStateExpr.R
+++ b/R/FilterStateExpr.R
@@ -95,30 +95,26 @@ FilterStateExpr <- R6::R6Class( # nolint
     },
 
     #' @description
-    #' Returns a formatted string representing this `FilterState`.
+    #' Returns a formatted string representing this `FilterStateExpr` object.
     #'
-    #' @param indent (`numeric(1)`)
-    #'   number of spaces before after each new line character of the formatted string;
-    #'   defaults to 0
-    #' @param wrap_width (`numeric(1)`)
-    #'   number of characters to wrap lines at in the printed output;
-    #'   allowed range is 30 to 120; defaults to 76
+    #' @param show_all `logical(1)` passed to `format.teal_slice`
     #'
     #' @return `character(1)` the formatted string
     #'
-    format = function(indent = 2L, wrap_width = 76L) {
-      checkmate::assert_number(indent, finite = TRUE, lower = 0L)
-      checkmate::assert_number(wrap_width, finite = TRUE, lower = 30L, upper = 120L)
-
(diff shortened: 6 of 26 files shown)


R/FilterStateChoices.R
```diff
@@ -210,40 +210,14 @@ ChoicesFilterState <- R6::R6Class( # nolint
...
-          width = wrap_width,
-          indent = indent * 2L
-        )
-      ), collapse = "\n")
+    format = function() {
+      sprintf(
+        "%s:\n%s",
+        class(self)[1],
+        format(self$get_filter_state())
+      )
```
gogonzo, line 220: class inherits from `FilterState` - this means that method can be implemented once in parent class. 
The same applies to subclasses of `FilteredDataset` and `FilterStates`.
Apart from this `format` in this PR is exactly what we need

gogonzo (commented):
Looks good. Please fix tests also

gogonzo:
@m7pr 

>1. Should format for FilterState contain fix like below

Please keep `get_state`. At this moment `get_filter_state` methods return `teal_slices` and `FilterState$get_state` returns `teal_slice`. It makes sense to keep naming consistent from top to bottom of the class structure but it might be debatable. Let's not change names at this moment.

> 2. print method - for both FilterState and FilterState**s** classes uses below implementation
>```{R}
>cat(shiny::isolate(self$format()), "\n")
>```
>Should this be changed to `cat(shiny::isolate(format(self), "\n")` as `teal_slice` and `teal_slices` have their own format function >already?

I guess the default `print` method for `R6` just reuse `format`. Please change format according to (1) and print should automatically use it.


>3. Just double checking you are aware that FilteredDataset and it's subclasses do not have any `format` method.

Yes, this gap should be filled.

m7pr:
@gogonzo 

Alrighty, to sum things up

- FilterState received format() method (that uses get_state()) and I removed format from its childs
- FilterStates received format() method (that uses get_filter_state()) and I removed format from its childs
- FilterDataset received format() method (that uses get_filter_state())
- changed tests for FilterState, FilterData and FilterDataset for format() method

Questions
- should we provide tests for FilterStates for format() method as well?
- EmptyFilterState received format now - it is a bit bizzare, are we OK with it?
- FilterState had format() method previously that raised a stop (error) - it is substituted with the format implementation that you proposed, are we OK with that? why it was raising stop before?

Notes
- there are 2 failing tests for ChoicesFilterState that were also there on the filter_panel_refactor branch already (at least for my session). I am leaving those unresolved

````{R}
x | 2      63 | ChoicesFilterState [1.2s]                                                     
----------------------------------------------------------------------------------------------
Failure (test-ChoicesFilterState.R:295:3): set_state sets intersection of choices and passed values
shiny::isolate(filter_state$get_state()$selected) not identical to "2000-01-01 12:00:00".
Lengths differ: 0 is not 1

Failure (test-ChoicesFilterState.R:305:3): set_state sets intersection of choices and passed values
shiny::isolate(filter_state$get_state()$selected) not identical to "2000-01-01 12:00:00".
Lengths differ: 0 is not 1
````

Session Info

```{R}
> sessionInfo()
R version 4.0.1 (2020-06-06)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19045)

Matrix products: default

locale:
[1] LC_COLLATE=English_United Kingdom.1252 
[2] LC_CTYPE=English_United Kingdom.1252   
[3] LC_MONETARY=English_United Kingdom.1252
[4] LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] teal.slice_0.2.0.9029 shiny_1.7.4          

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.8.3            teal.data_0.1.2.9013    teal.logger_0.1.1.9004 
 [4] jquerylib_0.1.4         pillar_1.9.0            compiler_4.0.1         
 [7] bslib_0.4.2             later_1.3.0             shinyjs_2.1.0          
[10] tools_4.0.1             shinyWidgets_0.7.6      digest_0.6.29          
[13] checkmate_2.0.0         jsonlite_1.8.4          lifecycle_1.0.3        
[16] tibble_3.2.1            gtable_0.3.3            pkgconfig_2.0.3        
[19] rlang_1.1.1             cli_3.6.1               rstudioapi_0.14        
[22] yaml_2.3.5              fastmap_1.1.1           dplyr_1.1.2            
[25] teal.widgets_0.2.0.9025 generics_0.1.3          vctrs_0.6.2            
[28] sass_0.4.1              grid_4.0.1              tidyselect_1.2.0       
[31] glue_1.6.2              R6_2.5.1                fansi_1.0.3            
[34] ggplot2_3.4.2           logger_0.2.2            magrittr_2.0.3         
[37] backports_1.4.1         formatters_0.4.1.9001   scales_1.2.1           
[40] promises_1.2.0.1        ellipsis_0.3.2          htmltools_0.5.5        
[43] mime_0.12               xtable_1.8-4            colorspace_2.0-3       
[46] httpuv_1.6.5            utf8_1.2.3              munsell_0.5.0          
[49] cachem_1.0.8  
```

chlebowa:
`EmptyFilterState$get_state` is odd, it should use the parent method. It won't need the `format` then.

Also, I think we should add a `show_all` argument to the `format` methods here so it can be passed to `format.teal_slice`.

```
    #' @description
    #' Virtual method for printing FilterState.
    #'
    format = function(show_all = FALSE) {
      sprintf(
        "%s:\n%s",
        class(self)[1],
        format(self$get_state(), show_all)
      )
    },
```
and then
```
fs <- init_filter_state(iris$Species, dataname = "iris", varname = "Species", selected = "versicolor")
print(fs, TRUE)
```

The documentation needs an update (e.g. remove "virtual" above).

Otherwise, great job :+1:

chlebowa:
Oddly enough, the tests pass on my machine :thinking: 
```
R version 4.2.2 Patched (2022-11-10 r83330)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.6 LTS
```

m7pr:
thanks @chlebowa for the review. I'll 
- bring the `show_all` argument (and extend tests for it) for FilterState
- update the documentation
- remove get_state method for EmptyFilterState
- try tunning the tests and development on the docker container that is used for github workflow tests for this repo

@chlebowa does FilterStates also have any arguments in format.teal_slices that you recommend brining up?

chlebowa:
> @chlebowa does FilterStates also have any arguments in format.teal_slices that you recommend brining up?

Let me explain. The `format` S3 method for `teal_slice` has a `show_all` argument because we didn't want to print all the fields all of the time but wanted to keep the ability to print them when needed. That is why 
```
> filter_var(data = "iris", var = "Species")
teal_slice
 $ dataname: "iris"
 $ varname : "Species"
 $ fixed   : FALSE
 $ disabled: FALSE

> filter_var(data = "iris", var = "Species") %>% print(T)
teal_slice
 $ dataname: "iris"
 $ varname : "Species"
 $ choices : NULL
 $ selected: NULL
 $ keep_na : NULL
 $ keep_inf: NULL
 $ fixed   : FALSE
 $ disabled: FALSE
```
The same goes for the `teal_slices` method, as that in turn calls `format.teal_slice`.

I think it makes sense to keep this mechanic for `FilterState` and higher classes. In that case, all classes should be able to pass `show_all` to the `format` call on the `teal_slice` or `teal_slices` they print.

On the other hand I don't use `show_all = TRUE` all that much, so maybe it's not worth the clutter. What do you think @gogonzo?

m7pr:
Hey @chlebowa 
- I removed `EmptyFilterState$get_state()` - now this is consistent with other `FilterState` children classes
- I added `show_all` parameter to `FilterState$format()` and `FilterStates$format()`
- I edited documentation
- I extended tests for FilterState class and childs
- I did not extend tests for FilterStates class and childs as I heard yesterday this will be removed soon

Would you mind taking a final look?

chlebowa:
Excellent work! Please forgive the mission creep, but can you please modify/add the `format` and `print` methods in `FilteredDataset` and `FilteredData` accordingly?

chlebowa (approved):

m7pr:
@gogonzo wanna take a look, since Aleks is off today?

gogonzo:
I think it's worth to make these tests dependent on checking constistency `Class$format` with `format.teal_slices` and avoid comparison with hardcoded string. Reason for this is if we change `format.teal_slices` (very probable) somebody will have to go through 1000 of lines in tests and fix after.

chlebowa:
> if you go and look for word unluckily in test/testthat folder files there are 4 lines marked with this word in a comment where I think we should create better examples

Oh, that's quite easy, all you have to do is omit `keep_inf` in a `filter_var` call for a non-numeric variable or omit `keep_na` for a variable that contains no NAs. The field then be NULL and there will be a difference between `show_all = FALSE` and `show_all = TRUE`.

> I think it's worth to change tests to checking constistency between Class$format and format.teal_slices (to avoid comparison with hardcoded string). Reason for this is if we change format.teal_slices (very probable) somebody will have to go through 1000 of lines in tests and fix after.

This is a good point, in fact this already happened with #286. I think pasting the class header (literal string) with the output of `format` called on the `teal_slices` that was used to set the state will work nicely.


Lastly, I only now noticed that several tests are inappropriately titled "set_filter_state sets `teal_slice`". I'm afraid a read-through of the titles is necessary. I'm sorry it falls on you, I just did the same thing a few weeks ago.

m7pr:
Thanks for comments! I think this is great input.
For the tests that verifies the format/print output with a long string I was trying to extend what we already had. I understand this is not ideal, so will try to just compare filter_object$format() to format(filter_object). I think it's gonna reduce the complexity of tests in a future potential changes. 

Also the naming of tests (what chlebowa mentioned in this last sentence) is crucial, so I'll make sure they are meaningful and accurate!

m7pr:
hey @chlebowa thanks for the great input, any chance you still have the patience for the last (lol) review :)?

chlebowa (changes requested):
```
testthat::expect_equal(
    shiny::isolate(filter_state$format()),
    shiny::isolate(format(filter_state))
)
```
I'm afraid these have to be modified because (unless I am mistaken) the `format` generic applied to an R6 object calls that object's `$format` method.

I suggest something like
```
filter_state <- filter_settings(filter_var(...))
dataset <- init_filtered_dataset(...)
dataset$set_state(filter_state)
```
and then compare `dataset$format()` vs `format(filter_state)` with header added as literal.

m7pr:
ah, sorry @chlebowa - it's on me - my rookie mistake

m7pr:
Hey @chlebowa so for the `test-FilteredData.R`, this is already happening as you requested. For other cases, where an R6 method was returned, I think we could use `$get_state()/$get_filter_state()` to extract the S3 class out of it and compare it with `format.teal_slice()` like below

```{R}
  posixct <- as.POSIXct("2000-01-01 12:00:00", tz = "GMT") + 0:9
  filter_state <- DatetimeFilterState$new(posixct, dataname = "data", varname = "variable")
  testthat::expect_equal(
    shiny::isolate(filter_state$format()),
    paste0(
      'DatetimeFilterState:\n',
      format(shiny::isolate(filter_state$get_state()))
    )
  )
```

instead of 

```{R}
posixct <- as.POSIXct("2000-01-01 12:00:00", tz = "GMT") + 0:9
filter_state <- DatetimeFilterState$new(posixct, dataname = "data", varname = "variable")
testthat::expect_equal(
    shiny::isolate(filter_state$format()),
    shiny::isolate(format(filter_state))
)
```
where filter_state is 

```{R}
> class(filter_state)
[1] "DatetimeFilterState" "FilterState"         "R6" 
```

chlebowa:
> ah, sorry @chlebowa - it's on me - my rookie mistake

No worries :slightly_smiling_face:

chlebowa:
> For other classes, where an R6 method was returned, I think we could use `$get_state()/$get_filter_state()` to extract the S3 class out of it and compare it with `format.teal_slice()` like below

Yes, that's exactly what I meant :)

m7pr:
Hey @chlebowa few last changes regarding how we test. Thanks to your input we are actually testing S3 `teal_slice` and `teal_slices` classes with their format/print equivalents

Tested on R 4.2.2, R 4.1.3, R 4.0.5

As always, asking for the review. Do you think we should squash all commits into one on this PR?

chlebowa:
>  Do you think we should squash all commits into one on this PR?

We usually squash, yes.

m7pr:
Thanks for the message - let me think about whether we should print superclass name or actual class name

m7pr:
@chlebowa if that makes sense I'll squash commits and we are ready to merge

chlebowa:
Ah, but the superclass will not be encountered in the wild. Typically, our code doesn't use constructors directly, but through wrapper functions, so 
```
init_filter_state(x = iris$Sepal.Length, dataname = "irirs", varname = "Sepal.Length")
```
rather than
```
FilterState$new(x = iris$Sepal.Length, dataname = "irirs", varname = "Sepal.Length")
```

Now, the latter will create a `FilterState` object, which will not print at all (`stop("this is a virtual method")`), so one will never see a `FilterState: ...`
The former is an S3 generic, which will dispatch to an appropriate subclass constructor, here `RangeFilterState`.

I don't really mind the subclass names being printed, I just think it may be confusing to the user (here, app dev, I guess). I'll leave the decision up to you. I uphold my approval.

m7pr:
ok, I hear you now @chlebowa ! I see it's rather child-classes that are visible for the end user

> Now, the latter will create a FilterState object, which will not print at all (stop("this is a virtual method")), so one will never see a FilterState: ...

☝️  the above will show

```{R}
FilterState:
teal_slice
 $ dataname: "irirs"
 $ varname : "Sepal.Length"
 $ fixed   : FALSE
 $ disabled: FALSE
```

currently after this PR, as `FilterState` receives a `$format()` that is inherited by it's child-classes.

So the final call is whether after the below

```{R}
> ini <- init_filter_state(x = iris$Sepal.Length, dataname = "irirs", varname = "Sepal.Length")
> ini
RangeFilterState:
teal_slice
 $ dataname: "irirs"
 $ varname : "Sepal.Length"
 $ choices : c(4.3, 7.9)
 $ selected: c(4.3, 7.9)
 $ fixed   : FALSE
 $ disabled: FALSE
 
> class(ini)
[1] "RangeFilterState" "FilterState"      "R6"
```

we return one of the below
a) RangeFilterState:
b) FilterState:
c) RangeFilterState:\nFilterState

So is the most usable for the end user. If you get `RangeFilterState` (a) then at least it's consistent with the top class returned by `class()` and gives the same information as you would run `class()` - it gives you the information about the main class. It allows to seek through the documentation to check methods. Also the documentation should point that it inherits from `FilterState` so that you can also track it's superclass methods. I think `RangeFilterState` is fine for the end user to start digging in the code and in the documentation. The only thing I would consider is returning (c) - besides the final class, also return the parent class, but if someone is interested in the main class, this is should be included in the documentation and the code. Lastly I think if someone pays attention to `RangeFilterState` character you can connect this character to `FilterState` easily. I would live it as it is for now, but will have your idea in my mind for the next few weeks. Thanks

chlebowa:
Right, I forgot a `$format` and `$print` were added in this very PR :sweat_smile: 
OK, I'm convinced, option (a) is sufficient.

m7pr:
Cool, squashing commits

m7pr:
@chlebowa would you do an honor of merging?

m7pr:
🔥

PR #289, opened 2023-05-17
234 range selection with `plotly`

### LINKED ISSUE #233: Research on improving numeric range input using plotly
continuation from #226

### Summary:
User would like the ability to drag numeric ranges and manually enter the range input.

### Plotly
Plotly is recommended as one of the solution.

#### Example from @nikolas-burkoff 

              > I would have loved @nikolas-burkoff's proposal if the red brackets are draggable

Using plotly this is possible:

The black lines here are draggable (sadly they are draggable vertically as well but I would hope that could be configured?)

image
(code below)

Separately the example [here](https://stackoverflow.com/questions/67289361/add-horizontal-scroll-bar-in-plotly-chart) (after "After @ firmo23's edit:") is really nice

```
library(shiny)
library(plotly)

data <- data.frame(x = runif(1000))

ui <- fluidPage(
  plotlyOutput("plot"),
  verbatimTextOutput("result")
)

  output$plot <- renderPlotly(
    plot_ly(data, x = ~x,type="histogram") %>%
      layout(shapes = list(
        list(type = "line", x0 = 0.4, x1 = 0.4, 
             y0 = 0, y1 = 1, yref = "paper"),
        list(type = "line", x0 = 0.8, x1 = 0.8, 
             y0 = 0, y1 = 1, yref = "paper"))) |>
      # allow to edit plot by dragging lines
      config(edits = list(shapePosition = TRUE))  
  )
  
  selected_range_rv <- reactiveVal(c(0.4, 0.8))
  
  observeEvent(plotly::event_data("plotly_relayout"), {
    x <- plotly::event_data("plotly_relayout")
    if ("shapes[0].x0" %in% names(x)){
      selected_range_rv(c(x[["shapes[0].x0"]], selected_range_rv()[2]))  
    }
    if ("shapes[1].x0" %in% names(x)){
      selected_range_rv(c(selected_range_rv()[1], x[["shapes[1].x1"]]))  
    }
  })

  output$result <- renderText({
    sprintf("The selected range is from %f to %f", min(selected_range_rv()), max(selected_range_rv()))
  })
    
}

shinyApp(ui, server)
```

_Originally posted by @nikolas-burkoff in https://github.com/insightsengineering/teal.slice/issues/226#issuecomment-1471626081_

#### Example from @chlebowa 

> > > I would have loved @nikolas-burkoff's proposal if the red brackets are draggable
> > 
> > 
> > Using plotly this is possible:
> 
> Oh, I assumed `plotly` is off the table, but If not - this is excellent, especially the built-in slider shown in the SO example:
> 
> ```
> library(plotly)
> 
> x <- c(1:100)
> random_y <- rnorm(100, mean = 0)
> data <- data.frame(x, random_y)
> 
> fig <- plot_ly(data, x = ~x, y = ~random_y, type = 'scatter', mode = 'lines') %>%
>   layout(xaxis = list(rangeslider = list()))
> 
> fig
> ```

### Definition of Done:
* This is a research task to explore plotly for numeric range input in the filter panel
* Present a rough implementation prototype using the provided example code

### LINKED ISSUE #234: Research on better solution to manually enter numeric ranges
Continuation from #226

### Summary:
The user desires the option to manually input ranges when filtering data with a numeric variable, as well as the ability to use a slider to select ranges. The proposed solution, #226, involves implementing a toggle button that activates a numericRangeInput, which allows the user to enter a value manually while retaining the slider input. The toggle button would allow the user to switch between the slider and text box to select or enter the range.

However, there is a problem when a value is manually entered, as it must match the slider's steps or ticks. This restricts the user from freely inputting any value.

### Acceptance Criteria:
This is a research task to see if there's a better or robust solution to achieve users requirements:
* Able to use slider to select ranges
* Able to manually enter ranges

chlebowa:
Closes #234 
Closes #233 

Replaces the range slider in `RangeFilterState` with an interactive `plotly` graph. Two shapes (lines) are drawn on the plot that can be dragged and their position is tracked. An observer listens to events emitted by the plot when shapes are altered (this event is called a "plotly_relayout") and updates selection. Another observer listens to the manual input and updates selection. Finally, a third observer listens to the selection and updates the manual input as well as the shapes on the plot.

Since the graph is slower to render, a spinner is added to it to alleviate the negative effect on UX.

Numeric (manual) input is now displayed simultaneously with the graphic input, not alternatively.

Numeric input receives a debounce.

Use this application to test:
```
options("teal.bs_theme" = bslib::bs_theme(version = 4))
devtools::load_all()
library(teal)
library(scda)

  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      })
        attr(data, "code")()
      })
    }
  )
}

### CDISC ###
options(teal.log_level = "TRACE", teal.show_js_log = FALSE)
devtools::load_all()
library(teal)
library(scda)

ADSL$numeric <- ADSL$numeric4 <- rnorm(nrow(ADSL))
ADSL$numeric2 <- sample(1:100000, size = nrow(ADSL))
ADSL$numeric3 <- rep(0, nrow(ADSL))

ADTTE$numeric <- rnorm(nrow(ADTTE))
ADTTE$numeric[sample(1:nrow(ADTTE), size = 10,)] <- NA

  filter_var("ADSL", "numeric", selected = c(0, 1.40), keep_na = TRUE, keep_inf = TRUE, fixed = FALSE),
  filter_var("ADSL", "numeric2", selected = c(10, 10000), keep_na = TRUE, keep_inf = TRUE, fixed = FALSE),
  filter_var("ADSL", "numeric3", selected = c(0, 0), keep_na = TRUE, keep_inf = TRUE, fixed = FALSE),
  filter_var("ADSL", "numeric4", selected = c(0, 1.40), keep_na = TRUE, keep_inf = TRUE, fixed = TRUE),
  filter_var("ADTTE", "AVAL", selected = c(10, 1000), keep_na = TRUE, keep_inf = TRUE, fixed = FALSE),
  exclude_varnames = list(ADTTE = setdiff(c("AVAL", "PARAMCD", "PARAM"), names(ADTTE)))
)

data <- cdisc_data(
  cdisc_dataset("ADSL", ADSL),
  cdisc_dataset("ADRS", ADRS)
)

app <- init(
  data = data,
  filter = default_filters
)

```


chlebowa:
Update : made a breakthrough in hyperreactivity and disabling. It's not quite ready but getting somewhere.

asbates:
This is so cool!!!

I will make my obligatory comment about not hard-coding colors ;) but that's out of scope for this.

The only real things I see are

1. It looks like some overlap of the code in `server_inputs` and `server_inputs_fixed` so that could be moved out into a method/function.
2. Selections are still limited to a specific set of values.

These are probably pretty obvious though and because this is a WIP. I can see this feature is going to be useful and cool.

chlebowa:
> I will make my obligatory comment about not hard-coding colors ;) but that's out of scope for this.

Duly noted, the colors used here are placeholders.

> 1. It looks like some overlap of the code in `server_inputs` and `server_inputs_fixed` so that could be moved out into a method/function.

I will try to optimize this but it may be necessary, actually, because the plot in `server_inputs` emits events that must be namespaced. Perhaps the `plot_*` building blocks can be stored in the object and only be put together in the server.

> Update: I think it would be a good idea to make this a stand-alone widget, maybe in`teal.widgets`. I think people would find this useful outside of teal.

Hmm, that's certainly something to think about.

Thanks, @asbates!

donyunardi:
Truly great stuff!

Played around with this a bit more and found an interesting behavior.
If I use keyboard down arrow in the range input and I press up or down quickly several times, the reactive loops kicks in and the app is stuck in computing state.

here's a video where I press down in the max range input box several times rapidly:
https://github.com/insightsengineering/teal.slice/assets/8597300/c2c7ca97-14ae-421f-9b17-a151dd3fb0de

lcd2yyz:
Great work! I think the main functionality and most of the expected behavior is already there, very exciting to play around. 

I only did UI review, and noted down some comments along the way...

1. Need to set a minimal version dependency on `bslib` and `shinyWidgets` 
When I used `bslib_0.3.1` (that came with NEST 2022_10_13 release on BEE), app errors out as soon as default filters are loaded. Once I installed `bslib_0.4.2` (latest CRAN release) it worked fine. Similarly, `shinyWidgets_0.6.0` errors out. Upgrade to `shinyWidgets_0.7.6` solved the problem. The true minimal version dependency is probably somewhere in between the two versions?

2. It takes a good 5 second for plotly display to show up upon expanding the first card for the first time. It's much faster (~1 sec) for all subsequent cards. Can we somehow shorten the delay time for the first action? 

3. When user manually specify a value outside the range of the input data, the draggable boundary line goes off the display. At this point, user has to double click twice to "zoom out" to see both boundary lines. 
Is it possible to swap these actions? Basically visualization range automatically adjust to the min/max of data and manual input value, and user can double click to "zoom in" to the original range of the input data. 

4. The draggable lines can be moved up and down, which can be dragged to above or below the plotting region. Can we somehow limit to two-dimension drag only? Or maybe try draw the line segment as `y0 = -Inf, y1 = Inf`, so it would never go outside the plotting area?

5. There are A LOT of hidden interactive actions on the plotly visualization. While it's quite powerful, some might be not very intuitive, and also are very sensitive to where you do the clicking. The one-time pop-up message (which doesn't show up beside the plot) is definitely not enough. Can we add a tooltip to the top-right corner of the plotly to provide some detailed instructions?

6. Various aesthetics
   6.1 I think users would prefer to see actual axis tick marks (both major and minor) and a smaller steps of the axis major tick labels (eg. something more similar to what we had during the user demo) 
   6.2 Draggable line extending into the axis label section feels very extruding. Can we shorten the line to only appear within plotting region?
image

   6.3 Can we increase the height of the histogram some more? The histogram chart feels very compressed at the moment.
   6.4 Can we use the same two-tone colors for the categorical count horizontal bar char (dark vs light colored)? Currently the histogram is showing color vs grey
   6.5 Increase top/bottom margin around plot - add some spacing between plot and numerical input box.

chlebowa:
> 1. Need to set a minimal version dependency on `bslib` and `shinyWidgets`
>    When I used `bslib_0.3.1` (that came with NEST 2022_10_13 release on BEE), app errors out as soon as default filters are loaded. Once I installed `bslib_0.4.2` (latest CRAN release) it worked fine. Similarly, `shinyWidgets_0.6.0` errors out. Upgrade to `shinyWidgets_0.7.6` solved the problem. The true minimal version dependency is probably somewhere in between the two versions?

I updated dependencies. For the record, the `shinyWidgets` one was not caused by this PR, but by the one that introduced updating `numericRangeInput`.

> 2. It takes a good 5 second for plotly display to show up upon expanding the first card for the first time. It's much faster (~1 sec) for all subsequent cards. Can we somehow shorten the delay time for the first action?

I admit, `plotly` graphs take a while to render, but 5 seconds seems a bit much. All I can say is it doesn't take anywhere near that long on my machine. Out of curiosity, what is the timing with the current `ggplot`?

> 3. When user manually specify a value outside the range of the input data, the draggable boundary line goes off the display. At this point, user has to double click twice to "zoom out" to see both boundary lines.
>    Is it possible to swap these actions? Basically visualization range automatically adjust to the min/max of data and manual input value, and user can double click to "zoom in" to the original range of the input data.

Out of bound values should be removed and selection should be adjusted to `private$choices`. This requirement was lifted at some stage but I restored it now. Actually, it will probably be removed again to satisfy another user requirement, whereupon axis range determination will have to be modified slightly.

> 4. The draggable lines can be moved up and down, which can be dragged to above or below the plotting region. Can we somehow limit to two-dimension drag only? Or maybe try draw the line segment as `y0 = -Inf, y1 = Inf`, so it would never go outside the plotting area?

One-dimensional dragging is impossible, to my knowledge. Setting coordinates to infinite will not solve the issue, a limited segment is drawn. But I will notice that with the current settings (-5 to +5) it takes quite a lot of effort to drag the line so far that one sees the end. I will increase the limits to (-10 to +10) to make it harder still.

> 5. There are A LOT of hidden interactive actions on the plotly visualization. While it's quite powerful, some might be not very intuitive, and also are very sensitive to where you do the clicking. The one-time pop-up message (which doesn't show up beside the plot) is definitely not enough. Can we add a tooltip to the top-right corner of the plotly to provide some detailed instructions?

Can you be more specific? I see only four actions: drag lines, drag axis to pan, drag across plot to zoom, and double-click to unzoom.
The plot is small, a tooltip will cover it completely and make it unusable.

> 6. Various aesthetics
>    6.1 I think users would prefer to see actual axis tick marks (both major and minor) and a smaller steps of the axis major tick labels (eg. something more similar to what we had during the user demo)

Zooming makes it very difficult to determine tick positions. Doing that for data of any order of magnitude is even harder. The default settings are suboptimal but as one zooms in closer, more ticks do appear.

>    6.2 Draggable line extending into the axis label section feels very extruding. Can we shorten the line to only appear within plotting region?
>    image

This cannot be done. The reason the lines cross the axis is the user can drag them upwards and see the bottom end.

>    6.3 Can we increase the height of the histogram some more? The histogram chart feels very compressed at the moment.

The histogram is now larger than the `ggplot` version. How large do you want it?

>    6.4 Can we use the same two-tone colors for the categorical count horizontal bar char (dark vs light colored)? Currently the histogram is showing color vs grey

This is out of scope.

>    6.5 Increase top/bottom margin around plot - add some spacing between plot and numerical input box.

How much?

chlebowa:
@donyunardi solved

chlebowa:
How about this?
https://github.com/insightsengineering/teal.slice/assets/114988527/f3ebd6ba-f41f-444c-a040-5bbec869f77f

It adds one line of code and one dependency.

donyunardi:
It felt pretty solid to me but want to give opportunity for Chendi to respond.
In my machine, the loading time for Plotly is not as long as 5 seconds. It does require some time to load, usually less than 1 second or at most 1.5 seconds, but it's not at an annoying level. However, waiting for 5 seconds does feel a bit too long.
Were you in BEE @lcd2yyz ?

chlebowa:
To be fair, deployment environments are likely to be slower than our laptops, so the need over there might be greater. On the other hand, spinners do add to the loading time a little.

chlebowa:
Any comments on my questions in the opening post?

lcd2yyz:
> > 2. It takes a good 5 second for plotly display to show up upon expanding the first card for the first time. It's much faster (~1 sec) for all subsequent cards. Can we somehow shorten the delay time for the first action?
> 
> I admit, `plotly` graphs take a while to render, but 5 seconds seems a bit much. All I can say is it doesn't take anywhere near that long on my machine. Out of curiosity, what is the timing with the current `ggplot`?

No noticeable delay with `ggplot` histogram. For the record, yes I was testing on BEE (R 4.1.1)
I see the in the video there is a spinner shown during loading with the latest update, but I still see a blank area (no spinner) for ~5 sec. I'm wondering if I need some package upgrade again. 

> > 3. When user manually specify a value outside the range of the input data, the draggable boundary line goes off the display. At this point, user has to double click twice to "zoom out" to see both boundary lines.
> >    Is it possible to swap these actions? Basically visualization range automatically adjust to the min/max of data and manual input value, and user can double click to "zoom in" to the original range of the input data.
> 
> Out of bound values should be removed and selection should be adjusted to `private$choices`. This requirement was lifted at some stage but I restored it now. Actually, it will probably be removed again to satisfy another user requirement, whereupon axis range determination will have to be modified slightly.

I actually prefer the previous version - we need to allow out of bound values. Our expert users all have spoken in favor of using whatever value is specified through the manual input, this to me would include oob values. I was simply asking if the plotly zoom-out action can be reversed - if not possible after some exploration, then we can drop this.
 
> > 4. The draggable lines can be moved up and down, which can be dragged to above or below the plotting region. Can we somehow limit to two-dimension drag only? Or maybe try draw the line segment as `y0 = -Inf, y1 = Inf`, so it would never go outside the plotting area?
> 
> One-dimensional dragging is impossible, to my knowledge. Setting coordinates to infinite will not solve the issue, a limited segment is drawn. But I will notice that with the current settings (-5 to +5) it takes quite a lot of effort to drag the line so far that one sees the end. I will increase the limits to (-10 to +10) to make it harder still.

Increasing the y-length definitely helps, it will take many more dragging attempts to see the end of the line segment (normal users probably won't reach that many). However, I noticed another bug. When the line segment is dragged out of the plotting region, user can still attempt to bring it back into view again, but this action crashes the app.

https://github.com/insightsengineering/teal.slice/assets/78242757/bfb73285-e6ca-43d7-8ade-f2d70b63b6ff

Error I get in console
```
[WARN] 2023-06-01 05:48:17.2560 pid:4015 token:[acf3cbea] teal.slice attempt to set state on fixed filter aborted: ADSL numeric4
Warning: Error in <-: replacement has length zero
  1: runApp
```


> > 5. There are A LOT of hidden interactive actions on the plotly visualization. While it's quite powerful, some might be not very intuitive, and also are very sensitive to where you do the clicking. The one-time pop-up message (which doesn't show up beside the plot) is definitely not enough. Can we add a tooltip to the top-right corner of the plotly to provide some detailed instructions?
> 
> Can you be more specific? I see only four actions: drag lines, drag axis to pan, drag across plot to zoom, and double-click to unzoom. The plot is small, a tooltip will cover it completely and make it unusable.

Exactly as you described, 3 of these 4 actions involve mouse drag action, but to different regions of a rather small plot. I think new users will be quite confused about what's happening, if they are not aware of the region-specific behavior.  
By "tooltip" (sorry about the vague naming) I actually meant something like a small action button with the question mark symbol, that popup nearby to display the instructions. Clicking outside of the popup will dismiss/close it, allowing users to return to interacting with the plot. 

> > 6. Various aesthetics
> >    6.1 I think users would prefer to see actual axis tick marks (both major and minor) and a smaller steps of the axis major tick labels (eg. something more similar to what we had during the user demo)
> 
> Zooming makes it very difficult to determine tick positions. Doing that for data of any order of magnitude is even harder. The default settings are suboptimal but as one zooms in closer, more ticks do appear.

I do not see any ticks appearing even when zoomed in. I would imagine plotly does offer some tickmark/grid mechanism out-of-the-box. Not asking for anything fancy, trying out some default setting may already improve it. 

> > 6.2 Draggable line extending into the axis label section feels very extruding. Can we shorten the line to only appear within plotting region?
> > image
> 
> This cannot be done. The reason the lines cross the axis is the user can drag them upwards and see the bottom end.

This is achievable in `ggplot` using clipping. Can you try to look into if `plotly` has a similar function? Basically line segment below the y = 0 axis will be masked.

Original points 6.3-6.5 can be done together when we do the comprehensive UX review, it will be better to look at everything together uniformly. So I will log a separate issue and add to the tracker list.

chlebowa:
I tested on Ocean. Plots do take a few seconds to load, although not as many as 5. I added the spinner, have a look. Notably, static plots (e.g. in fixed filters) load much faster. I guess the loading time is the trade of for the reactivity.

lcd2yyz:
Re 2 - Spinner looks great during initial loading now. 
But I feel it's adding a bit of rendering time now every time user drags the slider line. It's acceptable delay, but a bit annoying after a few interactions. Is it possible to show spinner only upon initial loading (eg. when card is being expanded)? 

Re 3 - Allowing oob value by manual inputs - Okay agree, I will create a separate issue.

Re 4 - Length of draggable line feels good, normal usage shouldn't break or come to the end of the line segment.

Re 5 - Good job on the action button. Maybe with 3 small tweaks: 
- Can we use the standard `fa-question-circle` icon with a reduced size (by ~30%-ish)? 
- Can we have the popup show up just below the question mark, in the same region as the filter panel
- Button seem to have added a margin to the top of the plot, is it possible to have it overlay on top of the plot? I can also add this to the UX review task if not easy to implement.

Re 6.1 - I literally meant the axis ticks itself (which is shown not in your video either). Tick labels looks good to me. 
image

Re 6.2 - Clipping of line segment - okay I can let this go if it proves to be difficult. 

Thanks for all the quick work and perseverance in addressing my comments!

chlebowa:
> Re 2 - Spinner looks great during initial loading now.
But I feel it's adding a bit of rendering time now every time user drags the slider line. It's acceptable delay, but a bit annoying after a few interactions. Is it possible to show spinner only upon initial loading (eg. when card is being expanded)?

It is not. However, while the spinner does add rendering time, even without it there is a non-zero time when the plot is re-rendered. You can see that there is a delay between dropping the line and Active Filter Summary updating.
I changed the spinner so that it appears on top of the plot rather than replacing it.

> Re 5 - Good job on the action button. Maybe with 3 small tweaks:
> * Can we use the standard `fa-question-circle` icon with a reduced size (by ~30%-ish)?

done

> * Can we have the popup show up just below the question mark, in the same region as the filter panel

I though about it but I can't do it.

> * Button seem to have added a margin to the top of the plot, is it possible to have it overlay on top of the plot? I can also add this to the UX review task if not easy to implement.

I had tried and failed. I tried again, done.

> Re 6.1 - I literally meant the axis ticks itself (which is shown not in your video either). Tick labels looks good to me. image

Oooh, sorry for the misunderstanding.
Minor ticks refuse to cooperate at the moment. I modified the major ones so that there are always 10, all with labels. Is this all right?

lcd2yyz:
Help button and tick mark looks great now :+1:

But I noticed another bug. The up/down button for the lower bound input is reversed.
https://github.com/insightsengineering/teal.slice/assets/78242757/092bbd23-795a-4f23-993a-b01d7ad6650a

The rendering delay from `plotly` is less than ideal, it may just be a trade-off we will have to take for the nicer UI functionalities.

chlebowa:
It looks right to me, the lower bound is a negative number.

chlebowa:
I cleaned up the code and restored unit tests. As far as I am concerned this is the final form of this feature, subject to detailed styling in a separate issue.

lcd2yyz:
> It looks right to me, the lower bound is a negative number.

You are absolutely right, very embarrassing mistake. I should not have been doing this on a late Friday night :sleeping:

m7pr:
This is really cool! Felt like a manual tester for a bit, which was very refreshing!
Few minor comments that I have, but feel free to omit!

## 1
Are we able to format numbers in `regular` way and not `scientific` for big numbers?
I used the app from this initial comment of this PR https://github.com/insightsengineering/teal.slice/pull/289#issue-1714000053

image

## 2
Are we able to extend `Plot actions` tooltip window with an information
- double click to reset pan axis to the original state 

so that double click doesn't meen to `zoom out` but to like `refresh to the original window`
image

## 3
Can we extend `Plot actions` tooltip with window with an information that blue bar is all the data, and gray one is the filtered data? It is a bit unclear from the beginning

## 4
Have you seen you can keep double-clicking to change axis labels on the plot? Check out below video. Is this a required behavior or this should not happen?

https://github.com/insightsengineering/teal.slice/assets/133694481/550ed978-a536-4126-8e30-08a105cedca8

kartikeyakirar:
The new range selection has an enhanced visual appeal looks amazing, thanks to the prominent navigation bar. It significantly simplifies the process of moving between points. However, there is a slight issue when users repeatedly click the up or down arrow. The system records each input and processes it in sequence. To address this, we can incorporate the concept of debounce or throttle,  [debounce/throttle](https://shiny.posit.co/r/reference/shiny/1.0.4/debounce.html) . By doing so, we can effectively discard any unnecessary intermediate inputs.

lcd2yyz:
> ## 1
> Are we able to format numbers in `regular` way and not `scientific` for big numbers? I used the app from this initial comment of this PR [#289 (comment)](https://github.com/insightsengineering/teal.slice/pull/289#issue-1714000053)

Formatting with scientific notation was added on my request. Initially I thought it can be a good way of handle the granular "steps" that might be coming through our old slider (eg. showing up as 100.0315 for example). But now that we've consolidated the slider with the manual input UI and after some testing, I'm also having second thoughts. 
I think whatever the user specify through the manual input, should show as-is for the summary line as well, because the backend filter algorithm is already applying the condition as-is, so they should stay consistent. Please help open a separate issue if needed.
 
> ## 4
> Have you seen you can keep double-clicking to change axis labels on the plot? Check out below video. Is this a required behavior or this should not happen?

Looks to be an artifact of the built-in functionality of "double click to zoom-out/reset" from plotly? 

> The new range selection has an enhanced visual appeal looks amazing, thanks to the prominent navigation bar. It significantly simplifies the process of moving between points. However, there is a slight issue when users repeatedly click the up or down arrow. The system records each input and processes it in sequence. To address this, we can incorporate the concept of debounce or throttle, [debounce/throttle](https://shiny.posit.co/r/reference/shiny/1.0.4/debounce.html) . By doing so, we can effectively discard any unnecessary intermediate inputs.

Agree, the idea of debounce processing was discussed previously (but maybe not logged as issue yet?). Although I would like to see this handled at a higher level (eg. for all filter classes), not just the numerical input. 

> we can utilize the tooltip feature with hover or onclick events using the shinyBS bsPopover/bsTooltip function.

Thanks for the suggestion. No preference from my side about the backend mechanism, as long as the help feature is accessible for users. I will defer to the coredev team to decide on the best approach from technical perspective.

chlebowa:
Thanks for the input. Let me respond in one comment to avoid splitting the discussion.

**Help content**. 
First, bar colors are not plot actions. Furthermore, showing all/filtered data in different colors is also done for categorical and logical values. I think we have advertised it enough that app devs know what they mean and they can instruct their apps' users. It should also be mentioned in the documentation somewhere (I do not know whether it is). That being said, we may decide to extend the help to include plot _properties_ as well. What do you think @lcd2yyz? 
Second, the double click is not exactly "reset to original", it is "extend to encompass all features (observations and shapes)". In the future, once we allow out of bounds selection, one will be able to put the lines further out than the data and then a double-click will set the range so that everything is visible, which will not be the same as the original. I believe "zoom out" better describes this behavior. I also tried to keep it as brief as possible.

**Summary format** The current setting is "print as scientific format if more than 4 significant digits". I agree, it can look a tad weird but it is a simple rule that serves comparatively well for all scales of values. Keep in mind that dragging the lines can produce any arbitrary number (subject to screen resolution) but it is rounded to 4 significant digits before setting selection. I think a better way could be found but that would require a separate discussion and would be out of scope of this issue. If no-one objects, I will change the summary to print the numbers as is and add this point to the [UX issue](https://github.com/insightsengineering/teal.slice/issues/320).

**Debounce**
Some modules in tmc/tmg/tmgoshawk have debounce added to plots, which take the longest to refresh. I am not against adding  debouncing in the filter panel. Since this is being thought about on a higher level, I would consider it out of scope. We would also have to check whether it stacks.

**Dialog vs tooltip** I had tried `bsPopover` but could not get it to work. Feel free to try.

kartikeyakirar:
#289 
> Debounce
> Some modules in tmc/tmg/tmgoshawk have debounce added to plots, which take the longest to refresh. I am not against adding debouncing in the filter panel. Since this is being thought about on a higher level, I would consider it out of scope. We would also have to check whether it stacks.

@chlebowa I just read your comment about debounce, and I completely agree that this issue should be addressed at a higher level, especially in cases involving numeric ranges and input fields with up and down buttons. When users rapidly change the inputs, it can result in increased latency. That's why I thought about incorporating debounce. Please let me know if we should retain it for now until it is implemented at higher levels, or if we should refactor it from this point.

lcd2yyz:
> **Help content**
> That being said, we may decide to extend the help to include plot _properties_ as well. What do you think @lcd2yyz?

I would not add plot properties to help, at least not right now, for two reasons: 
1. As @chlebowa already mentioned, we also have the two-tone coloring for categorical checkboxes. If we add here, we should also add a help button there, which I don't feel is necessary, given how well our user check-in demo last month. 
2. Ultimately we would want the filter panel coloring to be set by the bootstrap theme controlled by the app developer. So the coloring will not always be grey-blue, Not sure how we can robustly describe what color we are referring to in help?
So I would stick with what we have now, unless we get user feedback or user request in the future. 

> **Summary format**
> Keep in mind that dragging the lines can produce any arbitrary number (subject to screen resolution) but it is rounded to 4 significant digits before setting selection.
> If no-one objects, I will change the summary to print the numbers as is and add this point to the https://github.com/insightsengineering/teal.slice/issues/320.

Perfect, exactly what I have in mind :thumbsup:

kartikeyakirar:
Test is failing 
image

gogonzo:
If the issue below will be solved in this PR please close as well

https://github.com/insightsengineering/teal.slice/issues/114

R/zzz.R
```diff
@@ -18,3 +18,21 @@
...
+.rectify_dependencies_check <- function() {
+  dplyr::filter
+  grDevices::rgb
+  lifecycle::badge
+  logger::log_trace
+  plotly::plot_ly
+  shinycssloaders::withSpinner
+  shinyWidgets::pickerOptions
+  teal.widgets::optionalSelectInput
+}
```
gogonzo, line 35: I don't understand this. Can you explain?
Do I understand properly that checks doesn't recognise usage of the functions in methods of R6 classes?

R/FilterStateRange.R
```diff
@@ -360,23 +400,16 @@ RangeFilterState <- R6::R6Class( # nolint
...
+      if (length(values) != 2) stop("Vector of set values must have length two.")
+      if (values[1L] > values[2L]) stop("Vector of set values must be sorted.")
+
+      values
     },
-    # for numeric ranges selecting out of bound values is allowed
+    # Trim selection to limits imposed by private$choices
     remove_out_of_bound_values = function(values) {
+      if (values[1L] < private$choices[1L]) values[1L] <- private$choices[1L]
+      if (values[2L] > private$choices[2L]) values[2L] <- private$choices[2L]
```
gogonzo, line 412: way better to remove values in `remove_out_of_bound_values` instead of `cast_and_validate`!
Thanks

R/zzz.R
chlebowa: That is exactly what happens.

If `package` is listed in `Imports`, (1) check looks through `NAMESPACE` to see if there are any `importFrom(package,<foo>)` statements or an `import(package)` statement. If none are found, (2) check looks for `package::*` calls in the code. If none are found again, (3) check throws the NOTE, which is banned by our CI.
Now, when `package::*` statements are made within an `R6` class, they are not registered. Since `plotly` and `shinycssloaders` were only used within `RangeFilterState$server_inputs`, I kept getting 
```
  Namespaces in Imports field not imported from:
    'plotly' 'shinycssloaders'
    All declared Imports should be used.
```

R/FilterStateRange.R
chlebowa: The `values_adjusted` part removed from `cast_and_validate` did have the effect of removing out of bound values, but it was mostly to match arbitrary numbers to slider ticks. Before `contain_interval` was introduced a few months ago, `remove_out_of_bound_values` was actually doing its job.

gogonzo (approved):
It's amazing!

chlebowa:
Thanks @gogonzo. I will add the detailed explanation of the dependency check issue to comments in `zzz.R`.

R/teal.slice.R
```diff
@@ -8,15 +8,5 @@
 # Fix R CMD check notes
 #' @import R6
 #' @import shiny
-#' @importFrom dplyr filter
-#' @importFrom ggplot2 ggplot
-#' @importFrom grDevices rgb
-#' @importFrom lifecycle badge
```
m7pr, line 14: respect for deleting all those @impoftFrom functions lol

R/FilterStateRange.R
```diff
@@ -295,13 +338,10 @@ RangeFilterState <- R6::R6Class( # nolint
       # Required for displaying ticks on the slider, can modify choices!
       if (identical(diff(x_range), 0)) {
         choices <- x_range
-        private$slider_ticks <- signif(x_range, digits = 10)
-        private$slider_step <- NULL
       } else {
-        x_pretty <- pretty(x_range, 100L)
+        x_pretty <- pretty(x_range, 10000L)
```
lcd2yyz, line 342: Since we are no longer formatting/rounding, user can input as many sign digits as they like in the manual input, should not need tiny step size for the numeric input anymore. Let's revert to the original `100L` as the step size.

chlebowa: Are you sure?

10000 steps:

https://github.com/insightsengineering/teal.slice/assets/114988527/6c4d13ed-d573-42ec-96e1-4de67b726905

100 steps:

https://github.com/insightsengineering/teal.slice/assets/114988527/6f9a77b1-5842-400d-aaf8-a681f7875e2e

kartikeyakirar (approved):
Although my contribution was minimal, I must acknowledge that the overall work done by Aleks is truly amazing.

PR #304, opened 2023-05-29
242 update filter_panel vignette with shinyjs::useShinyjs

### LINKED ISSUE #242: Update example in vignette
The team looked into this together and it was working just fine using the `filter_panel_refactor@main` once we add the `useShinyjs()` after the `fluidPage`

```r
app <- shinyApp(
  ui = fluidPage( #here
    shinyjs::useShinyjs(),
    fluidRow(
      column(
        width = 9,
        id = "teal_primary_col",
...
```
We need to update the example to include have `useShinyjs()`.
Case closed for this ticket and open a new one for updating the vignette.

_Originally posted by @donyunardi in https://github.com/insightsengineering/teal.slice/issues/157#issuecomment-1499489662_

m7pr:
close #242

### DIFF
--- a/vignettes/filter-panel.Rmd
+++ b/vignettes/filter-panel.Rmd
@@ -347,6 +347,7 @@ datasets <- init_filtered_data(
 
 app <- shinyApp(
   ui = fluidPage(
+    shinyjs::useShinyjs(),
     fluidRow(
       column(
         width = 9,
--- a/vignettes/filter-panel.Rmd
+++ b/vignettes/filter-panel.Rmd
@@ -385,15 +385,12 @@ app <- shinyApp(
       )
     })
     observeEvent(input$remove_species_filter, {
-      states <- get_filter_state(datasets)
-      if (!is.null(states$iris$Species)) {
-        remove_filter_state(
-          datasets,
-          filter_settings(
-            filter_var(dataname = "iris", varname = "Species")
-          )
+      remove_filter_state(
+        datasets,
+        filter_settings(
+          filter_var(dataname = "iris", varname = "Species")
         )
-      }
+      )
     })
     observeEvent(input$remove_all_filters, clear_filter_states(datasets))
   }


chlebowa:
Can you please also update the app in the other vignette?

chlebowa:
By the way, can we enhance the code blocks so that they have "copy" buttons? This might be a separate issue, granted, but a very simple one.

m7pr:
@chlebowa I was thinking the same, however pkgdown websites does that for us automatically right? But it needs to be parsed by pkgdown and uploaded as a site?

m7pr:
@chlebowa yeah, check this out this pkgdown article https://insightsengineering.github.io/teal.slice/latest-tag/articles/filter-panel.html where you have copy buttons on the right corner of each chunk, if you hover on the chunk.

image

If you'd like to add a custom button, then for pkgdown websites we would have 2 buttons : P and that's weird.
I suspect our grand entrance area is pkgdown website and we are expecting people to view it more often than regular vignettes, as on the pkgdown website we do also have package README and function's reference.

Lastly I think it could be executed with a one-liner but only if you add another dependency to the package, which is not worth it in my opinion. If you'd like to talk more, we can start a Discussion in Discussion section?

donyunardi (changes requested):
Upon executing the example script, it appears that the functionality of the "Remove iris@Species filter" button is no longer functioning. The root cause of this issue lies within the if statement found in the following line:

https://github.com/insightsengineering/teal.slice/blob/2a8323ab2edb2bce3b3fe79f8a20ee2af1c9814e/vignettes/filter-panel.Rmd#L388-L389

As a result of the `filter_panel_refactor`, `get_filter_state` now returns a `teal_slice` object. Due to the change in object structure, we need to modify this specific line during the update.

I tried this and it works:
```r
if (!is.null(states) && states[[1]]$dataname == "iris" && states[[1]]$varname == "Species") {  
```

donyunardi:
> If you'd like to add a custom button, then for pkgdown websites we would have 2 buttons : P and that's weird. I suspect our grand entrance area is pkgdown website and we are expecting people to view it more often than regular vignettes, as on the pkgdown website we do also have package README and function's reference.

I concur with this viewpoint. Having two separate buttons for copying would be odd, and it would be challenging to devise a mechanism to enable or disable them individually.

m7pr:
@donyunardi on which branches were you testing it? I tried to run it on main as this actually should be a quick update to main, even though we provided a PR to filter_panel_refactor so that we do not merge filter_panel_refactor to the main. Should we create a separate issue for fixing vignettes at all, before merging filter_panel_refactor into main?

chlebowa:
I agree, there's no use grappling with the copy buttons.

This can't go into main because it uses `teal_slice`s. I'll have a look at Dony's bug.

m7pr:
thanks @chlebowa

chlebowa:
It seems to me the `if` can go altogether. The new api does not crash when one attempts to remove a non-existent filter state.

m7pr:
hey thanks for the input. I'll try incorporating that into the vignette

m7pr:
test message from Git Kraken

donyunardi:
> @donyunardi on which branches were you testing it? I tried to run it on main as this actually should be a quick update to main, even though we provided a PR to filter_panel_refactor so that we do not merge filter_panel_refactor to the main. Should we create a separate issue for fixing vignettes at all, before merging filter_panel_refactor into main?

The base branch of this PR is pointing to filter_panel_refactor@main so I thought I should test it using that branch.

m7pr:
Great, thanks. All clear

m7pr:
Hey @chlebowa it works with your proposed changes - thanks. I just wanted to double check the expected behavior. If no filter is used then we present NULL and NULL, if one filter is selected then we present a NULL and a list with a filter code (?) and if two filters are present then we show the filter code. Check below video and the text area that shows the code used for the analysis
https://github.com/insightsengineering/teal.slice/assets/133694481/6abb9dde-32d2-4343-a255-78368bb811f8

m7pr:
https://github.com/insightsengineering/teal.slice/assets/133694481/745fd618-7e7e-440e-a88e-e1d53fb46065

chlebowa:
That panel contains the output of `FilteredData$get_call`, which returns a list of calls, one per data set. Those calls in turn are `NULL` if no observations are filtered out. This seems to correspond to what you see.

m7pr:
@donyunardi let me know what you think on your end

donyunardi (approved):
LGTM!

PR #318, opened 2023-06-02
extend check and docs github workflows with a trigger on filter_panel_refactor branch

m7pr:
One of actions we aligned on during this discussion https://github.com/insightsengineering/teal.slice/discussions/313 was to setup automated testing for branches that for a longer time substituted main branch. In this case it's `filter_panel_refactor` branch.


donyunardi (approved):
lgtm

PR #328, opened 2023-06-07
fixing build fail and adding additional test to fix.

kartikeyakirar:
# fixing build fails

(https://github.com/insightsengineering/teal.slice/actions/runs/5197026426)

Here I fixed the build fail issue by adding internal calls to examples and fixing filter_var example.


m7pr (approved):
thanks, all good

PR #333, opened 2023-06-12
234 range selector few modification

kartikeyakirar:
@chlebowa  

I have implemented a debounce of 500ms for the input$selection_manual. Additionally, I have replaced the modal with a popover (which can alternatively be replaced with a tooltip).

LEt me know you thoughts and suggestions

PR #334, opened 2023-06-12
234 range selector modified by adding debounce and popover

kartikeyakirar:
@chlebowa

I have implemented a debounce of 500ms for the input$selection_manual. Additionally, I have replaced the modal with a popover (which can alternatively be replaced with a tooltip).

Initially, I attempted to address the issue by creating a separate JavaScript (JS) file named "popover.js" and calling it within the "filterState" function using the "include_js_files" method. However, I encountered a problem where the popover functionality only applied to the first defined help link, while the others did not display the popover. To overcome this issue, I decided to keep the script inline within the HTML code.

LEt me know you thoughts and suggestions


kartikeyakirar:
closing this PR https://github.com/insightsengineering/teal.slice/pull/333

kartikeyakirar:
#289 
> Debounce
> Some modules in tmc/tmg/tmgoshawk have debounce added to plots, which take the longest to refresh. I am not against adding debouncing in the filter panel. Since this is being thought about on a higher level, I would consider it out of scope. We would also have to check whether it stacks.

@chlebowa I just read your comment about debounce, and I completely agree that this issue should be addressed at a higher level, especially in cases involving numeric ranges and input fields with up and down buttons. When users rapidly change the inputs, it can result in increased latency. That's why I thought about incorporating debounce. Please let me know if we should retain it for now until it is implemented at higher levels, or if we should refactor it from this point.

R/FilterStateRange.R
```diff
@@ -456,10 +456,29 @@ RangeFilterState <- R6::R6Class( # nolint
...
           div(
-            actionLink(ns("plotly_info"), label = NULL, icon = icon("question-circle")),
+            actionLink(
+              ns("plotly_info"),
+              label = NULL,
+              icon = icon("question-circle"),
+              "data-toggle" = "popover",
+              "data-html" = "true",
+              "data-placement" = "left",
+              "data-trigger" = "hover",
```
chlebowa, line 473:
```suggestion
              "data-trigger" = "click",
```

R/FilterStateRange.R
```diff
@@ -456,10 +456,29 @@ RangeFilterState <- R6::R6Class( # nolint
...
+              "data-content" = '<p>Drag vertical lines to set selection.<br>
+                              Drag across plot to zoom in.<br>
+                              Drag axis to pan.<br>
+                              Double click to zoom out.</p>'
```
chlebowa, line 477:
```suggestion
              "data-content" = '<p>
                                <h5> Plot actions </h5>
                                Drag vertical lines to set selection.<br>
                                Drag across plot to zoom in.<br>
                                Drag axis to pan.<br>
                                Double click to zoom out.
                                </p>'
```

Can you remove the excess top and bottom padding from the popover?

chlebowa (commented):
I love the popover, added some suggestions in the code.

Let's wait for Chendi's decision regarding the debounce.

R/FilterStateRange.R
kartikeyakirar: Added additional JS code to close popover if clicked anywhere else on screen

lcd2yyz:
Let's keep the debounce since it's already implemented. The range value input is probably the place where it's needed the most any ways. This will give the users a chance to report on any negative experience from the debounce mechanism, before we implement at higher level. 

But just one small detail to confirm - the debounce is only added for the range input, but not on the plotly correct? I can feel the delay if I rapidly clicking the up/down arrow, however, the computing start right away when I move draggable lines.

Popover feels very nice, and the reactive area is small enough that it doesn't interfere with normal usage of the plotly itself.
But there is one extra empty line appearing for some reason, please remove. 
image

kartikeyakirar:
@cicdguy 

> But just one small detail to confirm - the debounce is only added for the range input, but not on the plotly correct? I can feel the delay if I rapidly clicking the up/down arrow, however, the computing start right away when I move draggable lines.

Yes, debounce is only added for the range input.I did not add it on plotly draggable bar as it seems different from range selection where the user's typical behavior may involve continuously clicking to obtain a specific number. We can add debounce functionality here as well if you think it would be beneficial.

> Popover feels very nice, and the reactive area is small enough that it doesn't interfere with normal usage of the plotly itself.
But there is one extra empty line appearing for some reason, please remove.

I have removed bottom white space. 

image

R/FilterStateRange.R
```diff
@@ -457,10 +457,40 @@ RangeFilterState <- R6::R6Class( # nolint
       tagList(
         div(
           class = "choices_state",
+          tags$head(tags$script(
```
chlebowa, line 460: This script suddenly grew to a substantial size. Please ad a short comment explaining why it has to be inline rather than in a separate .js file.

R/FilterStateRange.R
```diff
@@ -457,10 +457,40 @@ RangeFilterState <- R6::R6Class( # nolint
...
+            actionLink(
+              ns("plotly_info"),
+              label = NULL,
+              icon = icon("question-circle"),
+              "data-toggle" = "popover",
+              "data-html" = "true",
+              "data-placement" = "left",
+              "data-trigger" = "click",
+              "data-title"   = "Plot actions",
+              "data-content" = '<p>
```
chlebowa, line 485: Is separating the header from the bode necessary in any way? Just curious.

inst/css/filter-panel.css
```diff
@@ -357,3 +357,20 @@ a.remove_all:hover {
 .choices_state {
   margin-bottom: .45rem;
 }
+
+.popover-header {
+    padding: .5rem .75rem;
+    margin-bottom: 0;
+    font-size: 1rem;
+    background-color: #f7f7f7;
+    border-bottom: 1px solid #ebebeb;
```
chlebowa, line 366: I am not sure hard coding colors is the right way to go. Will they not default to theme colors? Nothing much seems to change when I delete these lines.

R/FilterStateRange.R
kartikeyakirar: added comment.

Issue #336, opened 2023-06-14
filter-card-title is partially covered by the disable switch


kartikeyakirar:
As per  @chlebowa  finding , it has been observed that the variable label in the first card is being partially obscured by the disable switch. Although this specific concern will be resolved soon with the removal of those switches in branch 135, the core issue persists. The problem lies in the rendering order of the div elements involved. The div containing the switch and the x mark `(filter-card-controls)` is displayed on top of the div containing the header` (filter-card-title)`. This leads to unintended visual overlap, impacting the usability and aesthetics of the interface.

![image](https://github.com/insightsengineering/teal.slice/assets/6700955/b6201f75-eaea-49ca-a176-a5499e7c6dd9)

kartikeyakirar:
The issue of overlapping elements occurs only when the screen is minimized to a certain size. To demonstrate this behavior across resizing screen sizes, I have created a demo video showcasing the overlapping issue. I think we should consider and discuss the expected behavior of the interface across all screen sizes. 

https://github.com/insightsengineering/teal.slice/assets/6700955/616e226a-42c9-4759-860e-772e8f44698c

gogonzo:
Not relevant anymore after https://github.com/insightsengineering/teal.slice/pull/300

PR #354, opened 2023-06-21
298 reset button

### LINKED ISSUE #298: [Research] A reset button to restore filters to default state
### Summary
As a user, I would like to restore the filters to their original default settings, just like when the app was first run.

### Acceptance Criteria
1. Conduct research to identify the necessary components and approaches for enabling the desired feature. 
2. Propose steps to complete the feature in small pieces. Breaking down the implementation into manageable steps can help ensure a systematic and organized approach

chlebowa:
Closes #298 

##### State history
+ `FilterState` receives a new private field, `private$state_history`, that holds a `reactiveVal` that contains a `list`. This list stores all states that this `FilterState` has had since istantiation. States are stored not as `teal_slice` but as lists because `teal_slice` are now `reactiveValues` that can be shared between `FilteredData` objects.
+ Updates to `private$get_selected` , `private$get_keep_na`, and `private$get_keep_inf` cause the current state to be appended to state history.
+ `FilterState` ui module receives two buttons, `back` and `reset`, placed next to the `remove` button.
+ The `reset` button sets the state to the one with in which `FilterState$server` started.
+ The `back` button sets the state to the penultimate one stored in `private$state_history`.
+ Since we do not envision a `forward` button, state history is only stored backwards, i.e. clicking the `back` button erases the last stored state.
+ The `setdiff_teal_slices` function is added to assist in removing subsets of filters.
+ Some behavior in getting and setting filter states is modified to be more amenable to module-specific situations.
+ Changes in `FilterStates` cause server updates when new states are added, not just when states with new names are added. The previous state caused pointers to removed filter states to linger and new states could access private fields of the removed ones. Solved by @gogonzo 

Note that buttons become active only when applicable (i.e. state history is of length > 1).

#### TESTING
```
# rm(list = ls()) # uncomment to clear environment before each app start

options(teal.log_level = "WARN", teal.show_js_log = FALSE)
options("teal.bs_theme" = bslib::bs_theme(version = 3))
# options(shiny.trace = TRUE)
devtools::load_all("../teal.slice")
devtools::load_all("../teal")


  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      output$filter_states <- renderText({
        stage1 <- filter_panel_api |> get_filter_state()
        stage2 <- lapply(stage1, as.list)
        stage3 <- yaml::as.yaml(stage2)
        stage3
      })
        attr(data, "code")()
      })
    }
  )
}


Iris <- iris
levels(Iris$Species) #%<>% toupper()
Iris$Species2 <- Iris$Species
Iris$Species3 <- Iris$Species
Iris$Sepal.Length[1] <- NA
Iris$Sepal.Length[2] <- Inf
Iris$Species3[2] <- NA
attr(Iris[[1]], "label") <- "variable label"
attr(Iris[[2]], "label") <- "variable label"
attr(Iris[[3]], "label") <- "variable label"
attr(Iris[[4]], "label") <- "variable label"
data <- teal_data(
  dataset("iris", Iris),
  dataset("mtcars", mtcars)
)

default_filters <- teal_slices(
  teal_slice("iris", "Sepal.Length", sel = c(5.4, 6.1), fixed = FALSE),
  teal_slice("iris", "Sepal.Width", sel = c(2.5, 4.0), fixed = FALSE, anchored = TRUE),
  teal_slice("iris", "Petal.Length", sel = c(4.5, 5.1), fixed = TRUE),
  teal_slice("iris", "Petal.Width", sel = c(0.3, 1.8), fixed = TRUE, anchored = TRUE),
  teal_slice("iris", "Species", sel = c("versicolor", "virginica")),
  # teal_slice("iris", "Species2", sel = c("versicolor", "virginica"), fixed = TRUE),
  # teal_slice("iris", "Species3", sel = c("versicolor", "virginica"), anchored = TRUE),
  count_type = NULL
)

app <- init(
  data = data,
  modules = list(funny_module(), funny_module("module2")), filter = default_filters
)

```


R/FilterState.R
```diff
@@ -254,13 +301,34 @@ FilterState <- R6::R6Class( # nolint
...
             tags$span(private$get_varlabel(), class = "filter-card-varlabel")
           ),
-          if (isFALSE(private$is_locked())) {
-            actionLink(
-              inputId = ns("remove"),
-              label = icon("circle-xmark", lib = "font-awesome"),
-              class = "filter-card-remove"
-            )
-          },
+          div(
```
gogonzo, line 304: I think it should be a dropdown menu with these buttons instead of listing two icons, what do you think @chlebowa @lcd2yyz ?

There is also CSS issue with long label which goes under these icons. CSS rule is made in a way that label is broken into another line when it reaches remove button.

image

gogonzo (commented):
PR introduces a new BUG when adding a new state from the column. Details in the comments below.

R/FilterState.R
chlebowa: > I think it should be a dropdown menu with these buttons instead of listing two icons, what do you think @chlebowa @lcd2yyz ?

A dropdown menu means two clicks per action so if one wants to go back three steps, that would be six clicks.

> There is also CSS issue with long label which goes under these icons. CSS rule is made in a way that label is broken into another line when it reaches remove button.
> 
> image

Should we reopen [this](https://github.com/insightsengineering/teal.slice/issues/336) then?

lcd2yyz: I like the two icons, it's easy access and elegant, provided that we can 1) fix CSS of the variable label, 2) add tooltip on hover with delay/expire to show prompt like "Undo" and "Reset".

lcd2yyz:
Please update the testing code in the original post, it's a bit outdated as some function names have changed.

Below are some prelim comments and issues I've noticed thus far. It's not a thorough review yet, I will try to find more time this week to do some more tests, please expect additional comments to come. 

I have not really played with module-specific settings yet. In general, I've experienced quite a lot of bugs with app/filter behavior, it feels very much work-in-progress, rather than ready-for-review state at the moment. For your reference, I'm again testing on BEE R4.1.1, so it's possible some of these are system/version related.

1) When there are no fixed filter defined, the "Add filter" popup is showing fixed + interactive filter in one-line as title. Also note spelling of "Interactive".
image

2) When I try to add the default but inactive filters, in rapid succession, it goes into a reactive loop. I have to click on one of the filter cards to stop the reactivity, but then only that card get activated.

https://github.com/insightsengineering/teal.slice/assets/78242757/f6de63e6-3112-4644-8c4d-b0fd69e7a8cf

3) The "reset" or "create snapshot -> apply" in the global filter manager doesn't seem to work. It's not crashing the app, but app stuck in a continuous "computing state" (window grayed out), but I still able to interact with the filter cards to some degree. Because of this freezing behavior, I have not been able to test as much as I wanted. 

https://github.com/insightsengineering/teal.slice/assets/78242757/cecf7dd1-0f90-4d06-9fb7-55c80f02d2fc

A general comment about the filter manager is that the feature and UX needs to be largely improved still. Please spend some time think about more complex (and realistic) usage, when there are 20 modules and 10 different default filters, the matrix table can get quite confusing and difficult to read. 

On the plus side, icons + tooltip for snapshot manager feels nice and easy to understand, I really like the clean look. One small inconvenience is that once I enter a name for the snapshot, the whole filter manager window disappears, I would have to go to gear button to re-open to see the new snapshot that I created.

4) If a numerical variable is already in standby filter list, but user is not aware such filter exist, and goes to activates a second copy by adding a new filter variable. The plotly doesn't load on this new filter card. 
However, this does not happen if user actives the standby filter, removes it, then adds a second copy. 

https://github.com/insightsengineering/teal.slice/assets/78242757/23f9e565-4852-4b87-a4c0-288ecf869438

5) When user activates second copy by adding a new filter variable, the newly create filter card cannot be removed, the under/reset doesn't appear either. Also note in this example video below, the selected values does not appear for the categorical `iris.Species` variable either, which is odd.

https://github.com/insightsengineering/teal.slice/assets/78242757/35984618-2241-4389-a7a2-a25286c1666a

R/FilteredData.R
```diff
@@ -138,7 +139,39 @@ FilteredData <- R6::R6Class( # nolint
...
+    #' Activate snapshot manager module to allow capturing, saving, and restoring
+    #' snapshots of the global application filter state.
+    #'
+    #' @param x (`logical(1)`) flag
+    #' @return invisible `NULL`
+    activate_snapshot_manager = function(x) {
+      checkmate::assert_flag(x)
+      private$snapshot_manager_active <- x
+      invisible(NULL)
+    },
```
gogonzo, line 170: I think we could expose this as an api attribute, e.g. `module_snapshot`. I know it might be tricky as:
- `teal::teal_slices(module_snapshot)` would call `teal::snapshot_manager_ui/srv` only when `module_specific = TRUE`
- `teal.slice::teal_slices(module_snapshot)` would call `FilteredData::snapshot_manager` (through `ui/srv_active` module).

We can discuss this soon

lcd2yyz:
> Points 1, 2, 4, and 5 are not related to this PR. They are problems of the module-specific implementation. Please raise separate issues.

Point taken, sorry for jumping the gun. Is there an associated open issue that I can report it to? Or should I open a new issue altogether? I will focus more on the reset/undo mechanism from now on.

> Point 3 - I cannot reproduce this. Does the console say anything when the screen grays out?

The change you made last night made it working quite nicely now! 
But at the same time, I see that the undo/reset button on the filter cards have disappeared. Also the "apply" icon is missing from the snapshot manager, although the action button is still there, because I can click in the blank space and it's restoring the filter setting.
image

> I added a programmatic click to reopen the filter manager, see if you like it.

Perfect!

lcd2yyz:
On a related note, it is possible to make the popup disappear if other buttons are clicked? Basically what I have in mind is the following:
- Click camera icon => window pops up to input snapshot name => click accept => goes back to filter manager popup (exactly as we have now, this will allow the user to see the snapshot they created and able to download filter config right away)
- Click save to file => initiate download => filter manager stays (as we have now, users might want to download multiple config at once)
- Click reset button => all filters reset to initial state, and filter manager closes automatically (right now user has to make another click to close)
- Click apply button => restore to snapshot filter setting, and filter manager closes automatically (right now user has to make another click to close)
For the latter two action, both are to re-apply filter setting, I can't think of any follow-up actions user would want to do other than expecting to see the app/module refreshes.

For "save to file" function, is it possible to allow user to name the file? Using `teal_filtersnapshot_<name of snapshot>_<datetime stamp>` as the default.

chlebowa:
I modified file name and dialog closing as requested. 

Filter state history is not part of the snapshot and therefore is reset when a snapshot is restored. This is because the underlying `FilterState` objects are deleted and new ones are created in their stead, according to the snapshot state.

I don't know why icons don't show up for you, but I do recall you have had this problem before when running on BEE.

The filter manager issues - Dawid is fixing them as we spot them. We are due to merge the refactor branch to main this week, I suggest we wait until then and start logging issues in the `teal` repo is they persist.

lcd2yyz:
> I modified default file name and dialog closing as requested.

Nice flow, feels awesome!

> I don't know why icons don't show up for you, but I do recall you have had this problem before when running on BEE.

The icon was there before, the first time I looked at it (when it was stuck in the computing state), which is why I know there was something there. I didn't change code or anything in between the two reviews :confused:

chlebowa:
Now that `filter_panel_refactor@main` has been merged, I invite all of you to review again.

R/FilteredData.R
```diff
@@ -1223,6 +1262,162 @@ FilteredData <- R6::R6Class( # nolint
...
+          showModal(
+            modalDialog(
+              textInput(
+                session$ns("snapshot_name"),
+                "Name the snapshot",
+                width = "100%",
+                placeholder = "Meaningful, unique name"
+              ),
+              footer = tagList(
+                actionButton(session$ns("snapshot_name_accept"), "Accept", icon = icon("thumbs-up")),
```
gogonzo, line 1319: Please use `ns` everywhere in this module - as it is in teal::srv_snapshot_manager
```suggestion
                actionButton(ns("snapshot_name_accept"), "Accept", icon = icon("thumbs-up")),
```

R/FilteredData.R
```diff
@@ -1223,6 +1262,162 @@ FilteredData <- R6::R6Class( # nolint
...
+
+        observeEvent(snapshot_history(), {
+          lapply(names(snapshot_history())[-1L], function(s) {
+            id_pickme <- sprintf("pickme_%s", make.names(s))
+            id_saveme <- sprintf("saveme_%s", make.names(s))
+            id_rowme <- sprintf("rowme_%s", make.names(s))
+
+            # Create observer for restoring snapshot.
+            if (!is.element(id_pickme, names(observers))) {
+              observeEvent(input[[id_pickme]], {
```
gogonzo, line 1378:
```suggestion
              observers[[id_pickme]] <- observeEvent(input[[id_pickme]], {
```

R/FilteredData.R
```diff
@@ -1223,6 +1262,162 @@ FilteredData <- R6::R6Class( # nolint
...
+                  snapshot_state <- reassemble_slices(snapshot)
+                  teal.slice::slices_store(tss = snapshot_state, file = file)
+                }
+              )
+              handlers[[id_saveme]] <- id_saveme
+            }
+
+            # Create a row for the snapshot table.
+            if (!is.element(id_rowme, names(divs))) {
+              div(
```
gogonzo, line 1406:
```suggestion
              divs[[id_rowme]] <- div(
```

R/FilteredData.R
chlebowa: Gladly! I thought there was a convention module servers within `R6` classes usually (always?) use `session$ns`.

R/FilteredData.R
```diff
@@ -1223,6 +1262,162 @@ FilteredData <- R6::R6Class( # nolint
...
+              type = "message"
+            )
+            updateTextInput(inputId = "snapshot_name", value = , placeholder = "Meaningful, unique name")
+          } else {
+            snapshot <- disassemble_slices(self$get_filter_state())
+            snapshot_update <- c(snapshot_history(), list(snapshot))
+            names(snapshot_update)[length(snapshot_update)] <- snapshot_name
+            snapshot_history(snapshot_update)
+            removeModal()
+            shinyjs::click(id = "teal-main_ui-filter_panel-active-show_snapshot_manager", asis = TRUE)
```
gogonzo, line 1347: I'd prefer modal to be closed with notification instead or reopening with js shot based on the id which package has no control of 
```suggestion
```

R/FilteredData.R
gogonzo: No convention, I just checked diffs between this module and one in teal and it was one of them

R/FilteredData.R
chlebowa: The request was that first modal (snapshot manager) remain open to facilitate immediate further actions but it is closed as the second one (snapshot name) is opened. I agree that this method is not ideal to say the least.

gogonzo: Line doesn't guarantee that modal will be open as we might change `teal` namespaces and we would never realise that it doesn't work anymore. This line has a dependency on teal (reverse) and is impossible to test. The request is understandable but we can't guarantee this to work in a long term. 
@lcd2yyz what do you think if alternatively we close the window and show a notification that "Snapshot has been added"

gogonzo (commented):
`teal` and `teal.slice` contains almost the same module and maintaining them in both repos in a long run will be hard (I understand the reason why they differ and why we can't easily make only one and export).

What we should do in the future before making any further changes in `FilteredData` or `teal::srv/ui_filter_manager`:
1. enable teal::ui/srv_filter_manager also for a `module_specific = FALSE` and remove `FilteredData$ui/srv_snapshot_manager`
2. enable `FilteredData$ui/srv_available_filters` for a `module_specific = FALSE`
3. There might be two variants of module_filter_manager and module_snapshot_manager to address above.

My feeling is that we are doing wrong by enabling/disabling `available_filters` and `snapshot_manager` in `FilteredData` depending on some configuration in `teal`. We should always initialize `FilteredData` in the same way and adjust in `teal` accordingly. 

In general, I think this PR fullfils the requirement and looks good actually. I'm fine with merging this PR if other parties feel good, but I'd create a new issue move this functionality out of `FilteredData` to `teal` based on (1), (2), (3) above

-----

Alternative could be to think more about ⚙️ icon in the `module_active`, but it will be very hard to enable for `module_specific` (even though I initially signalled this as a good idea). Reason:
- Imagine we have a `module_specific` filter-panel. Each module have filters which are activated from a shared list (`slices_global`).
- Restoring slices from one module would introduce completely new states to the shared list, and among of them there could be filters which duplicate existing filters in other modules.

It's hard to imagine solution based on restoring filters within a module which wouldn't be confusing.

chlebowa:
Snapshot manager has been removed. Snapshots will be handled by `teal`, where the `snapshot_manager` module will be called by the `filter_manager` module.

gogonzo (approved):
This is amazing!

Module specific
image

Global filters
image

@lcd2yyz please have a look

R/FilteredData.R
lcd2yyz: Thanks for explaining. While I do want to have the re-open action, future reliability is also important, so your proposal sounds reasonable compromise to me.

lcd2yyz:
Snapshot manager inside the global filter manager feels concise and smooth, all four buttons and associated functionalities are working quite nicely as expected.

But I noticed an interesting bug on the reset button on the individual filter card.
Essentially if the variable (`Sepal.Length`) has a default condition set, then the reset button works fine as expected. 
However, if no default was set (`Species` & `mpg`), then
- Clicking on reset button has no effect
- Undo button can only return to the second step, effectively it only "remember" filter state after the first user interaction. 

https://github.com/insightsengineering/teal.slice/assets/78242757/ad978381-cf4f-4e2b-a652-5d2732ab3395

For reference, the default filter setting I used is as below
```
default_filters <- teal::teal_slices(
  teal_slice("iris", "Sepal.Length", sel = c(5.5, 6.5), fixed = FALSE),
  teal_slice("iris", "Species"),
  teal_slice("mtcars", "mpg")
  )
```

chlebowa:
Great catch @lcd2yyz, thanks!
The reason was state history was initiated in the `FilterState` constructor by adding the `slice` that was used for instantiation but a a `teal_slice` with `selected = NULL` was used, there was no state to return to.
I changed the constructor to initiate state history as an empty list and now the server method adds the initial state as the first element of history.

In short: fixed.

Issue #299, opened 2023-05-25
Decouple `scda` from teal.slice


donyunardi:
Related to https://github.com/insightsengineering/nestdevs-tasks/issues/1

### Summary
After careful observation, it appears that the scda package is not essential for teal.slice. Consequently, it is advantageous for us to separate scda from teal.slice.

There are currently no R or unit test files that utilize the scda package. Therefore, it is sufficient to remove scda from the DESCRIPTION file and staged_deps.

### Acceptance Criteria
* Update DESCRIPTION file to remove `scda` and related packages.
* Update staged_dependencies.yml

PR #357, opened 2023-06-22
299 decouple scda from teal.slice

m7pr:
A resolution for #299 where we removed package from DESCRIPTION file and staged_dependencies.yaml as the package is not used anywhere in particular.


staged_dependencies.yaml
```diff
@@ -12,12 +12,6 @@ upstream_repos:
   insightsengineering/teal.data:
     repo: insightsengineering/teal.data
     host: https://github.com
-  insightsengineering/scda:
-    repo: insightsengineering/scda
-    host: https://github.com
-  insightsengineering/scda.2022:
-    repo: insightsengineering/scda.2022
-    host: https://github.com
```
gogonzo, line 20: changing `staged_dependencies.yaml` need to be mirrored in other packages also as staged.dependencies is symmetrical.


https://github.com/insightsengineering/scda/blob/097dcfcb3bfb02e85730fbd2bc575114c26b5ac1/staged_dependencies.yaml#L15

https://github.com/insightsengineering/scda.2022/blob/13284e884111bf59be225196171da5a0551956e0/staged_dependencies.yaml#L27

m7pr: I wasn't aware of that. Thanks

m7pr: https://github.com/insightsengineering/scda.2022/pull/121
https://github.com/insightsengineering/scda/pull/123

m7pr: @gogonzo I added to PRs - if you can accept it, then we can merge all 3

pawelru:
@m7pr let me do the merge here

Issue #343, opened 2023-06-15
srv_available_filters dropdownMenu initializes partially off the screen


gogonzo:
`ui_available_filters` initializes slightly off the screen when clicked for the first time.

 
<img width="367" alt="Skärmavbild 2023-06-18 kl  23 53 55" src="https://github.com/insightsengineering/teal.slice/assets/6959016/5ed68c92-d095-445d-9860-5632f5af3082">

```r
library(teal.slice)
library(teal)

funny_module <- function (label = "Filter states", datanames = "all") {
  checkmate::assert_string(label)
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderPrint({
        logger::log_trace("rendering text1")
        filter_panel_api %>% get_filter_state()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}

ADSL <- synthetic_cdisc_data("latest")$adsl
ADSL$numeric <- runif(nrow(ADSL))
ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- Inf

default_filters <- teal::teal_filters(
  filter_var(dataname = "ADSL", varname = "numeric"),
  mapping = list(`filter module` = "ADSL numeric")
)

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL)
  ),
  modules = modules(
    funny_module("filter module")
  ),
  filter = default_filters
)


runApp(app)
```

kartikeyakirar:
PR is merged.

PR #365, opened 2023-06-28
bug fix: srv_available_filters dropdownMenu initializes partially off the screen

kartikeyakirar:
This PR fixes: https://github.com/insightsengineering/teal.slice/issues/343

To address the issue caused by the dynamic transformation of the tippy_popper CSS class, particularly the transform property where the first element is dynamically changing based on screen size but on initialization its not behaving correctly,
image

The transform property was not calculating the correct position in the DOM during the initial setup in these scenarios. I attempted to make it more static by fixing the transformX property, but this solution only worked properly in the regular version or with Bootstrap 4/5.

I delved deeper into other options available with tippy.JS, which can be found at [atomiks.github.io/tippyjs/v5/all-props/](https://atomiks.github.io/tippyjs/v5/all-props/), specifically exploring the dropMenuOptions ([dropMenuOptions](https://dreamrs.github.io/shinyWidgets/reference/dropMenuOptions.html)). However, none of these options provided the desired result.

Upon further investigation, I discovered that increasing the width of the displayed section correctly calculated the position of the element in comparison. Therefore, I made modifications to the inner element property to maximize its width, which successfully resolved the current issue. 

Personally, I'm not entirely satisfied with this solution, but it is the only workaround that fixes the issue in all versions of Bootstrap and the regular version. If you have any suggestions for alternative ways to address this issue, I would greatly appreciate it.

for bs version
`options("teal.bs_theme" = bslib::bs_theme(version = 4))`

testing app
```
library(teal.slice)
library(teal)
library(scda)
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
      )
    },
      output$filter_states <-  renderPrint({
        filter_panel_api %>% get_filter_state()
      })
        attr(data, "code")()
      })
    }
  )
}

ADSL$numeric <- runif(nrow(ADSL))
ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- NA
ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- Inf

default_filters <- teal::teal_filters(
  filter_var(dataname = "ADSL", varname = "numeric"),
  mapping = list(`filter module` = "ADSL numeric")
)

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL)
  ),
  modules = modules(
    funny_module("filter module")
  ),
  filter = default_filters
)


runApp(app)
```


chlebowa:
I guess the problem now is that the content of the popup may not fit into the preset width. As long as our state ids are concatenated with spaces, they will be wrapped, but we face a potential overflow if they are concatenated with underscores or are otherwise very long strings.
Also, we do not really understand why this is happening, do we? Only that setting a fixed width magically fixes our problem?

I would tentatively accept this solution for the time being.

kartikeyakirar:
The content will adjust within a fluid format without overflowing from the popup. Here's a small test I conducted by adding a long string within an h3() tag.

image

It appears that the defined element  for tippy.js and popper.js to calculate its exact position in the DOM. However, when using shinydashboard++ and bs4dash, there seems no issue ( i used right-sidebar near to same position.) 

So yes we don't fully understand whats ahppening but adjusting width seems to resolve it.

chlebowa:
Now do that again but with no whitespace.

kartikeyakirar:
> Now do that again but with no whitespace.

Testing long string result for ref.
image


Now I see this issue. thanks. Instead of width I am setting min-width so that solution will be applicable to both current situation and for long string.

chlebowa:
Great!

gogonzo (approved):
There is small difference between first and second time when any of the choice has a length > 15em, but I expect this to be very rare.

PR #377, opened 2023-07-10
duplicated available slices

### LINKED ISSUE #374: filter conflict crashes app
Adding filter from available filters crashes app when a filter on that variable already exists, having been added by the user.

https://github.com/insightsengineering/teal.slice/assets/114988527/f15bca89-9269-46c8-a587-f938cd44b325
```
Warning: Error in lapply: argument is missing, with no default
  1: runApp
```

gogonzo:
Closes #374 

Fixes problem when adding new filter based on the variable for which there is already filter in `available_teal_slices`. Filters duplicated are disabled from the selection. Filter is disabled when:
- there is already filter which is based on the same variable. Determining if filter refers to the same column is based on `dataname`, `varname`, `experiment` and `arg`. If there is any filter which has exactly the same values in all of these fields then it is duplpicated.
- when non of these duplicated filters is active then it is possible to select any of them (nothing is disabled)
- when any of duplicated filters is active then other (duplicated) filters are disabled

Fixed in this PR:
- when using `$set_filter_states`, `teal_slice` attributes were changed only for datasets which had `teal_slice` object - loop was done on slices-datasets instead of all datasets.
- changed slice_list_remove to accept vector instead of scalar as documentation stated `character`. This also matches logic in `state_list_add` which also accepts multiple elements.

image


R/FilteredData.R
```diff
@@ -1113,45 +1123,53 @@ FilteredData <- R6::R6Class( # nolint
...
-              after = 0
-            )
-          })
+          htmltools::tagInsertChildren(
+            checkbox,
+            br(),
+            tags$strong("Fixed filters"),
+            non_interactive_choice_mock,
+            tags$strong("Iteractive filters"),
+            interactive_choice_mock,
```
chlebowa, line 1169: If the `tag$strong` are moved into the function that generates the mock elements, it would pre-emptively solve  this bug:
image

chlebowa (approved):

PR #387, opened 2023-07-17
350 audit set selected

### LINKED ISSUE #350: audit `FilterState$set_selected`
`$set_selected` involves multiple steps of validating the selected value. Investigate whether all of those steps are strictly necessary. `$cast_and_validate` and `$validate_selection` are at least partially redundant.

Start by understanding `$set_selected` and the methods that it calls.
Compare across subclasses.
Remove unnecessary methods/functions.

chlebowa:
Fixes #350 

In this PR:
+ `$validate_selection` method is removed from all `FilterState` classes.
+ `$check_multiple` is renamed to `$check_length`
+ `$set_selected` sequence is changed so that `$check_length` is called before `$remove_out_of_bound_values`
+ `$check_length` is added to `RangeFilterState`, `DateFilterState` and `DatetimeFilterState` to ensure selection is of length 2 and sorted


chlebowa:
For range-type filter states checking whether the selection is sorted has different behaviors: unsorted numerics raise an error, while unsorted dates and date-times raise a warning and reset the maximum range. I believe these should be unified. The question is, which one should we apply. Any thoughts?

_Nota bene_, choice-type filter states also check if multiple values are allowed and if not - they abort setting selection and reset the previous (valid) one.

gogonzo (approved):
Please remove `check_in_subset` and `check_in_range` as they are not used anymore. Otherwise I didn't find any issues, good to go.

PR #390, opened 2023-07-18
447 remove formatters

### LINKED ISSUE #389: Remove formatters from package dependencies
Part of https://github.com/insightsengineering/coredev-tasks/issues/447

There are two functions used from formatters
- `formatters::var_labels`
- `formatters::var_relabel`

We should create a copy of this function in `teal.widgets` and export them there.

Acceptance criteria:
- remove formatters from dependencies
- update staged.dependencies.yaml files in all affected packages

m7pr:
Closes #389 which is a follow-up of https://github.com/insightsengineering/coredev-tasks/issues/447

@gogonzo I decided to copy the body of the function and expose/export it in teal.slice. We can also not export it and use with ::: because this is used only in one place. However if we ever apply for CRAN upload, then ::: is forbidden.

Once all `teal.*` packages get rid of `formatters` I will also merge staged.dependencies.yml reduction in formatters https://github.com/insightsengineering/formatters/pull/183


chlebowa:
If you define a function in your package, you don't need `:::` :thinking:

gogonzo:
@m7pr 
why:
- in teal.slice?
- why to export and reuse at all?

I think this function fits better to `teal.data` or `teal.widgets` if needs to be exported.  It's hard to defend that handling labels is related with a package which filters data.

m7pr:
@gogonzo oh yeah, total absurd from my side, this should go to teal.widgets which are imported by all other packages

m7pr:
Hey @gogonzo I decided to expose `var_labels` as `formatters_var_labels` in `teal.widgets`. Will move to other packages and will continue updating `teal.widgets`'s PR accordingly when we need more functions

gogonzo (approved):
👍

Issue #381, opened 2023-07-11
remove `slices_field` function


chlebowa:
Remove the function `slices_field` from the package and replace it with appropriate `vapply` or `lapply` calls to `[[`, e.g.
```
slices_field(slices, "experiment")
unique(vapply(slices, `[[`, character(1L), "experiment"))
```

m7pr:
@chlebowa slices_field is lit! IMHO slices_field is easier to read than unique(vapply(

chlebowa:
I tend to agree but we decided to remove unnecessary sugar.

kartikeyakirar:
Both unique(vapply(slices,[[, character(1L), "experiment")) and unique(unlist(lapply(slices, function(x) x[["experiment"]]))) are valid ways to find the unique experiments in a list of slices. However, there are some subtle differences between the two approaches.

The first approach uses the vapply() function to apply the [[ operator to each slice in the list, and then uses the unique() function to remove any duplicate values. This approach is more concise, but it can be less efficient if the list of slices is large.

The second approach uses the lapply() function to apply the x[["experiment"]] expression to each slice in the list, and then uses the unlist() function to flatten the resulting list of lists into a single list. This approach is less concise, but it can be more efficient if the list of slices is large.

<html>
<body>


Approach | Pros | Cons
-- | -- | --
unique(vapply(slices,[[, character(1L), "experiment")) | More concise | Less efficient for large lists
unique(unlist(lapply(slices, function(x) x[["experiment"]]))) | More efficient for large lists | Less concise


</body>
</html>

any thoughts? @chlebowa @gogonzo

chlebowa:
> but it can be more efficient if the list of slices is large.

I don't know about that :thinking: 

The rule of thumb is: always go with `vapply` unless you can't because some elements can return `NULL` and you get errors on `FUN.VALUE`, then go with `lapply`. You will find that all slices have `id` but not all have `experiment`, for instance.

PR #391, opened 2023-07-18
381 remove slices_field function

kartikeyakirar:
this PR fixes https://github.com/insightsengineering/teal.slice/issues/381


R/FilterStates.R
```diff
@@ -271,7 +271,7 @@ FilterStates <- R6::R6Class( # nolint
         }
 
         # Drop teal_slices that refer to excluded variables.
-        varnames <- slices_field(state, "varname")
+        varnames <- unique(unlist(lapply(state, function(x) x[["varname"]])))
```
gogonzo, line 274: FYI @chlebowa I think lapply is fine as not all `teal_slice` objects have `"varname"`. This means we can't use `vapply` in all cases

R/FilterStates.R
```diff
@@ -438,7 +438,7 @@ FilterStates <- R6::R6Class( # nolint
           avail_column_choices <- reactive({
             data <- private$data
             vars_include <- private$get_filterable_varnames()
-            active_filter_vars <- slices_field(self$get_filter_state(), "varname")
+            active_filter_vars <- unique(unlist(lapply(self$get_filter_state(), function(x) x[["varname"]])))
```
gogonzo, line 441: shorter version
```suggestion
            active_filter_vars <- unique(unlist(lapply(self$get_filter_state(), `[[`, "varname")))
```

R/FilteredData.R
```diff
@@ -486,7 +486,7 @@ FilteredData <- R6::R6Class( # nolint
         }
 
         checkmate::assert_class(state, "teal_slices")
-        datanames <- slices_field(state, "dataname")
+        datanames <- unique(vapply(state, "[[", character(1L), "dataname"))
```
gogonzo, line 489: Here we can use `vapply` as all `teal_slice` objects have "dataname" 👍

R/FilterStates.R
chlebowa: [My point exactly.](https://github.com/insightsengineering/teal.slice/issues/381#issuecomment-1640176164)

kartikeyakirar:
> 👍 I'd prefer `lapply(, '[[', "field")` instead of `lapply(, function(x) x[["field"]])`. But PR fixes the "issue"

I have updated in all the calls.

Issue #340, opened 2023-06-15
[BUG] Forbid `teal_slices` for `excluded_varnames`


gogonzo:
`teal_slices` has `included_varnames` and `excluded_varnames` argument which says which variables can be added as filters.
There is a possibility in `teal_slices` to specify `teal_slice` for `varname` which can't be filtered. Please make a proper assertion.

```r
teal_slices(
  teal_slice(dataname = "iris", varname = "Sepal.Length")
  exclude_varnames = list(iris = "Sepal.Length")
)

teal_slices(
  teal_slice(dataname = "mae", varname = "age"),
  teal_slice(dataname = "mae", experiment = "experiment", varname = "gene.id"),
  exclude_varnames = list(
    mae = list(subjects = "age", experiment = "gene.id")
  )
)

```

gogonzo:
I think we need to close this issue as we can't control slices for all types of datasets. `teal`'s objective is to be data-type independent, so relying on `experiment` and `arg` (both are passed through `...`) shouldn't happen.

kartikeyakirar:
Alright!

PR #393, opened 2023-07-20
340 Forbid teal_slices for excluded_varnames

kartikeyakirar:
this fixes: https://github.com/insightsengineering/teal.slice/issues/340

In PR, I have implemented updates to the logic of teal_slices by adding an assertion to guarantee that varnames are distinct from excluded_varnames. This logic iterates over each dataname to verify the assertion.

Furthermore, there is another similar checkin `set_filter_state() ` that filters the exclude_varnames, but it does not assert the condition. This check is still necessary because if a user defines an incorrect teal_slice varname to filter out those teal_slices.

For example, if exclude_varname  and varname is same . it will fail assertion
```
teal_slices(
  teal_slice(dataname = "iris", varname = "Sepal.Length")
  exclude_varnames = list(iris = "Sepal.Length")
)

teal_slices(
  teal_slice(dataname = "mae", varname = "age"),
  teal_slice(dataname = "mae", experiment = "experiment", varname = "gene.id"),
  exclude_varnames = list(
    mae = list(subjects = "age", experiment = "gene.id")
  )
)
```

In another scenario, consider an example where teal_slice contains a varname that does not exist in the data. Specifically, the ARRAY_TYPE varname is being filtered out from teal_slice due to its absence in the data.
```

datasets <- teal.slice:::FilteredData$new(
  list(iris = list(dataset = iris),
       mae = list(dataset = miniACC)
  )
)
fs <-
  teal_slices(
    teal_slice(dataname = "iris", varname = "Sepal.Length", selected = c(5.1, 6.4),
               keep_na = TRUE, keep_inf = FALSE),
    teal_slice(dataname = "iris", varname = "Species", selected = c("setosa", "versicolor"),
               keep_na = FALSE),
    teal_slice(dataname = "mae", varname = "years_to_birth", selected = c(30, 50),
               keep_na = TRUE, keep_inf = FALSE),
    teal_slice(dataname = "mae", varname = "vital_status", selected = "1", keep_na = FALSE),
    teal_slice(dataname = "mae", varname = "gender", selected = "female", keep_na = TRUE),
    teal_slice(dataname = "mae", varname = "ARRAY_TYPE",
               selected = "", keep_na = TRUE, datalabel = "RPPAArray", arg = "subset")
  )
datasets$set_filter_state(state = fs)
```

### DIFF
--- a/R/teal_slices.R
+++ b/R/teal_slices.R
@@ -66,7 +66,7 @@
 #'   filter_2,
 #'   filter_3,
 #'   exclude_varnames = list(
-#'     "dataname1" = "varname2"
+#'     "dataname1" = "varname3"
 #'   )
 #' )
 #'
@@ -87,6 +87,12 @@ teal_slices <- function(...,
                         allow_add = TRUE) {
   slices <- list(...)
   checkmate::assert_list(slices, types = "teal_slice", any.missing = FALSE)
+  checkmate::assert_list(exclude_varnames, names = "named", types = "character", null.ok = TRUE, min.len = 1)
+  checkmate::assert_list(include_varnames, names = "named", types = "character", null.ok = TRUE, min.len = 1)
+  checkmate::assert_character(count_type, len = 1, null.ok = TRUE)
+  checkmate::assert_subset(count_type, choices = c("all", "none"), empty.ok = TRUE)
+  checkmate::assert_logical(allow_add)
+
   slices_id <- shiny::isolate(vapply(slices, `[[`, character(1L), "id"))
   if (any(duplicated(slices_id))) {
     stop(
@@ -94,11 +100,24 @@ teal_slices <- function(...,
       toString(unique(slices_id[duplicated(slices_id)]))
     )
   }
-  checkmate::assert_list(exclude_varnames, names = "named", types = "character", null.ok = TRUE, min.len = 1)
-  checkmate::assert_list(include_varnames, names = "named", types = "character", null.ok = TRUE, min.len = 1)
-  checkmate::assert_character(count_type, len = 1, null.ok = TRUE)
-  checkmate::assert_subset(count_type, choices = c("all", "none"), empty.ok = TRUE)
-  checkmate::assert_logical(allow_add)
+
+  datanames <- vapply(slices, `[[`, character(1L), "dataname")
+  varnames <- lapply(structure(datanames, names = datanames), function(x) {
+    unique(unlist(lapply(Filter(function(xx) xx$dataname == x, slices), `[[`, "varname")))
+  })
+
+  lapply(intersect(names(varnames), names(include_varnames)), function(n) {
+    if (!all(varnames[[n]] %in% include_varnames[[n]])) {
+      varnames_not_allowed <- setdiff(varnames[[n]], include_varnames[[n]])
+      stop(sprintf("filters for %s not allowed in %s", toString(varnames_not_allowed), n), call. = FALSE)
+    }
+  })
+  lapply(intersect(names(varnames), names(exclude_varnames)), function(n) {
+    if (any(varnames[[n]] %in% exclude_varnames[[n]])) {
+      varnames_forbidden <- intersect(varnames[[n]], exclude_varnames[[n]])
+    }
+      stop(sprintf("filters for %s forbidden in %s", toString(varnames_forbidden), n), call. = FALSE)
+  })
 
   structure(
     slices,
--- a/R/teal_slices.R
+++ b/R/teal_slices.R
@@ -116,7 +116,7 @@ teal_slices <- function(...,
     if (any(varnames[[n]] %in% exclude_varnames[[n]])) {
       varnames_forbidden <- intersect(varnames[[n]], exclude_varnames[[n]])
     }
-      stop(sprintf("filters for %s forbidden in %s", toString(varnames_forbidden), n), call. = FALSE)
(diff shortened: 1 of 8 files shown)


R/teal_slices.R
chlebowa, line 1: You are manipulating values before assertions that ensure their proper type.

R/teal_slices.R
```diff
@@ -94,6 +94,17 @@ teal_slices <- function(...,
       toString(unique(slices_id[duplicated(slices_id)]))
     )
   }
+
+  varnames <- shiny::isolate(stats::setNames(
+    lapply(slices, function(slice) slice$varname),
+    sapply(slices, function(slice) slice$dataname)
+  ))
+  excluded_datanames <- unlist(intersect(names(varnames), names(exclude_varnames)))
+  if (!is.null(excluded_datanames) && length(excluded_datanames) > 0) {
```
chlebowa, line 103: redundancy

chlebowa (changes requested):
Logic is shady but I will take a closer look when redundant conditions are removed.
Checks on `include_varnames` have not been implemented.

kartikeyakirar:
Issue with Logic and Updated Changes:

The previous issue was that the logic was causing the creation of a duplicate name list. For example:
```
$mtcars
NULL
$mtcars
"cyl"
```
To address this problem, changes were made to iterate over the dataname from each slice element and extract the varname, while ignoring null values. An additional check for include_varname was introduced. This new check enforces that the filter for dataname should have varname from include_varname. However, this change resulted in some test failures for examples used in the package.

One specific test failure occurred when calling get_filter_state(datasets) on the dataset. The assertion failed with the following application code:
```

  x = list(
    iris = list(dataset = iris),
    mae = list(dataset = miniACC)
  )
)
fs <- teal_slices(
  teal_slice(dataname = "iris", varname = "Species", selected = c("setosa", "versicolor")),
  teal_slice(dataname = "iris", varname = "Sepal.Length", selected = c(5.1, 6.4)),
  teal_slice(
    dataname = "mae", varname = "years_to_birth", selected = c(30, 50),
    keep_na = TRUE, keep_inf = FALSE
  ),
  teal_slice(
    dataname = "mae", varname = "vital_status", selected = "1",
    keep_na = FALSE
  ),
  teal_slice(
    dataname = "mae", varname = "gender", selected = "female",
    keep_na = TRUE
  ),
  teal_slice(
    dataname = "mae", varname = "ARRAY_TYPE", selected = "",
    keep_na = TRUE, experiment = "RPPAArray", arg = "subset"
  )
)

# set initial filter state
set_filter_state(datasets, filter = fs)

# get filter state
get_filter_state(datasets)
```
The error message received was:
```
Error in FUN(X[[i]], ...) :
Assertion on 'varnames[[name]]' failed: Must be a subset of {'patientID','years_to_birth','vital_status','days_to_death','days_to_last_followup','tumor_tissue_site','pathologic_stage','pathology_T_stage','pathology_N_stage','gender','date_of_initial_pathologic_diagnosis','radiation_therapy','histological_type','residual_tumor','number_of_lymph_nodes','race','ethnicity','Histology','C1A.C1B','mRNA_K4','MethyLevel','miRNA.cluster','SCNA.cluster','protein.cluster','COC','OncoSign','purity','ploidy','genome_doublings','ADS'}, but has additional elements {'ARRAY_TYPE'}.

```
Upon investigation, it was found that the "ARRAY_TYPE" variable does not exist in the miniACC dataset. Previously, before assertion, this would get filtered out at set_filter_state(), but currently, it fails assertion if teal_slice does not contain the varname defined in include_varname for a particular dataname.

The approach taken to remove the teal_slice containing the "ARRAY_TYPE" varname from the example to pass the tests seems appropriate. However, it's worth noting that the example containing "ARRAY_TYPE" might have served a specific purpose, and without sufficient understanding of miniACC, it's difficult to determine if its removal is entirely appropriate.

I wanted to confirm if this is correct approach for assertion or what should be expected behavior in above scenarios?

chlebowa:
> Upon investigation, it was found that the "ARRAY_TYPE" variable does not exist in the miniACC dataset. Previously, before assertion, this would get filtered out at set_filter_state(), but currently, it fails assertion if teal_slice does not contain the varname defined in include_varname for a particular dataname.
> 
> The approach taken to remove the teal_slice containing the "ARRAY_TYPE" varname from the example to pass the tests seems appropriate. However, it's worth noting that the example containing "ARRAY_TYPE" might have served a specific purpose, and without sufficient understanding of miniACC, it's difficult to determine if its removal is entirely appropriate.
> 
> I wanted to confirm if this is correct approach for assertion or what should be expected behavior in above scenarios?

Wrong.
```
library(MultiAssayExperiment)
data(miniACC)
ex <- miniACC[["RPPAArray"]]
names(rowData(ex))
```

kartikeyakirar:
Oh, I was accessing it incorrectly. working on logic again,

kartikeyakirar:
"ARRAY_TYPE" from miniACC data is filtered out and which also causing the test to fail . logging bug https://github.com/insightsengineering/teal.slice/issues/399

chlebowa:
Blocked by #400.

kartikeyakirar:
As per discussion with @gogonzo,
Both` include_varnames` and` exclude_varnames `are currently considered "experimental." We should defer the decision on how to effectively communicate the variables to be included or excluded until later. Consequently, the asserts in `teal_slices` should not be a significant concern at this time.

PR #398, opened 2023-07-21
382 rename internal functions

### LINKED ISSUE #382: consistent naming of internal functions in the filter panel API
By @gogonzo [here](https://github.com/insightsengineering/teal/pull/859)

> Now we have store, restore, as.list, slices_to_list, as.list, reassemble, disassemble. We need to figure out consistent way of converting slices to list and oposite.

There are a number of internal functions that deal with `teal_slice` and `teal_slices` objects and their names have become somewhat inconsistent. Discuss and revise the naming convention. Note the differences between to and from list conversions as there are several.

Remember, some of these functions are duplicated in `teal`.

chlebowa:
Fixes #382 

+ `disassemble_slices` function is registered as `as.list` method for `teal_slices`
+ `as.list_teal_slices` has a `recursive` argument to control whether constituent `teal_slice` are also converted to lists
+ `reassemble_slices` function is renamed to `as.teal_slices`
+ `as.teal_slices` builds `teal_slices` from lists, which may but need not be `teal_slice`
+ `as.teal_slices` function is renamed to `list_to_teal_slices`
+ `slices_to_list` function is removed; `format.teal_slices` uses `as.list.teal_slices` now


R/teal_slices.R
```diff
@@ -267,13 +215,18 @@ format.teal_slices <- function(x, show_all = FALSE, trim_lines = TRUE, ...) {
   checkmate::assert_flag(show_all)
   checkmate::assert_flag(trim_lines)
 
-  slices_list <- slices_to_list(x)
+  x <- as.list(x, recursive = TRUE)
+  attrs <- attributes(x)
+  attributes(x) <- NULL
+  slices_list <- list(slices = x, attributes = attrs)
+  slices_list <- Filter(Negate(is.null), slices_list) # drop attributes if empty
```
m7pr, line 222: I am having seconds thoughts if we should keep empty attributes (as it happens now) and I reckon they can be removed

m7pr:
```suggestion

```
but this will never empty as `teal_slices` has default non-NULL-able parameter `allow_add = TRUE` hence this always fill be in attributes

m7pr (approved):
This is all great. Happy to see the cleanup and unification of functions. **Even though I approve, I left one comment**. Tested functionality with the below code

```{R}
filter_1 <- 
  teal_slice(
   dataname = "dataname1",
   varname = "varname1",
   choices = letters[1:2],
   selected = "b",
   keep_na = TRUE,
   fixed = FALSE,
   extra1 = "extraone"
  )

filter_2 <- 
  teal_slice(
   dataname = "dataname1 asd asd as dasd as das da sdasd ",
   varname = "varname2",
   choices = 1:2,
   keep_na = TRUE,
   selected = 2,
   fixed = TRUE,
   locked = FALSE,
   extra2 = "extratwo"
 )
filter_3 <- 
  teal_slice(
   dataname = "data",
   id = "FA",
   title = "Female adults",
   expr = "SEX == 'F' & AGE >= 18"
  )
 
all_filters <- 
  teal_slices(
   filter_1,
   filter_2,
   filter_3,
   exclude_varnames = list("dataname1" = "varname2")
  )

all_filters_default <- 
  teal_slices(
    filter_1,
    filter_2,
    filter_3
  )
  
format(filter_1)
print(filter_1)

format(all_filters)
print(all_filters)

format(all_filters_default, trim_lines = FALSE)
print(all_filters_default, show_all = TRUE, trim_lines = FALSE)

slices_store(all_filters_default, file = 'teal_slice_7.json')
y <- slices_restore('teal_slice_7.json')


teal.slice:::as.list.teal_slices(y)
teal.slice:::as.list.teal_slices(y, recursive = TRUE)
````

File: R/teal_slices.R (hunk 1 above)
Comment (chlebowa) [reply R/teal_slices.R:222]: I thought about it. While this is true now, perhaps we will allow having no attributes in the future. Some attributes are only listed if they are not defaults, e.g. `count_type`.

Comment (m7pr) [reply R/teal_slices.R:222]: Yeah, but in the future if you remove `all_add` parameter from `teal_slices` it will have an extra implication of attributes removal from `print`, `format` and `store`. No other argument will have this implication. And you need to remember about it. I think it's gonna be less painful in the future to bring back this `Filter()` if we allow all attributes to be NULL, then figure out why default call for `print` for `teal_slices` dropped attributes. This is minor and I defer to you to make a final decision

Comment (chlebowa) [reply R/teal_slices.R:222]: Hold on. This is not about not printing NULL attributes, NULL attributes are never printed explicitly. This `Filter` is only here because we explicitly extract attributes and that can **add** a NULL to the list that is passed to `do.call`.

**Review by gogonzo — APPROVED** [review]
👍 nice solution with the `recursive`

File: R/teal_slices.R (hunk 1 above)
Comment (chlebowa) [reply R/teal_slices.R:222]:
```
slices <- as.list(letters[1:3])
attrs <- NULL
list(slices = slices, attrs = attrs)
```
```
$slices
$slices[[1]]
[1] "a"

$slices[[2]]
[1] "b"

$slices[[3]]
[1] "c"


$attrs
NULL
```

Better safe than sorry :slightly_smiling_face:

Issue #375, opened 2023-07-05
`exclude_varnames` leaks


chlebowa:
Filters can still be added for excluded variables (in this example, `Petal.Length`).
Oddly, only affects some modules in the app.

https://github.com/insightsengineering/teal.slice/assets/114988527/9025294e-ba92-478d-96bb-5c1891382419

### FILTERS
```
default_filters <- teal::teal_filters(
  filter_var("iris", "Sepal.Length"),
  filter_var("iris", "Sepal.Width"),
  filter_var("iris", "Species"),
  filter_var("mtcars", "mpg"),
  exclude_varnames = list(
    iris = c("Petal.Length"),
    mtcars = c("qsec", "drat")
  ),
  mapping = list(
    funny = c("iris Sepal.Length"),
    funny2 = c("iris Sepal.Width"),
    global_filters = "mtcars mpg"
  )
)
```
### MODULE
```
funny_module <- function (label = "Filter states", datanames = "all") {
  checkmate::assert_string(label)
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderPrint({
        logger::log_trace("rendering text1")
        filter_panel_api %>% get_filter_state()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}
```
### APP
```
app <- init(
  data = teal_data(
    dataset("iris", iris),
    dataset("mtcars", mtcars)
  ),
  modules = modules(
    tm_data_table(
      "table",
      variables_selected = list(),
      dt_args = list()
    ),
    modules(
      label = "tab1",
      funny_module("funny", datanames = NULL),
      funny_module("funny2", datanames = "iris")
    )
  ),
  filter = default_filters
)


runApp(app, launch.browser = TRUE)
```

`teal` on `filter_panel_refactor@main`
`teal.slice` on `filter_panel_refactor@main`

chlebowa:
[...]

PR #401, opened 2023-07-24
prevent to specify the same dataset in both, include_varnames and exc…

gogonzo:
During exploring #375 I figured out that one assert is missing - preventing to specify include_varnames and exclude_varnames for the same data set.

### DIFF
--- a/R/teal_slices.R
+++ b/R/teal_slices.R
@@ -10,7 +10,8 @@
 #'
 #' `include_varnames` and `exclude_varnames` determine which variables can have filters assigned.
 #' The former enumerates allowed variables, the latter enumerates forbidden values.
-#' Since these can be mutually exclusive in some cases, they cannot both be set in one `teal_slices` object.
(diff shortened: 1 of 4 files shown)


R/teal_slices.R
```diff
@@ -10,7 +10,8 @@
 #'
 #' `include_varnames` and `exclude_varnames` determine which variables can have filters assigned.
 #' The former enumerates allowed variables, the latter enumerates forbidden values.
-#' Since these can be mutually exclusive in some cases, they cannot both be set in one `teal_slices` object.
+#' Since these can be mutually exclusive, they cannot both be set in one `teal_slices` object for one
```
chlebowa, line 13:
```suggestion
#' Since these could be mutually exclusive, it is impossible to set both allowed and forbidden variables for one data set in one `teal_slices`.
```

chlebowa (approved):
:+1:

PR #402, opened 2023-07-24
#227 make subsetting calls prettier

### LINKED ISSUE #217: Filter panel switch
Change the switch for disabling the filter panel to match that of the individual cards. Also move it to the Active Filter Variables section.

See https://github.com/insightsengineering/teal.slice/pull/207#issuecomment-1447262012 for some background.

### LINKED ISSUE #227: make subsetting calls prettier
Here is an example of a `FilterState` that does not select any values.

```
filter_state <- init_filter_state(
  x = c(letters, NA_character_),
  slice = teal_slice(
    varname = "x",
    dataname = "dataname"
  )
)

filter_state$get_call() %>% isolate
filter_state$set_state(
  teal_slice(
    varname = "x",
    dataname = "dataname",
    keep_na = TRUE
  )
)

filter_state$get_call() %>% isolate
filter_state$set_state(
  teal_slice(
    varname = "x",
    dataname = "dataname",
    keep_na = FALSE
  )
)
filter_state$get_call() %>% isolate
# !is.na(x) & x %in% c("a", "b", "c", "d", "e", "f", "g", "h", 
#    "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t",
#    "u", "v", "w", "x", "y", "z"
```

The last call should return just `!is.na(x)` as it's enough to filter data according to the teal_slice configuration.
`get_call` method should not enumerate values if they cover all possible choices. `is_any_filtered` shouldn't decide about returning the whole (combination of enumeration and is.na) call but it should be more flexible.

m7pr:
Closes #227 and motivates #407

This pull request introduces 

- changes to the output of `get_call()`
    - if all choices are selected and NAs exist in the data and keep_na=FALSE the choices are not a part of the filter (as it was dummy to enumerate all selected variables for all possible choices
    - if there are no NAs in the data then `is.na()` nor `!is.na()` are not returned in `get_call()` no matter whether `keep_na` is set to `FALSE` or `TRUE` - tests were adjusted accordingly
    - if selected is empty (`selected = character()`), `get_call` returns `FALSE` (besides the case when NAs exist in the data and `keep_na=TRUE`

- changes in methods
    - `FilterState$set_na_rm()` method was removed as it was barely used
    - edited `FilterState$add_keep_na_call` so it takes `varname` as a second parameter, and not `dataname` to make the code less repeatable and more readable

## Example

Below does not list all possible choices for `ChoicesFilterState$get_call()` if all choices are all data levels. 

Final `$get_call()` returns `!is.na(x)` in below example.
Previously it would return 
`!is.na(x) & x %in% c("a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l".... "t", "u", "v", "w", "x", "y", "z")` 
where `x %in% c("a", ...` part is redundant.

### Code

```{R}
library(dplyr)
filter_state <- teal.slice:::init_filter_state(
  x = c(letters, NA_character_),
  slice = teal_slice(
    varname = "x",
    dataname = "dataname",
    keep_na = FALSE
  )
)

filter_state$get_call() %>% isolate

filter_state <- teal.slice:::init_filter_state(
  x = c(letters, NA_character_),
  slice = teal_slice(
    varname = "x",
    dataname = "dataname",
    keep_na = TRUE
  )
)

filter_state$get_call() %>% isolate

filter_state <- teal.slice:::init_filter_state(
  x = c(letters, NA_character_),
  slice = teal_slice(
    varname = "x",
    dataname = "dataname",
    keep_na = TRUE,
    selected = letters[1:8]
  )
)

filter_state$get_call() %>% isolate

filter_state <- teal.slice:::init_filter_state(
  x = c(letters, NA_character_),
  slice = teal_slice(
    varname = "x",
    dataname = "dataname",
    keep_na = FALSE,
    selected = letters[1:8]
  )
)

filter_state$get_call() %>% isolate

```

### Execution

```{R}
> library(dplyr)
> filter_state <- teal.slice:::init_filter_state(
+   x = c(letters, NA_character_),
+   slice = teal_slice(
+     varname = "x",
+     dataname = "dataname",
+     keep_na = FALSE
+   )
+ )
> filter_state$get_call() %>% isolate
!is.na(x)
> filter_state <- teal.slice:::init_filter_state(
+   x = c(letters, NA_character_),
+   slice = teal_slice(
+     varname = "x",
+     dataname = "dataname",
+     keep_na = TRUE
+   )
+ )
> filter_state$get_call() %>% isolate
NULL
> filter_state <- teal.slice:::init_filter_state(
+   x = c(letters, NA_character_),
+   slice = teal_slice(
+     varname = "x",
+     dataname = "dataname",
+     keep_na = TRUE,
+     selected = letters[1:8]
+   )
+ )
> filter_state$get_call() %>% isolate
is.na(x) | x %in% c("a", "b", "c", "d", "e", "f", "g", "h")
> filter_state <- teal.slice:::init_filter_state(
+   x = c(letters, NA_character_),
+   slice = teal_slice(
+     varname = "x",
+     dataname = "dataname",
+     keep_na = FALSE,
+     selected = letters[1:8]
+   )
+ )
> filter_state$get_call() %>% isolate
x %in% c("a", "b", "c", "d", "e", "f", "g", "h")
```


R/FilterState.R
```diff
@@ -602,14 +602,18 @@ FilterState <- R6::R6Class( # nolint
     # @param dataname `character(1)` name of data set to prepend to variables
     # @return a `call`
     add_keep_na_call = function(filter_call, dataname) {
-      if (isTRUE(private$get_keep_na())) {
+      if (isTRUE(private$get_keep_na()) & length(filter_call) > 0) {
```
chlebowa, line 605: I prefer `!is.null(filter_call)` over `length(filter_call) > 0`.

R/FilterStateChoices.R
```diff
@@ -223,7 +223,11 @@ ChoicesFilterState <- R6::R6Class( # nolint
           )
         } else {
           # This handles numerics, characters, and factors.
-          call(fun_compare, varname, make_c_call(choices))
+          if (!all(na.omit(private$x) %in% choices)) {
```
chlebowa, line 226:
```suggestion
          if (!setequal(na.omit(private$x), choices)) {
```

chlebowa (commented):
The logic checks out but it is very hard to follow.

gogonzo:
Table of expected calls

|                           | all selected | some selected/ limited choices   | single selected      | none selected |
|---------------------------|--------------|----------------------------------|----------------------|---------------|
| No NA's in data           | `NULL`         | `x %in% c("a", "b")`               |  `x == "a"`            | `FALSE`         |
| No NA's in data, keep_na  | `NULL`         | `x %in% c("a", "b")`               |  `x == "a"`            | `FALSE`         |
| No NA's in data, !keep_na | `NULL`         |  `x %in% c("a", "b")`  | `x == "a"`             | `FALSE`         |
| keep_na                   | `NULL`         | `is.na(x) or  x %in% c("a", "b")`  | `is.na(x) or x == "a"` | `is.na(x)`      |
| !keep_na                  | `!is.na(x)`    | `x %in% c("a", "b")`               | `!is.na(x) & x == "a"`             | `FALSE`         |

There is a problem with limited choices - when choices are limited, selection should always be returned.
I have a slight problem with none selected - when nothing is selected then I'd expect FALSE as `x %in% character(0)` is bit weird.

Please make tests covering all the scenarios. Please make sure it works also with numerics and dates

chlebowa:
Oh, right, I forgot to check "none selected".
I am not sure about `FALSE`. Perhaps then it would be better to say `!x %in% <choices>`?

chlebowa:
Indeed, I get `is.na(x)` even with no `NA`s in the data.

R/FilterStateChoices.R
chlebowa, line 226: Perhaps this check should go in `FilterState$is_any_filtered`?

m7pr: no because it would return `FALSE` for `$is_any_filtered()`, hence `$get_call()` method would stop sooner 
```{R}
      if (isFALSE(private$is_any_filtered())) {
        return(NULL)
      }
````

and it would not append `$add_keep_na_call()` method at the end

File: R/FilterState.R (hunk 1 above)
Comment (m7pr) [reply R/FilterState.R]: 3df261712f9e191ffbea8651259c6c1d0fbde195

**Comment by m7pr** [comment]
@gogonzo I added ~70 candidates for test scenarios to tests/tes-ChoicesFilterState.R https://github.com/insightsengineering/teal.slice/pull/402/commits/36aba2ee16658b627d87d4717ac2f9d4fa71bdc8
I will create tests and go one by one to see if we cover all scenarios properly.

For `length(selected) == 0` (none selected) case I would suggest `$get_call()` returns 
- `x[0]`  for all situations besides the below
- `is.na(x)` only if keep_na = TRUE and NAs in the data


Lastly @gogonzo @chlebowa where do you see/use this `na_rm` ?

**Comment by chlebowa** [comment]
> Lastly @gogonzo @chlebowa where do you see/use this `na_rm` ?

I honestly have no idea.

File: R/FilterState.R [hunk 3]
```diff
@@ -602,14 +602,18 @@ FilterState <- R6::R6Class( # nolint
...
         call("|", call("is.na", private$get_varname_prefixed(dataname)), filter_call)
-      } else if (isTRUE(private$na_rm) && private$na_count > 0L) {
+      } else if (isFALSE(private$get_keep_na()) & length(filter_call) == 0) {
+        call("!", call("is.na", private$get_varname_prefixed(dataname)))
```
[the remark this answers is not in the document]
m7pr: 391999ece7d6cd37b3e0f947558edfa806e33895

R/FilterStateChoices.R
m7pr: moved in here https://github.com/insightsengineering/teal.slice/pull/402/commits/2e4990ada5fab287ee7935fa3ac96848eefd13ac

m7pr:
`na_rm` was removed with 5f6ddf9a16d73774c7879e59437b621ae88ef8e8

m7pr:
Hey @gogonzo for this proposition cell I reckon it's ok just to return `x == 'a'` as this also drops NAs
image

```{R}
iris2 <- iris
iris2[1, 5] <- NA
library(dplyr)

# NA exists, but keep_na is FALSE
# no need for !is.na(Species)
iris2 %>% filter(Species == 'setosa' & !is.na(Species)) %>% nrow
iris2 %>% filter(Species == 'setosa') %>% nrow

```

☝️  both gives 49 rows

m7pr:
Hey @gogonzo and @chlebowa the PR is ready for your review. I added 24 tests in tests/testthat/test-ChoicesFilterState.R so that I was able to run all possible cases multiple times, as I was doing constant changes all over again. I'm not sure this is totally needed to be included in the package, but it doesn't hurt to have to checks. I would also like to do the same for `numeric`, `dates` and `POSIXct` classes, so that GHA workflows shows that all cases are working as intended and were tested.

chlebowa:
> Hey @gogonzo for this proposition cell I reckon it's ok just to return `x == 'a'` as this also drops NAs image

I think, `!is.na(x)` should be explicit.

m7pr:
@chlebowa but that's redundant. `x == 'a'` disqualifies NAs

gogonzo:
> @chlebowa but that's redundant. x == 'a' disqualifies NAs. we do not put !is.na(x) anywhere else

@m7pr it is not. `filter` and `subset` function discards NA but there are some cases when NAs are not handled. 
Simple logical condition returns NA
```r
c(1:3, NA) == 1
# TRUE  FALSE  FALSE  FALSE  NA
```

In case of `MultiAssayExperiment` it even throws an error

```r
library(MultiAssayExperiment)
subsetByColData(miniACC, miniACC$protein.cluster == 1)
# Error: logical subscript contains NAs
```

We used to use `FilterState$set_na_rm` for `MultiAssayExperiment` filters, so it used to add `!is.na(x)` to every call where variable contained `NA`. So it means we have to include it in all calls. (seems na_rm isn't used anymore and should be removed)

m7pr:
Thanks @gogonzo now it make sense. I wasn't aware of this - I was only testing for filter/subset. I'll try to get this sorted out, maybe this can be figured out based on the class of `FilterState/s`, or I can bring back is.na()

chlebowa:
> We used to use `FilterState$set_na_rm` for `MultiAssayExperiment` filters, so it used to add `!is.na(x)` to every call where variable contained `NA`. So it means we have to include it in all calls. (seems na_rm isn't used anymore and should be removed)

THAT's what it was for?? :roll_eyes:

m7pr:
Hey @gogonzo we were able to update the code, together with @chlebowa so it returns possible scenarios the same way as you explained in the table, with **one difference** - we also return `!is.na(x) &` in the case marked in blue. This made the implementation code shorter, and the output code more direct of how we handle NAs, even though this doesn't make change to `%in%` operator that handles HAs.

image

chlebowa:
> with **one difference** - we also return `!is.na(x) &` in the case marked in blue.

I approve of this. If data contains missing values, the call always handles them explicitly. (Unless none selected, in which case `FALSE`.)

gogonzo (approved):
Checked against several apps. Analysed each test result and it looks good 👍

PR #408, opened 2023-07-30
867 modify mapping matrix

chlebowa:
Related to [this issue](https://github.com/insightsengineering/teal/issues/867).

`FilteredData` receives a new public method, `$get_available_teal_slices` that returns the contents of the private field `private$available_teal_slices`.

### DIFF
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -152,8 +152,9 @@ FilteredData <- R6::R6Class( # nolint
       checkmate::assert_class(x, "reactive")
       private$available_teal_slices <- reactive({
         # Available filters should be limited to the ones relevant for this FilteredData.
-        allowed <- attr(self$get_filter_state(), "include_varnames")
-        forbidden <- attr(self$get_filter_state(), "exclude_varnames")
+        current_state <- isolate(self$get_filter_state())
+        allowed <- attr(current_state, "include_varnames")
+        forbidden <- attr(current_state, "exclude_varnames")
         foo <- function(slice) {
           if (slice$dataname %in% self$datanames()) {
             if (slice$fixed) {
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -155,6 +155,16 @@ FilteredData <- R6::R6Class( # nolint
       invisible(NULL)
     },
 
+    #' Get list of filter states available for this object.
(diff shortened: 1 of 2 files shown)


R/FilteredData.R
```diff
@@ -155,6 +155,16 @@ FilteredData <- R6::R6Class( # nolint
       invisible(NULL)
     },
 
+    #' Get list of filter states available for this object.
```
gogonzo, line 158: https://github.com/insightsengineering/teal/blob/867_mapping_matrix_NA%40main/R/module_filter_manager.R#L208
`set_available_teal_slices` filters only available datanames - but should it also filter by `attr(self$get_filter_state(), "include_varnames")`?

chlebowa: Good question. 
`FilterStates` will not allow such a state to be created from the UI but what if a `teal_slices` is passed that refers to forbidden variables? This is why #340 would be nice to have.

Let me run some tests. I will modify `$set_available_teal_slices` if necessary.

gogonzo: Yup, #340 is nice to have but MAE make `teal_slice` objects hard to match with include/exclude_varnames without introducing MAE-specific-code (against open-close principle).

chlebowa: `teal_slice_expr` will be included if `dataname` checks out, even if they refer to variables not allowed.

R/FilteredData.R
```diff
@@ -150,11 +152,35 @@ FilteredData <- R6::R6Class( # nolint
       checkmate::assert_class(x, "reactive")
       private$available_teal_slices <- reactive({
         # Available filters should be limited to the ones relevant for this FilteredData.
-        Filter(function(x) x$dataname %in% self$datanames(), x())
+        allowed <- attr(self$get_filter_state(), "include_varnames")
+        forbidden <- attr(self$get_filter_state(), "exclude_varnames")
```
gogonzo, line 156: We want this reactive to be dependent on `x` (slices_global)

```suggestion
        allowed <- attr(isolate(self$get_filter_state()), "include_varnames")
        forbidden <- attr(isolate(self$get_filter_state()), "exclude_varnames")
```

gogonzo (approved):
👍

PR #423, opened 2023-08-03
410 note experimental@main

### LINKED ISSUE #410: Include information about experimental feature to arguments
- count_type (tea.slice)
- module_specific and mapping (teal)

Doesn't mean that their fate is unsure, we just need to know users opinion which might result in some improvements.

chlebowa:
Closes #410


gogonzo (approved):
👍

Issue #424, opened 2023-08-04
[Bug]: The `Active Filter Variables` is not bold


vedhav:
### What happened?

In the filter panel the label `Active Filter Variables` is not bold as the rest of them.

<img width="1470" alt="Screenshot 2023-08-04 at 8 57 58 PM" src="https://github.com/insightsengineering/teal.slice/assets/49812166/2bd02af1-de95-4965-9894-db20e86e28fc">

### sessionInfo()

```R
R version 4.3.1 (2023-06-16)
Platform: x86_64-apple-darwin20 (64-bit)
Running under: macOS Ventura 13.4.1

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.3-x86_64/Resources/lib/libRblas.0.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.3-x86_64/Resources/lib/libRlapack.dylib;  LAPACK version 3.11.0

locale:
[1] C/UTF-8/C/C/C/C

time zone: Asia/Kolkata
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] teal.slice_0.3.0.9028             testthat_3.1.9                   
 [3] sparkline_2.0                     teal.modules.general_0.2.15.9033 
 [5] shinyTree_0.2.7                   ggmosaic_0.3.3                   
 [7] ggplot2_3.4.2                     teal.modules.clinical_0.8.15.9013
 [9] tern_0.8.5.9004                   rtables_0.6.2.9001               
[11] formatters_0.5.1.9001             teal.goshawk_0.1.14.9020         
[13] teal_0.13.0.9025                  teal.transform_0.3.0.9009        
[15] teal.data_0.2.0.9007              goshawk_0.1.14.9016              
[17] stringr_1.5.0                     scda.2022_0.1.5.9000             
[19] scda_0.1.6.9010                   dplyr_1.1.2                      
[21] magrittr_2.0.3                    DescTools_0.99.49                
[23] shiny_1.7.4.1                    

loaded via a namespace (and not attached):
  [1] rstudioapi_0.14          jsonlite_1.8.7           estimability_1.4.1      
  [4] farver_2.1.1             rmarkdown_2.23           geepack_1.3.9           
  [7] fs_1.6.3                 vctrs_0.6.3              memoise_2.0.1           
 [10] usethis_2.2.1            htmltools_0.5.5          forcats_1.0.0           
[...]

PR #426, opened 2023-08-07
Bold the primary text in the filter panels to stay consistent 

vedhav:
Closes https://github.com/insightsengineering/teal.slice/issues/424

### DIFF
--- a/inst/css/filter-panel.css
+++ b/inst/css/filter-panel.css
@@ -64,6 +64,10 @@
   height: 3rem;
 }
 
+.text-primary.mb-4 {
+  font-weight: 700;
+}
+
 .filter_datelike_input input {
   height: 3rem !important;
   text-align: center !important;


kartikeyakirar:
@vedhav, I'm uncertain about altering .text-primary.mb-4, as it's utilized extensively throughout the application and could potentially have unintended consequences elsewhere. Could we consider using the <strong> tag instead? whats do you think ?

chlebowa:
The same mechanism is used for styling the other panel headers, I see no issue with this.

inst/css/filter-panel.css
```diff
@@ -64,6 +64,10 @@
   height: 3rem;
 }
 
+.text-primary.mb-4 {
```
vedhav, line 67:
```suggestion
#teal-main_ui-filter_panel .text-primary.mb-4
```

How about explicitly targeting the filter panel?

kartikeyakirar: Let's avoid introducing an extra class just for this. I concur with @chlebowa; I haven't identified any unforeseen repercussions elsewhere within this package.

vedhav: So, if you're okay with the change, can you approve it?

kartikeyakirar:
> The same mechanism is used for styling the other panel headers, I see no issue with this.
In such a scenario, the font-weight should match. We should investigate precisely what is causing this change in the CSS.

kartikeyakirar:
@vedhav can you try by reverting the css changes and replace `tag$span` with `tag$label`
like this 
`tags$label("Active Filter Variables", class = "text-primary mb-4")` in filterData.R

inst/css/filter-panel.css
kartikeyakirar: while looking at other places where same class is used I realised all the titles are under label tag. and  for title 'Active Filter Variables' its under span. please refer below comment for suggested changes

vedhav:
> @vedhav can you try by reverting the css changes and replace `tag$span` with `tag$label` like this `tags$label("Active Filter Variables", class = "text-primary mb-4")` in filterData.R

I would still prefer to change the CSS in one place. In case of changing the `span` to `label` we still need to explicitly change `display: flex` to `display: block` for just the "Active Filter Variables" title to correct the alignment. So I'd prefer to make a global CSS change to handle the width like this. WDYT?
Screenshot 2023-08-07 at 2 31 16 PM

inst/css/filter-panel.css
vedhav: There seems to be a reason for doing it differently for this because It is wrapped inside an additional div with the class `filter-panel-active-header` which has a flex property. So it's better not to mess with this.

kartikeyakirar:
I see now. this solution is okay then.

kartikeyakirar (approved):

PR #430, opened 2023-08-08
428 get call bug

### LINKED ISSUE #428: [Bug]: Unexpected error when deselecting values in ChoicesFilterState
### What happened?

image

```
Error in calls_combine_by: Assertion on 'calls' failed: May only contain the following types: {call,name}, but element 1 has type 'logical'.
```

To reproduce run app and add `RACE` and `SEX` filter and deselect all from one and another. Then you can run app with with uncommented `# filter = default_filters` - same error but prevents to run the app.

```
library(shiny)
library(scda)
library(scda.2022)
library(teal.data)
library(teal.transform)
pkgload::load_all("teal")
library(teal.modules.general)

default_filters <- teal_slices(
  teal_slice(dataname = "ADSL", varname = "RACE", selected = character(0)),
  teal_slice(dataname = "ADSL", varname = "SEX", selected = character(0), multiple = FALSE)
)

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL),
    cdisc_dataset("ADRS", ADRS)
  ),
  modules = modules(
    tm_variable_browser()
  ),
  # filter = default_filters
)

runApp(app)

```

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

chlebowa:
Fixes #428 

Having `$get_call` return `FALSE` in some circumstances caused errors downstream. Specifically, `calls_combine_by` demands that all arguments be calls, logical vectors are not allowed.

`$get_call` would return `FALSE` when no values are selected, including NAs, meaning that no observations are to be included. This change was introduced to [prettify filtering calls in PR 402](https://github.com/insightsengineering/teal.slice/pull/402).

In this PR the call for the latter case has been modified to be explicit. For available choices of `c("a", "b", "c")`:
`!var %in% c("a", "b", "c")` with no NAs in data
`is.na(var) | !var %in% c("a", "b", "c")` with NAs present, included
`!is.na(var) & !var %in% c("a", "b", "c")` with NAs present, excluded

Admittedly, this is cumbersome and somewhat goes against the original work in #402. However, I believe this is the way to go because no-selection filter calls will be rare and this unseemly sight will sore users' eyes less than the old all-inclusive filter calls.

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -588,14 +588,8 @@ FilterState <- R6::R6Class( # nolint
         return(filter_call)
       }
 
-      # Deal with empty selection (filter_call == FALSE).
-      if (isFALSE(filter_call) && isTRUE(private$get_keep_na())) {
-        call("is.na", varname)
-      } else if (isFALSE(filter_call) && isFALSE(private$get_keep_na())) {
-        filter_call
-
         # Deal with NAs.
-      } else if (is.null(filter_call) && isFALSE(private$get_keep_na())) {
+      if (is.null(filter_call) && isFALSE(private$get_keep_na())) {
         call("!", call("is.na", varname))
       } else if (!is.null(filter_call) && isTRUE(private$get_keep_na())) {
         call("|", call("is.na", varname), filter_call)
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -588,7 +588,6 @@ FilterState <- R6::R6Class( # nolint
         return(filter_call)
       }
 
-        # Deal with NAs.
       if (is.null(filter_call) && isFALSE(private$get_keep_na())) {
         call("!", call("is.na", varname))
       } else if (!is.null(filter_call) && isTRUE(private$get_keep_na())) {
--- a/R/FilterStateChoices.R
+++ b/R/FilterStateChoices.R
@@ -206,10 +206,11 @@ ChoicesFilterState <- R6::R6Class( # nolint
         if (setequal(na.omit(private$x), selected)) {
           filter_call <- NULL
         } else {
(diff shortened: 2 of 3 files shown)


gogonzo:
WDYT of this?
https://github.com/insightsengineering/teal.slice/pull/431/files

chlebowa:
> WDYT of this?

Not as contained. A problem pops up in one place and is taken care of in a different one. Plus, one would expect a method called `$get_call` to return a `call`.

m7pr:
I reckon @gogonzo solution to be more concise, but I totally agree with @chlebowa that the problem should be fixed in the place where it appears. We can also rename `get_call` to `get_filtering_input` as it's no longer a call when sometimes it's `FALSE`. I also would change `choices` to `selected` as @chlebowa did in this PR, if we got with #431

gogonzo (approved):
> A problem pops up in one place and is taken care of in a different one

This is conditional on whether you see `get_call` returning `FALSE` a problem. But when I think about hit I admit that:
1. deselecting everything is not probable scenario 
2. even if (1) is TRUE, for example keeping only NAs and filtering all non-NA values, it might be confusing seeing `filter(FALSE)` instead.

I like my solution more probably just because it is mine :D

chlebowa:
Well, I do like the brevity of you solution :wink:  

A better phrasing of "problem popping up" would be the reason for the error arises far away from where the error is raised. I think we should return calls where we are supposed to return calls and handle calls where we are supposed to handle calls. Producing non-calls here and handling them there is easy to write but hard to read.

I'm glad we agree on the final points :+1:

Issue #429, opened 2023-08-08
[Bug]: Prevent to deselect any value in DateFilterState


gogonzo:
### What happened?

![image](https://github.com/insightsengineering/teal.slice/assets/6959016/297d1993-d273-4f66-b51a-e9291c599a65)

To reproduce please run the app and select one of date variables, instead of selecting values just hit backspace to delete start or end date - app crashes. 

Acceptance criteria:
- We need to show notification similar to [this](https://github.com/insightsengineering/teal.slice/blob/cf44c48feae4b314cd7feea82c606b08c86e969d/R/FilterStateDate.R#L347) and setting value back to the previous. 

- Please check also `DatetimeFilterState`, after deleting value "text" input stays empty but doesn't fail. Instead of empty value, previous datetime should be displayed

```
library(shiny)
library(scda)
library(scda.2022)
library(teal.data)
library(teal.transform)
pkgload::load_all("teal.slice")
pkgload::load_all("teal")
library(teal.modules.general)

ADSL <- synthetic_cdisc_data("latest")$adsl
ADTTE <- synthetic_cdisc_data("latest")$adtte
ADRS <- synthetic_cdisc_data("latest")$adrs


default_filters <- teal_slices(
  teal_slice(dataname = "ADSL", varname = "RACE", selected = character(0)),
  teal_slice(dataname = "ADSL", varname = "SEX", selected = character(0), multiple = FALSE)
)

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL),
    cdisc_dataset("ADTTE", ADTTE),
    cdisc_dataset("ADRS", ADRS)
  ),
  modules = modules(
    tm_variable_browser()
  ),
  # filter = default_filters
)

runApp(app)

```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

kartikeyakirar:
> Please check also DatetimeFilterState, after deleting value "text" input stays empty but doesn't fail. Instead of empty value, previous datetime should be displayed

[...]

PR #433, opened 2023-08-09
Prevent to deselect any value in DateFilterState 429

kartikeyakirar:
this fixes: https://github.com/insightsengineering/teal.slice/issues/429

I have updated the DateFilterState and incorporated restoration features for previous selections. Each start and end date is now individually assessed for more accurate notifications.

The use of shinyWidgets::airDatepickerInput in DateTimefilterState poses a challenge, as it lacks a return value when cleared. This requires modifications to enable the observation of user actions and the reversion to the previous selection.

Demo

https://github.com/insightsengineering/teal.slice/assets/6700955/17b4e04a-2e8c-4518-a05e-522255d251ec

### DIFF
--- a/R/FilterStateDate.R
+++ b/R/FilterStateDate.R
@@ -355,7 +355,7 @@ DateFilterState <- R6::R6Class( # nolint
               }
 
               if (is.na(end_date)) {
-                end_date <- private$get_selected()[1]
+                end_date <- private$get_selected()[2]
                 updateDateRangeInput(
                   session = session,
                   inputId = "selection",
--- a/R/FilterStateDate.R
+++ b/R/FilterStateDate.R
@@ -343,12 +343,31 @@ DateFilterState <- R6::R6Class( # nolint
               logger::log_trace("DateFilterState$server@2 selection changed, id: { private$get_id() }")
               start_date <- input$selection[1]
               end_date <- input$selection[2]
-              if (start_date > end_date) {
-                showNotification(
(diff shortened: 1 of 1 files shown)


R/FilterStateDate.R
```diff
@@ -343,12 +343,33 @@ DateFilterState <- R6::R6Class( # nolint
...
+              }
+
               if (start_date > end_date) {
-                showNotification(
-                  "Start date must not be greater than the end date. Setting back to default values.",
+                showNotification("Start date must not be greater than the end date. Setting back to default values.",
                   type = "warning"
                 )
               }
+
```
gogonzo, line 372: previous solution `start_date > end_date` was wrong as it just showed notification and sent invalid `start_date` and `end_date` to set selected (without escaping operation). So I suggest to combine it all in one condition to avoid writing updateRangeInput multiple times.

```suggestion
              if (is.na(start_date) || is.na(end_date) || start_date > end_date) {
                updateDateRangeInput(
                  session = session,
                  inputId = "selection",
                  start = private$get_selected()[1],
                  end = private$get_selected()[2]
                )
                showNotification(
                  "Start date must not be greater than the end date. Setting back to previous value.",
                  type = "warning"
                )
                return(NULL)
              }
              
```

kartikeyakirar: @gogonzo! Are the same notifications applicable to all three conditions? I believe they can be comprehended for each of the three scenarios.

gogonzo (approved):

PR #441, opened 2023-08-16
435 filter card header style

### LINKED ISSUE #435: [Bug]: Filterpanel description overlap with rewind state button
### What happened?

A bug happened!

There is a bug within the `filterPanel `widget specifically related to the "`Active filter Variables`"  widget. When a column contains a lengthy description, it causes an undesirable overlap with both the `rewind state` and `restore original state `buttons.

image

Steps to reproduce

-  open [exploratory](https://genentech.shinyapps.io/NEST_exploratory_main/) app
-  Go to DataTable tab
-  In ADSL dataset select TRT02P or SUBJID
-  Select or deselect some option

chlebowa:
Fixes #435 

Filter card elements are rearranged and styling is changed.
Title controls (rewind, restore, remove) will cause varlabel to collapse with an ellipsis.
Byproduct: varname also becomes flexible and the overflow is trimmed

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -283,9 +283,9 @@ FilterState <- R6::R6Class( # nolint
 
       # Filter card consists of header and body, arranged in a single column.
       # Body is hidden and is toggled by clicking on header.
-      # Header consists of title and summary, arranged in a column.
-      # Title consists of conditional icon, varname, conditional varlabel, and controls, arranged in a row.
-      # Summary consists of value and controls, arranged in a row.
+      ## Header consists of title and summary, arranged in a column.
+      ### Title consists of conditional icon, varname, conditional varlabel, and controls, arranged in a row.
+      ### Summary consists of value and controls, arranged in a row.
 
       div(
         id = id,
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -283,9 +283,9 @@ FilterState <- R6::R6Class( # nolint
 
       # Filter card consists of header and body, arranged in a single column.
       # Body is hidden and is toggled by clicking on header.
-        # Header consists of title and summary, arranged in a column.
-          # Title consists of conditional icon, varname, conditional varlabel, and controls, arranged in a row.
-          # Summary consists of value and controls, arranged in a row.
+      # Header consists of title and summary, arranged in a column.
+      # Title consists of conditional icon, varname, conditional varlabel, and controls, arranged in a row.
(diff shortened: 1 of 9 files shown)


R/FilterState.R
```diff
@@ -281,74 +281,80 @@ FilterState <- R6::R6Class( # nolint
     ui = function(id, parent_id = "cards") {
       ns <- NS(id)
 
-      tags$div(
+      # Filter card consists of header and body, arranged in a single column.
+      # Body is hidden and is toggled by clicking on header.
+      ## Header consists of title and summary, arranged in a column.
+      ### Title consists of conditional icon, varname, conditional varlabel, and controls, arranged in a row.
+      ### Summary consists of value and controls, arranged in a row.
```
kartikeyakirar, line 288:
```suggestion
      # Filter card consists of header and body, arranged in a single column.
      # Body is hidden and is toggled by clicking on header.
      # Header consists of title and summary, arranged in a column.
      # Title consists of conditional icon, varname, conditional varlabel, and controls, arranged in a row.
      # Summary consists of value and controls, arranged in a row.
```

kartikeyakirar:
@chlebowa, I noticed that` tags$ `has been removed from` filterState.` Is there a specific reason behind this change? If not, we could maintain consistency throughout the file by reintroducing `tags$<html fun>`.

kartikeyakirar:
> div calls are almost always prefixed with tags$

I see, I had a thought while going through the entire package – when it comes to using HTML tags, there's actually no fundamental difference between tags$<html fun>and<html fun>`. Both are totally fine! But here's an idea:  we can create refactoring task to make it consistent across all the teal packages and make these HTML calls uniform throughout all the teal packages?

kartikeyakirar:
Overall! The PR is looking fantastic and works like a charm! 🚀

chlebowa:
`tags` issue: https://github.com/insightsengineering/teal.slice/issues/445

R/FilterState.R
kartikeyakirar: Could you kindly review this, @chlebowa?

kartikeyakirar (approved):
It looks great. can you please also review above suggested modifications above as well before merging PR.

R/FilterState.R
chlebowa: This is done on purpose to improve readability. I tried to use indention but the bot removed it.

Issue #436, opened 2023-08-10
[Bug]: The RangeFilterState's x-axis tick labels are somewhat obstructed.


kartikeyakirar:
### What happened?

The range selector widget for numeric input partially blocks the x-axis tick labels, which could benefit from improved visibility.

<img width="330" alt="image" src="https://github.com/insightsengineering/teal.slice/assets/6700955/5e4e325c-3ab0-4471-9023-d47502cad6ea">

chlebowa:
I only see this when I zoom in with the browser.

100%
![image](https://github.com/insightsengineering/teal.slice/assets/114988527/a7acb636-d9ab-422e-bcd8-49532a8af6e8)

110%
![image](https://github.com/insightsengineering/teal.slice/assets/114988527/c2577e00-4b00-4c34-af9e-b10fc548afce)

120%
![image](https://github.com/insightsengineering/teal.slice/assets/114988527/4ba49caa-15ac-4720-8735-20e3d1282f47)

kartikeyakirar:
I can observe this  with 100% zoom on my laptop. Does the size of my laptop screen affect the partial obstruction on the display?

kartikeyakirar:
@m7pr  and @averissimo  can also check this?

m7pr:
I see this in the browser for zoom higher than 90%, where for 120% it even rotates axes' labels by 90 degrees

80%
![image](https://github.com/insightsengineering/teal.slice/assets/133694481/c55a4a8e-cedb-4463-8a3f-daeb31a08daf)


90%
![image](https://github.com/insightsengineering/teal.slice/assets/133694481/90eb72cd-30e6-499e-8067-74cce31bb191)


100%
![image](https://github.com/insightsengineering/teal.slice/assets/133694481/a61ea761-fec7-4954-8690-3b1df7b4dae5)


120%
![image](https://github.com/insightsengineering/teal.slice/assets/133694481/05ec59c5-5466-4e41-8f61-596ddf91805f)

chlebowa:
:laughing:

averissimo:
Linux / Firefox: never rotates the labels _(but that may just be due to few ticks)_ and only hides it on zoom levels > 100%

I second @chlebowa :laughing: 

![image](https://github.com/insightsengineering/teal.slice/assets/211358/4455920c-0bbc-44d5-a96c-dace0ffc02c7)

averissimo:
[...]

PR #442, opened 2023-08-16
The RangeFilterState's x-axis tick labels are somewhat obstructed.

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal.slice/issues/436

This problem doesn't occur in every system. While I managed to replicate it on a Mac, I couldn't reproduce it on my Windows computer. I'm somewhat uncertain whether this discrepancy is due to screen size or the operating system, but it appears that adjusting the` ticklen` helps resolve it. If the problem persists on other systems, we could attempt to address it by setting `ticks = "inside"` within Plotly's `plot_layout.`

### DIFF
--- a/R/FilterStateRange.R
+++ b/R/FilterStateRange.R
@@ -192,7 +192,7 @@ RangeFilterState <- R6::R6Class( # nolint
               rangeslider = list(thickness = 0),
               showticklabels = TRUE,
               ticks = "outside",
-              ticklen = 2,
+              ticklen = 1.5,
               tickmode = "auto",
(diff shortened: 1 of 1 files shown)


kartikeyakirar:
@m7pr  could you please check if this still exist for you?

chlebowa:
Works like a dream :tada: Let's just double check that @lcd2yyz is doesn't mind the shorter ticks.

lcd2yyz:
Looks okay to me, as long as it's still visible 😉

chlebowa (approved):

PR #444, opened 2023-08-16
add disabling and reenabling of rewind and reset buttons

### LINKED ISSUE #443: [Bug]: Application Crashes and Throws Script Out-of-Bounds Error When Rewind State Button is Clicked Twice
### What happened?

There is a critical bug in the application where clicking the rewind state button twice consecutively leads to a complete application crash. Additionally, this action triggers a script out-of-bounds error, displaying the message:` "Warning: Error in [[: subscript out of bounds."` 

image

image

chlebowa:
Fixes #443 

When state history changes, the rewind and reset buttons are hidden if history only contains the current state. The hiding is delayed because otherwise the buttons disappear for a moment when history is updated. This creates the possibility of the rewind button being clicked again during that delay window.

In this PR the buttons are disabled for the duration of the delay.

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -243,16 +243,26 @@ FilterState <- R6::R6Class( # nolint
             }
           )
 
+          # Buttons for rewind/reset are disabled upon change in history to prevent double-clicking.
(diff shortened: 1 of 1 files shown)


kartikeyakirar (approved):
This looks great! 👍

Issue #319, opened 2023-06-03
[Feature Request]: Allowing out of bound value input in range selection UI


lcd2yyz:
### Feature description

Relates to https://github.com/insightsengineering/teal.slice/pull/289 and https://github.com/insightsengineering/teal.slice/pull/316#issuecomment-1574575689

Currently behavior of our range selector UI: when user manually enters a value smaller/larger than the actual min/max of the data, the range value automatically snaps back to the min/max.

User should be allowed to specify any numerical value, including out-of-bound values, and expect the slider to apply that the chosen range selection for all subsequent analysis/reports.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

chlebowa:
Before we start on this, let us consider that removing out of bounds values from the selection is the behavior of _all_ `FilterState` classes. Should this behavior be removed for all data types or only numeric variables?

lcd2yyz:
Good question. I can only think of numerical variables that might require use of oob values - but in addition to the range slider, I would also consider date/datetime be a numerical values, so behaviour should be applied to date/datetime filter states classes.

Do you have more examples of other filter state classes that might need to consider this?

m7pr:
during the refinement we agreed that this card should only consider numeric input and should not focus on anything else

chlebowa:
Based on unit tests, we expect `ChoicesFilerState` to set selection to intersection of `selected` and `choices`, so I wouldn't touch those, personally. 

> during the refinement we agreed that this card should only consider numeric input and should not focus on anything else

There are distant plans to reorganize `FilterState` subclasses into range-based (numbers, dates, date-times) and selection-based (factors, logicals, anything with a low number of unique values). Unless they have been scrapped or put off indefinitely, I would like to repeat that discussion (which I missed) in this context.

It makes sense to me that dates and date-times allow out of bounds selections as well.

PR #449, opened 2023-08-29
407 omit is.infinite() if there are no Inf values in data for get_call()

### LINKED ISSUE #407: is.infinite() is returned in get_call() even though the variable does not have infinite values
During https://github.com/insightsengineering/teal.slice/issues/227 and https://github.com/insightsengineering/teal.slice/pull/402 a change of `get_call()` will be provided so that `is.na()` is not returned in the call if there are no NAs in the data even though `keep_na` was set to `TRUE`. I would recommend doing the same for `keep_inf = TRUE` and do not return `is.infinite()` if there are no infinite values within the variable.

Currently the below test passes because `is.infinite()` is returned, even though there are no infinite values in `Sepal.Length`

```{R}
testthat::test_that("get_call returns the filter call of the dataset", {
  dataset <- DefaultFilteredDataset$new(dataset = iris, dataname = "iris")
  fs <- teal_slices(
    teal_slice(dataname = "iris", varname = "Sepal.Length", selected = c(5.1, 6.4), keep_na = TRUE, keep_inf = TRUE),
    teal_slice(dataname = "iris", varname = "Species", selected = c("setosa", "versicolor"), keep_na = FALSE)
  )
  shiny::isolate(dataset$set_filter_state(state = fs))
  filter_call <- shiny::isolate(dataset$get_call())$filter

  testthat::expect_equal(
    filter_call,
    quote(
      iris <- dplyr::filter(
        iris,
          (is.infinite(Sepal.Length) | Sepal.Length >= 5.1 & Sepal.Length <= 6.4) &
          Species %in% c("setosa", "versicolor")
      )
    )
  )
})
```

m7pr:
Closes #407

### DIFF
--- a/R/FilterStateRange.R
+++ b/R/FilterStateRange.R
@@ -158,6 +158,7 @@ RangeFilterState <- R6::R6Class( # nolint
 
         checkmate::assert_numeric(slice$choices, null.ok = TRUE)
         if (is.null(slice$keep_inf) && any(is.infinite(x))) slice$keep_inf <- TRUE
+        if (isTRUE(slice$keep_inf) && !any(is.infinite(x))) slice$keep_inf <- FALSE
 
         private$set_choices(slice$choices)
         if (is.null(slice$selected)) slice$selected <- slice$choices
--- a/tests/testthat/test-FilteredDataset.R
+++ b/tests/testthat/test-FilteredDataset.R
@@ -102,6 +102,7 @@ testthat::test_that("print returns a string representation of filters", {
 
 # get_call ----
 testthat::test_that("get_call returns the filter call of the dataset", {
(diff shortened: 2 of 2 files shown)


chlebowa:
I'm not sure about this. 
If there is no `Inf`s in the data, `is.infinite` is only added to the call the `keep_inf = TRUE` in the `teal_slice`. Since the app user is unable to modify `teal_slice`s, the control rests with the app developer. Granted, this level of granularity is probably out of sight.

I guess we can make this change for the time being but this will come up again when we discuss [out of bounds values](https://github.com/insightsengineering/teal.slice/issues/319).

chlebowa:
Agreed :+1:

PR #456, opened 2023-08-30
415 range selector colors

### LINKED ISSUE #451: [Bug]: The visual state of filter is same when no data is selected vs when all data is selected
### What happened?

When the `count_type = "all"` in the `teal::teal_slices` and no filters are applied the numeric filter graph is fully colored and when we apply some filter the numeric filter graph is partially greyed out to indicate data that is unavailable for the filter.
However, when the filter is applied such that there is no data present in the numeric variable the colors still indicate that there is something to filter rather than greying out fully. Please check the GIF recording for more context.
Kapture 2023-08-29 at 16 41 11

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

chlebowa:
Fixes #451 

The error resulted from adding the second histogram to the plot, which is based on the filtered data. All data was filtered out, `x_reactive` returnes `integer(0)`, which results in all bins being colored.

Here, an `if` statement is added when creating the second histogram trace, such that the trace itself is `NULL` if there are no data.

### DIFF
--- a/R/FilterStateRange.R
+++ b/R/FilterStateRange.R
@@ -212,11 +212,13 @@ RangeFilterState <- R6::R6Class( # nolint
         })
         private$plot_filtered <- reactive({
           finite_values <- Filter(is.finite, private$x_reactive())
-          list(
-            x = finite_values,
-            bingroup = 1,
-            color = I(fetch_bs_color("primary"))
-          )
+          if (length(finite_values)) {
+            list(
+              x = finite_values,
+              bingroup = 1,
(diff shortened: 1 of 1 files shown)


kartikeyakirar:
Looks good! 

https://github.com/insightsengineering/teal.slice/assets/6700955/33f91249-d935-4ef5-902e-f0ad9381b818

kartikeyakirar (approved):

PR #461, opened 2023-09-01
911 c.teal_slices

chlebowa:
Revealed by [this issue](https://github.com/insightsengineering/teal/issues/911).

Modified `c.teal_slices` to handle attributes agnostically rather than explicitly.
Attributes are coalesced, i.e. the first non-NULL element is taken. This is applied to all attributes recursively. 
If a list attribute called `example` has element `name1 = value1` in one `teal_slices` and `name2 = value2` in another, the result will have `example = list(name1 = value1, name2 = value2)`. If the first `teal_slice` has `name1 = value1` and the other `name1 = value2`, then the result will have `example = list(name1 = value2)`.

This is handled by a new internal function, `coalesce_r`.


pawelru:
Hey @chlebowa. Can we please add some tests somewhere in the appropriate place so that it won't appear ever again?

chlebowa:
We will certainly try.

gogonzo (approved):
👍

PR #462, opened 2023-09-07
add onclick event to prevent default behavior which cause app to refresh

donyunardi:
Fixes insightsengineering/coredev-tasks#473

I prevented the default behavior on <a> that was created by `actionLink()` for the three buttons, rewind, reset, remove.
The fix is not obvious unless you deploy it.

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -333,6 +333,7 @@ FilterState <- R6::R6Class( # nolint
                   title = "Rewind state",
                   class = "filter-card-back",
                   style = "display: none",
+                  onclick = "event.preventDefault();",
                   href = "href"
                 )
               },
@@ -344,6 +345,7 @@ FilterState <- R6::R6Class( # nolint
                   title = "Restore original state",
                   class = "filter-card-back",
                   style = "display: none",
+                  onclick = "event.preventDefault();",
                   href = "href"
                 )
               },
@@ -353,6 +355,7 @@ FilterState <- R6::R6Class( # nolint
                   label = icon("circle-xmark", lib = "font-awesome"),
                   title = "Remove filter",
                   class = "filter-card-remove",
+                  onclick = "event.preventDefault();",
                   href = "href"
                 )
               }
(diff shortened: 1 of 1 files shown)


donyunardi:
I'm not sure how to add a test for this without involving shinytest2 on a deployed app.
I'll keep it like this for now.

chlebowa:
I don't understand why adding `preventDefault` to the buttons doesn't make them inert.

In the interest of reducing duplication, can you try adding your change to the div containing the buttons, `filter-card-controls`?

Also, have you checked you fix on other bootstrap versions? I found that `stopPropagation` was ignored under bs5.

averissimo:
## Context

After talking with @chlebowa we figured out that `preventDefault()` can be safely used in addition to `stopPropagation()` which would simplify the code avoiding repeated code.

> The preventDefault() method of the Event interface tells the user agent that if the event does not get explicitly handled, its default action should not be taken as it normally would be. [(docs)](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

Allowing Shiny to still catch the event. The `stopPropagation()` that @chlebowa added prevents the event from propagating to the parent `div`, while the `event.preventDefault()` will prevent it from scrolling up _(solved before with `href = "href"`)_

## Action 1

@donyunardi can you test the following deployed apps:

* https://connect.workbench.roche.com/adsconf_av_test_bs3/
* https://connect.workbench.roche.com/adsconf_av_test_bs4/
* https://connect.workbench.roche.com/adsconf_av_test_bs5/

They are part of the part of [teal.slice@473-link-issue](https://github.com/insightsengineering/teal.slice/tree/473-link-issue) branch

App deployed from [averissimo/example_deploy_remove_me](https://github.com/averissimo/example_deploy_remove_me) (branches `main`, `main-bs4`, `main-bs4`)

donyunardi:
Thanks for looking into this @chlebowa @averissimo 
Let's just do it on this PR for simplicity.

donyunardi:
@averissimo 
I didn't realize that you already had created `473-link-issue` branch.
Let's go with the branch since it's less commits.

averissimo:
I only created that branch so that the code could be shared, but ok. I'll create a new PR :+1:

PR #463, opened 2023-09-08
add single onclick event to prevent default behavior which cause app to refresh

averissimo:
Fixes insightsengineering/coredev-tasks#473

- Follow-up on #462

### Changes description

- Remove `href` that was causing an issue on deployed apps
- Single use of `event.preventDefault()` chained after `event.stopPropagation()`
  - first will prevent any navigation action by the browser (event is still caught by Shiny)
  - second will prevent the event from bubbling to parent elements _(that is, parents won't trigger the onclick event)_

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -321,7 +321,7 @@ FilterState <- R6::R6Class( # nolint
               class = "filter-card-controls",
               # Suppress toggling body when clicking on this div.
               # This is for bootstrap 3 and 4. Causes page to scroll to top, prevented by setting href on buttons.
-              onclick = "event.stopPropagation();",
+              onclick = "event.stopPropagation();event.preventDefault();",
               # This is for bootstrap 5.
               `data-bs-toggle` = "collapse",
               `data-bs-target` = NULL,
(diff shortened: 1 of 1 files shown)


chlebowa:
Could you please deploy an app from this branch (can be any one from `teal.gallery`) just to tripple check that the fix works?

donyunardi (approved):
Deployed an example app to my personal account.
https://donyunardi.shinyapps.io/early-dev/

Tested all buttons (and more) and everything looks good.

Here's the lock file
[renv copy.txt](https://github.com/insightsengineering/teal.slice/files/12563958/renv.copy.txt)

donyunardi:
ran the workflow for teal.gallery, tested early-dev_dev and exploratory_dev, and I confirmed the refresh is no longer happening.
Thanks everyone!

averissimo:
Great!

Here's the [`renv.lock`](https://github.com/averissimo/teal.gallery/blob/test_deployment%40473-link-issue/longitudinal/renv.lock) to deploy the longitudinal on rsconnect

PR #432, opened 2023-08-09
373 slices_restore will restore Dates and POSIXT classes

m7pr:
Closes #373 
A very small proposition on how we can guess that a character contains a Date a POSIXT time. Up to be discussed as we might want to just use jsonlite::serializeJSON and jsonlite::unserializeJSON

# The code

```{R}
x <- 
  teal_slice(
    dataname = "ADSL", 
    varname = "COUNTRY", 
    selected = as.POSIXct(c("2023-06-29 18:05:32 CEST", "2023-06-29 18:15:32 CEST")), 
    fixed = TRUE
  )
class(as.list(x)$selected)

t <- 
  teal_slice(
    dataname = "ADSL", 
    varname = "TIME", 
    selected = as.Date(c("2023-06-29", "2023-06-30")), 
    fixed = TRUE
  )
class(as.list(t)$selected)

y <- teal_slices(x, t)
slices_store(y, file = 'test1.json')

z <- slices_restore('test1.json')

class(shiny::isolate(z[[1]]$selected))
class(shiny::isolate(z[[2]]$selected))

````

# The code with results

```{R}
> x <- 
+   teal_slice(
+     dataname = "ADSL", 
+     varname = "COUNTRY", 
+     selected = as.POSIXct(c("2023-06-29 18:05:32 CEST", "2023-06-29 18:15:32 CEST")), 
+     fixed = TRUE
+   )
> class(as.list(x)$selected)
[1] "POSIXct" "POSIXt" 
> 
> t <- 
+   teal_slice(
+     dataname = "ADSL", 
+     varname = "TIME", 
+     selected = as.Date(c("2023-06-29", "2023-06-30")), 
+     fixed = TRUE
+   )
> class(as.list(t)$selected)
[1] "Date"
> 
> y <- teal_slices(x, t)
> slices_store(y, file = 'test1.json')
> 
> z <- slices_restore('test1.json')
> 
> class(shiny::isolate(z[[1]]$selected))
[1] "POSIXct" "POSIXt" 
> class(shiny::isolate(z[[2]]$selected))
[1] "Date"
```

PR #468, opened 2023-09-27
#467 slices_restore restores factors

### LINKED ISSUE #467: `slices_restore` do not restore `factor` classes
### What happened?

Similarly to https://github.com/insightsengineering/teal.slice/issues/373 slices_restore restore `factor` as `character`

```r
x <- 
  teal_slice(
    dataname = "ADSL", 
    varname = "COUNTRY", 
    selected = factor(c('POLAND', 'SPAIN')), 
    fixed = TRUE
  )
class(as.list(x)$selected)
#> [1] "factor"

y <- teal_slices(x)
slices_store(y, file = 'test1.json')
z <- slices_restore('test1.json')
class(shiny::isolate(z[[1]]$selected))
#> [1] "character"
```

Potential fixes need to be provided in https://github.com/insightsengineering/teal.slice/blob/main/R/FilterStateChoices.R#L138
where `selected` or `choices` of class `factor` may be all transferred to `characters`. Check out what actions/changes does it populate in other parts of the tool

m7pr:
Closes #467 
A continuation of https://github.com/insightsengineering/teal.slice/pull/432

In this short PR we enhanced `format` with a new parameter called `add_levels` that allows to store `levels` of `factor` variables for `selected` and `choices` fields. This is only used in `store` where we store levels next to variables in new fields called `selected_levels` and `choices_levels`. Those two fields gets removed on a `restore` procedure. 

I extended tests to prove that results for `factors` and `characters` for `selected` and `choices` are the same when storing and restoring.

### DIFF
--- a/R/teal_slice-store.R
+++ b/R/teal_slice-store.R
@@ -31,7 +31,7 @@ slices_store <- function(tss, file) {
   checkmate::assert_class(tss, "teal_slices")
   checkmate::assert_path_for_output(file, overwrite = TRUE, extension = "json")
 
-  cat(format(tss, trim_lines = FALSE), "\n", file = file)
+  cat(format(tss, trim_lines = FALSE, add_levels = TRUE), "\n", file = file)
 }
 
 #' Restore teal_slices object from a file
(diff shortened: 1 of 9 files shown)


chlebowa:
Again, do we need saving factor levels?

m7pr:
hmm, alrighty, so it does not make any sense to keep selected or choices as factors, we can reshape them to characters on a teal_slice creation?

m7pr:
Alternative in here https://github.com/insightsengineering/teal.slice/pull/469

m7pr:
we decided to go with https://github.com/insightsengineering/teal.slice/pull/469

PR #470, opened 2023-10-04
Introduce teal_data class

gogonzo:
Following:
- https://github.com/insightsengineering/teal.data/pull/178

* constructor `init_filtered_data.tdata` will be removed. teal should call default method. `init_filtered_data.tdata` is introduced as a convenience during a POC coding.


DESCRIPTION
```diff
@@ -45,6 +45,7 @@ Imports:
     shinycssloaders (>= 1.0.0),
     shinyjs,
     shinyWidgets (>= 0.6.2),
+    teal.code (>= 0.4.1),
     teal.data (>= 0.3.0),
```
chlebowa, line 49: This will have to be bumped, I guess.

R/FilteredData-utils.R
```diff
@@ -51,11 +57,32 @@ init_filtered_data.TealData <- function(x, # nolint
...
+
+    # todo: to be removed soon https://github.com/insightsengineering/teal.slice/issues/471
+    code = teal.data:::CodeClass$new(
+      code = paste(teal.code::get_code(x), collapse = "\n"),
+      dataname = teal.data::get_dataname(x)
+    ),
+    # todo: to remove
+    check = check
+  )
+}
```
chlebowa, line 79: Looks like `join_keys` is ignored, so the default value is redundant.
I guess `join_keys`, `code` and `check` are only here to maintain consistent arguments between methods?

gogonzo: Yes, but this is S3 method so this argument has to be listed. Anyway, don't you think that `init_filtered_data` constructor should be generic only, and `init_filtered_data` should be build from `teal_data` in the internals of teal?

We should give an answer in a near future, when we attempt to include teal_data to the modules. Then we decide what will be the role of FilteredData and whether FD will conduct the code from data loading stage.

DESCRIPTION
gogonzo: I wonder why I bumped teal.code instead of teal.data 🤔 I think I have mistaken

R/FilteredData-utils.R
chlebowa: > Yes, but this is S3 method so this argument has to be listed. 

ok :+1: 

> Anyway, don't you think that `init_filtered_data` constructor should be generic only, and `init_filtered_data` should be build from `teal_data` in the internals of teal?

We won't be needing the multiple methods soon, so it would make sense to have a single function, yes. But then it would be impossible to instantiate `FilteredData` with a list of datasets.

> Then we decide what will be the role of FilteredData

Good point, we are nearing a state where the filter panel has `teal_data` in and `teal_data` out.

pawelru:
Please add NEWS + vbump deps if needed

chlebowa (approved):
I can no longer find anything wrong here :ok_hand:

R/FilteredData-utils.R
m7pr: Maybe passing a note that `join_keys` is ignored and taken from `x@join_keys` would be helpful

PR #475, opened 2023-10-12
Replace the broken showcase gif with a new gif

### LINKED ISSUE #474: [Bug]: Showcase gif is broken
https://github.com/insightsengineering/teal.slice/blob/main/assets/img/showcase.gif

This gif that is used in README to showcase the filter panel is cut in the middle of the animation. I think we should come up with a new gif that presents basic features of filter panel

vedhav:
Closes #474


m7pr (approved):
thanks for the help!

PR #488, opened 2023-11-15
471 remove `CodeClass` from `FilteredData`

chlebowa:
Removed private fields `$code` and `$check` from `FilteredData` class and all associated methods.
Removed `code` and `check` arguments from `init_filtered_data.default`.

[companion](https://github.com/insightsengineering/teal/pull/964)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal.slice 0.4.0.9022
+# teal.slice 0.4.0.9023
 
 ### Miscellaneous
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,7 @@
 * Setting filters using a list is now deprecated. Use `teal_slices` and `teal_slice` instead.
 * Removed `CDISCFilteredData` and `CDISCFilteredDataset` and implementing `JoinKeys` handling in their parent classes (`FilteredData` and `DefaultFilteredDataset`).
 * Specifying set of filterable columns is done through `include_varnames` and `exclude_varnames` in `teal_slices`. Specifying `attr(, "filterable")` is hard deprecated.
+* Removed private fields `$code` and `$check` from `FilteredData` class and made appropriate changes to constructor and `init_filtered_data`.
 
 # teal.slice 0.3.0
 
--- a/R/FilteredData-utils.R
+++ b/R/FilteredData-utils.R
@@ -60,12 +60,10 @@ init_filtered_data.TealData <- function(x, # nolint
 
 #' @keywords internal
 #' @export
-init_filtered_data.default <- function(x, join_keys = teal.data::join_keys(), code = NULL, check = FALSE) { # nolint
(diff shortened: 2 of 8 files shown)


NEWS.md
```diff
@@ -27,6 +27,7 @@
 * Setting filters using a list is now deprecated. Use `teal_slices` and `teal_slice` instead.
 * Removed `CDISCFilteredData` and `CDISCFilteredDataset` and implementing `JoinKeys` handling in their parent classes (`FilteredData` and `DefaultFilteredDataset`).
 * Specifying set of filterable columns is done through `include_varnames` and `exclude_varnames` in `teal_slices`. Specifying `attr(, "filterable")` is hard deprecated.
+* Removed private fields `$code` and `$check` from `FilteredData` class and made appropriate changes to constructor and `init_filtered_data`.
```
gogonzo, line 30: Private fields is not a concern of end user. Please write about changes in a `init_filtered_data` only, that it no o
```suggestion
*  Removed `check` and `code` from formals of `FilteredData$initialize`. Storing arbitrary `code` and `check` in a `FilteredData` class is no longer possible.
```

R/FilteredData-utils.R
```diff
@@ -60,12 +60,10 @@ init_filtered_data.TealData <- function(x, # nolint
 
 #' @keywords internal
 #' @export
-init_filtered_data.default <- function(x, join_keys = teal.data::join_keys(), code = NULL, check = FALSE) { # nolint
+init_filtered_data.default <- function(x, join_keys = teal.data::join_keys(), code, check) { # nolint
```
gogonzo, line 63:
```suggestion
init_filtered_data.default <- function(x, join_keys = teal.data::join_keys(), code, check) { # nolint
  lifecycle::deprecate_stop(
    "0.4.1", 
    "init_filtered_data(check = 'No longer supported')"
  )
  lifecycle::deprecate_stop(
    "0.4.1", 
    "init_filtered_data(code = 'No longer supported')"
  )
```

gogonzo (changes requested):
`code` and `check` are still in the documentation of `init_filtered_data` but they are ignored. Please fix @param check and @param code and write there that they are deprecated

chlebowa:
> `code` and `check` are still in the documentation of `init_filtered_data` but they are ignored. Please fix @param check and @param code and write there that they are deprecated

They have to stay because of other methods and the generic. Will be removed along with old classes.

R/FilteredData-utils.R
chlebowa: This raises errors on all calls, even when the arguments are not passed. I suppose it is because the `NULL` defaults in the generic.

gogonzo: okey, let's do this on a feature branch or don't do this at all

gogonzo (approved):
looks good

PR #504, opened 2023-12-27
490 error on GRAN

chlebowa:
Fixes https://github.com/insightsengineering/coredev-tasks/issues/490

As of `R 4.4.0` `NULL` is no loner considered `atomic`, which interrupted the logic in `coalesce_r`. The logic has been updated.

### DIFF
--- a/R/teal_slices.R
+++ b/R/teal_slices.R
@@ -271,8 +271,9 @@ setdiff_teal_slices <- function(x, y) {
 #'
 coalesce_r <- function(x) {
   checkmate::assert_list(x)
-  if (all(vapply(x, is.atomic, logical(1L)))) {
-    return(Filter(Negate(is.null), x)[[1L]])
+  xnn <- Filter(Negate(is.null), x)
+  if (all(vapply(xnn, is.atomic, logical(1L)))) {
+    return(xnn[[1L]])
   }
   lapply(x, checkmate::assert_list, names = "named", null.ok = TRUE, .var.name = "list element")
   all_names <- unique(unlist(lapply(x, names)))


m7pr:
Lol on the NULL not being atomic

R/teal_slices.R
```diff
@@ -271,8 +271,9 @@ setdiff_teal_slices <- function(x, y) {
 #'
 coalesce_r <- function(x) {
   checkmate::assert_list(x)
-  if (all(vapply(x, is.atomic, logical(1L)))) {
-    return(Filter(Negate(is.null), x)[[1L]])
+  xnn <- Filter(Negate(is.null), x)
```
pawelru, line 274: Q: should we use not-null list elements going further after early return?

```suggestion
  x <- Filter(Negate(is.null), x)
```

chlebowa: Your suggestion valid, meaning it produces the intended outcome.
The reason I did it this way is I didn't want to disrupt the recursive logic: the heart of the function is lines 278-280 and so I only modified the exit condition. And I used `Filter` because I thought it clearer than 
`all(vapply(x, function(x) !is.null(x) && is.atomic(x), logical(1L)))`.

What do you think?

pawelru: got it and agreed - let's not interfere with the logic. I am also seeing explicit `null.ok = TRUE` right afterwards that indicates that null values are being considered.
Please just note that even though its clearer it's memory inefficient - you have made a hard copy of (potentially big) object. Usually when I see a list type of data I am automatically assuming a _heavy_ objects. Now please tell me if that's an issue - I mean: how big objects are expected here. The _uglier_ solution does not have this problem.

chlebowa: A `list` of a thousand `NULL`s is only 848 bytes so I don't think that will be a problem.

As far as stack inflation resulting from recursion, this function is used to put together attributes from an unspecified number of `teal_slices` objects during concatenation. These attributes can be single-element vectors (logical or character) or lists of character vectors. In the latter case the list can be as long as the number of datasets in the app times two and the list elements can be as long as the number of variables in the respective datasets.
All in all - not too large _with the current use cases_, IMO.

pawelru: > A `list` of a thousand `NULL`s is only 848 bytes so I don't think that will be a problem.

according to the current definition `xnn` would not duplicate nulls but not-null values instead

> this function is used to put together attributes

> All in all - not too large with the current use cases, IMO.

Fair enough. Approving 👍

pawelru (approved):

PR #501, opened 2023-12-21
Change dispatch method to S3 instead of custom (target: `@132_filter_panel_ui@main`)

averissimo:
Tested with:

- Overwriting one of the `get_filter_call` method  with a blank package called {overwrite} (see below)
  - The `@importFrom` roxygen2 tag is fundamental
- Custom teal app (at the end)

note: My concern with the order of library / pkgload::load_all doesn't matter as {overwrite} imports from {teal.slice} and will re-register the method

#### {overwrite} package

```r
## NAMESPACE

# Generated by roxygen2: do not edit by hand

S3method(get_filter_call,data.frame)
importFrom(teal.slice,get_filter_call)

## R/hello.R

#' A title
#' @importFrom teal.slice get_filter_call
#' @export
get_filter_call.data.frame <- function(data, states_list) {
  print("This was overwritten by {overwrite}")
}
```

##### Example of the {overwrite} in place

Note that this is non-reproducible as the {overwrite} package needs to be installed

``` r
library(teal)
#> Loading required package: shiny
#> Loading required package: teal.data
#> Loading required package: teal.code
#> Loading required package: teal.slice
#> Loading required package: teal.transform
#> Loading required package: magrittr
#> Registered S3 method overwritten by 'teal':
#>   method        from      
#>   c.teal_slices teal.slice
#> 
#> You are using teal version 0.14.0.9028
#> 
#> Attaching package: 'teal'
#> The following objects are masked from 'package:teal.slice':
#> 
#>     as.teal_slices, teal_slices

get_filter_call(data.frame(a = 1:10), "doesn't matter")
#> Error in states_list[[1L]]$get_state: $ operator is invalid for atomic vectors

library(overwrite)
#> Registered S3 method overwritten by 'overwrite':
#>   method                     from      
#>   get_filter_call.data.frame teal.slice

get_filter_call(data.frame(a = 1:10), "doesn't matter")
#> [1] "This was overwritten by {overwrite}"
```

##### Teal app

Tested with branch of {teal} `132_filter_panel_ui@main`.

```r
library(teal)

td <- teal.data::teal_data() |> within({
  new_iris = transform(iris, id = seq_len(nrow(iris)))
  new_mtcars = transform(mtcars, id = seq_len(nrow(mtcars)))
  array = array(1:3, c(2,4))
  matrix = matrix(c(1,2,3, 11,12,13, 21, 22, 23), nrow = 2, ncol = 3, byrow = TRUE,
                  dimnames = list(c("row1", "row2"),
[...]

PR #505, opened 2023-12-27
Use `S4` dispatch for filter panel methods

chlebowa:
Since #501 failed, here is another attempt using `S4`. It seems the method registration mechanism is more suited to our purpose.

Documentation has not been updated, still mentions S3 in instructions for adding methods.

It has been suggested that the methods be rearranged into files class-wise rather than generic-wise, because that is how a user would add support for their own class. In that vein, we could add a function that would prepare such a file, something like `methods::method.skeleton` but for the entire suite, including some `roxygen`.


R/methods-get_filter_call.R
```diff
@@ -1,24 +1,40 @@
...
-}
+#'
+# get_filter_call generic ----
+setGeneric("get_filter_call", function(data, states_list) {
+  NULL
+})
+
+## default method ----
+setMethod("get_filter_call", c(data = "ANY"), function(data, states_list) {
+  if (length(states_list) == 0L) NULL
```
averissimo, line 29:
```suggestion
  if (length(states_list) == 0L) return(NULL)
```

R/methods-get_filter_call.R
```diff
@@ -1,24 +1,40 @@
...
+
+  stop("get_filter_call not implemented for class ", toString(class(data)), call. = FALSE)
+})
+
+## data.frame method ----
+setMethod("get_filter_call", c(data = "data.frame"), function(data, states_list) {
+  if (length(states_list) == 0L) NULL
```
averissimo, line 36:
```suggestion
  if (length(states_list) == 0L) return(NULL)
```

R/methods-get_filter_call.R
```diff
@@ -28,11 +44,27 @@ get_filter_call.data.frame <- function(data, states_list) {
     expr = dataname_lang <- rhs,
     env = list(dataname_lang = dataname_lang, rhs = rhs)
   )
-}
+})
+
+## DataFrame method ----
+setMethod("get_filter_call", c(data = "DataFrame"), function(data, states_list) {
+  if (length(states_list) == 0L) NULL
```
averissimo, line 51:
```suggestion
  if (length(states_list) == 0L) return(NULL)
```

R/methods-get_filter_call.R
```diff
@@ -28,11 +44,27 @@ get_filter_call.data.frame <- function(data, states_list) {
...
+
+  substitute(
+    expr = dataname_lang <- rhs,
+    env = list(dataname_lang = dataname_lang, rhs = rhs)
+  )
+})
+
+## matrix method ----
+setMethod("get_filter_call", c(data = "matrix"), function(data, states_list) {
+  if (length(states_list) == 0L) NULL
```
averissimo, line 66:
```suggestion
  if (length(states_list) == 0L) return(NULL)
```

R/methods-get_filter_call.R
```diff
@@ -42,11 +74,58 @@ get_filter_call.array.Matrix <- function(data, states_list) {
     expr = dataname_lang <- rhs,
     env = list(dataname_lang = dataname_lang, rhs = rhs)
   )
-}
+})
+
+## Matrix method ----
+setMethod("get_filter_call", c(data = "Matrix"), function(data, states_list) {
+  if (length(states_list) == 0L) NULL
```
averissimo, line 81:
```suggestion
  if (length(states_list) == 0L) return(NULL)
```

R/methods-get_filter_call.R
```diff
@@ -42,11 +74,58 @@ get_filter_call.array.Matrix <- function(data, states_list) {
...
+
+  substitute(
+    expr = dataname_lang <- rhs,
+    env = list(dataname_lang = dataname_lang, rhs = rhs)
+  )
+})
+
+## SummarizedExperiment method ----
+setMethod("get_filter_call", c(data = "SummarizedExperiment"), function(data, states_list) {
+  if (length(states_list) == 0L) NULL
```
averissimo, line 96:
```suggestion
  if (length(states_list) == 0L) return(NULL)
```

R/methods-module_active.R
```diff
@@ -15,39 +14,61 @@
...
+})
 
-#' @rdname module_active
-#' @export
-srv_active <- function(id, data, reactive_state_list, remove_state_callback) {
-  UseMethod("srv_active", data)
-}
+## default method ----
+setMethod("ui_active", c(data = "ANY"), function(id, data, label) {
+  stop("ui_active not implemented for class ", toString(class(data)), call. = FALSE)
```
averissimo, line 53:
```suggestion
  tagList(NULL)
```

R/methods-module_active.R
```diff
@@ -75,121 +137,434 @@ ui_active_MultiAssayExperiment <- function(id, data, label) {
...
+setGeneric("srv_active", function(id, data, reactive_state_list, remove_state_callback) {
+  NULL
+})
+
+## default method ----
+setMethod(
+  "srv_active",
+  c(data = "ANY"),
+  function(id, data, reactive_state_list, remove_state_callback) {
+    stop("srv_active not implemented for class ", toString(class(data)), call. = FALSE)
```
averissimo, line 156:
```suggestion
    NULL
```

averissimo (commented):
Looks good!

I made some changes to be able to run the example app 

Note, that it needs `teal@132_filter_panel_ui@main` version to be able to run.

<details>

<summary>Install depedencies</summary>

```R
staged.dependencies::dependency_table(
  project = "insightsengineering/teal@https://github.com",
  project_type = "repo@host",
  ref = "132_filter_panel_ui@main",
  verbose = 1
) |> staged.dependencies::install_deps()
```
</details>

<details>

<summary>Example app (with a lot of data types)</summary>

```R
withr::local_options(
  list(
    teal.log_level = "TRACE",
    teal.show_js_log = TRUE,
    teal.bs_theme = bslib::bs_theme(version = 3)
  )
)

tdata <- teal_data() |>
 within({
  set.seed(1)
  library(scda)
  library(airway)
  library(MultiAssayExperiment)
  library(stats)
  ts_example <- ts(matrix(rnorm(300), 100, 3), start = c(1961, 1), frequency = 12)
  data(miniACC, envir = environment())
  data(airway, envir = environment())
  a_matrix <- matrix(rnorm(100), ncol = 5, dimnames = list(rownames = NULL, colnames = c("a", "b", "c", "d", "e")))
  a_vector <- "elo"
})

datanames(tdata) <- c("ADSL", "ADTTE", "ADRS", "miniACC", "airway", "a_vector", "a_matrix", "ts_example")
join_keys(tdata) <- default_cdisc_join_keys[c("ADSL", "ADTTE", "ADRS")]

default_filters <- teal.slice::teal_slices(
  teal_slice(dataname = "ADSL", varname = "categorical", selected = c("a"), id = "categorical", multiple = FALSE),
  count_type = "all"
)

setMethod("get_filter_call", c(data = "data.frame"), function(data, states_list) {
  cat("\n\nTHIS IS A LOG--------------------------------------------------------------\n\n")
  if (length(states_list) == 0L) return(NULL)

  dataname_lang <- str2lang(states_list[[1L]]$get_state()$dataname)
  states_predicate <- lapply(states_list, function(state) state$get_call())
  combined_predicate <- calls_combine_by(states_predicate, "&")
  rhs <- as.call(c(str2lang("dplyr::filter"), c(list(dataname_lang), combined_predicate)))

  substitute(
    expr = dataname_lang <- rhs,
    env = list(dataname_lang = dataname_lang, rhs = rhs)
  )
})

app <- init(
  data = tdata,
  modules = modules(
    modules(
      label = "tab1",
      example_module("funny"),
      example_module("funny", datanames = "airway"),
      example_module("funny2", datanames = "ADTTE"), # will limit datanames to ADTTE and ADSL (parent),
      teal.modules.general::tm_data_table()
    )
  ),
  filter = default_filters
)

runApp(app)
```
</details>

image

R/methods-get_filter_call.R
chlebowa: By `NEST` policy we avoid explicit `return` calls at the end of functions.

R/methods-module_active.R
chlebowa: Fair point :+1:

R/methods-get_filter_call.R
averissimo: Take a look at the code as these are not the last expression on the function and will do nothing.

That is, it won't return `NULL` and just continue to the next expression

averissimo: This is valid to every other comment.

If you prefer, wrap the following expressions around an `else` and keep it without `return()` call

chlebowa: You're right.
Now that I look at it, I'm not sure why it is written this way, actually.

averissimo:
One last thing, if we load the package before `MultiAssayExperiment` we get a bunch of warnings.

We need to figure out how to avoid these (especially if {MAE} is not installed)

image

chlebowa:
It's not just about `MultiAssayExperiment`, actually. `DataFrame` is from `S4Vectors`, which is always installed, I believe. I tried adding some `#' @importClassesFrom package class` but with no success.
You're right, it's something we should figure out if we choose to go this way.

PR #506, opened 2023-12-28
Pre-release activities

kartikeyakirar:
part of https://github.com/insightsengineering/nestdevs-tasks/issues/48

- [x] Review and update:
      - README.md (check example code)
      - NEWS.md
- [x] Run urlchecker::url_check() to identify broken links and fix
- [x] Review functions:
       - @example tag, make sure it runs, fix if otherwise
       - Make sure functions has @return tag to document the return value
       - no \dontrun tag, replace with if(interactive()) if needed
- [x] Sanity check of all vignettes, make sure there is no typo, no wrong format, etc.
- [x] Run R CMD check --as-cran make sure everything pass
- [x] Package Title is not duplicated in Package Description in DESCRIPTION file (e.g. this happens in teal.slice currently)
- [x] All package names in Title and Description fields of DESCRIPTION file are quoted with '
- [x] You have checked the Package Release Template https://github.com/insightsengineering/teal.reporter/pull/205/files
- [x] Make sure there are no ::: in examples
- [x] Make sure all teal.* mentions are lower-cased and quoted
- [x] Make Sure inst/WORDLIST is minimalized
- [x] Make sure non-exported functions do not have examples
- [x] Make sure each link to our documentation hosted with pkgdown on github pages do not have /main/ in the address but it has /latest/ instead, so we always expose the documentation of the latest release and not what's currently on main branch but not yet released

PR #508, opened 2024-01-03
507 reevaluate `variable_types`

### LINKED ISSUE #507: reevaluate `variable_types` function
`variable_types` is a generic function with methods for `data.frame`, `DataTable`, `DFrame`, `matrix` and a `default` method.

Actually, the `default `method handles `data.frame`, `DataTable`, and `DFrame`, the methods for these classes merely call `default`. 
Q1. Why do these methods even exist?

The `matrix` method goes through all specified columns and gets their classes, which is redundant because a matrix is an atomic.
Q2. Is it really necessary to have a generic with five methods? Why not have a single function that asserts input type to be a data.frame-like or a matrix and have a single `if` condition in there? This would have the added benefit of failing on data types that are not explicitly supported.

Q3. Bonus question: What is a `DataTable` class?

chlebowa:
Closes #507 

Replaced `variable_types` function family with single function.
The new function returns a named vector so no need to use `setNames` in its caller.


R/variable_types.R
```diff
@@ -30,76 +30,23 @@
...
-          vapply(columns, identical, logical(1L), "")
-      )
-    )
-    vapply(
-      columns,
-      function(x) ifelse(x == "", "", class(data[, x])[1]),
-      character(1),
-      USE.NAMES = FALSE
-    )
+  checkmate::assert_multi_class(data, c("data.frame", "DFrame", "matrix"))
```
gogonzo, line 33: I realized "DFrame" inherits from "DataFrame". Please change to `DataFrame` to include more cases.
Skärmavbild 2024-01-08 kl  11 50 57


```suggestion
  checkmate::assert_multi_class(data, c("data.frame", "DataFrame", "matrix"))
```

chlebowa: The `DataFrame` function produces a `DFrame` object:
```
> DataFrame() %>% class
[1] "DFrame"
attr(,"package")
[1] "S4Vectors"
```

Do we support `SQLDataFrame` and `ParquetDataFrame`? I don't think we depend on packages they are defined in.

gogonzo:
I think we can include this before release.

R/variable_types.R
chlebowa: Right, I see now that 
```
> DataFrame() %>% is("DataFrame")
[1] TRUE
```

gogonzo: > DataFrame() %>% class 

This is surprising. However in S4Vectors there is "RectangularData" class which is a parent of `DataFrame` and `DelayedMatrix`. I don't expect any of these objects to be present in MAE@colData but this is a tip for this PR https://github.com/insightsengineering/teal.slice/pull/498

R/variable_types.R
```diff
@@ -30,76 +30,23 @@
...
-          vapply(columns, identical, logical(1L), "")
-      )
-    )
-    vapply(
-      columns,
-      function(x) ifelse(x == "", "", class(data[, x])[1]),
-      character(1),
-      USE.NAMES = FALSE
-    )
+  checkmate::assert_multi_class(data, c("data.frame", "DataFrame", "matrix"))
```
gogonzo, line 33: There is `DFrame` in the docs of `data` argument

chlebowa: Docs will be updated after #506 as that removes examples.

chlebowa, line 33: Shall I add `"Martix"` here as well? `Matrix` does not inherit from `matrix`.

chlebowa: Hmm :thinking:  Allowing `Matrix` would require `MatrixFilterStates` for `Matrix`, wouldn't it? I guess we don't support `Matrix` at the moment. No addition then.

pawelru:
Watch out here! TMG is using this private function: https://github.com/insightsengineering/teal.modules.general/blob/38ff2b0e06afa226d2507bc0100801bd7a663a99/R/tm_variable_browser.R#L1192

chlebowa:
Thanks or the warning @pawelru. The `setNames` is redundant and in no way disruptive. I will look find other uses of this function in other packages and test them.

PR #510, opened 2024-01-05
use demos to store examples for internal functions

chlebowa:
Examples using internal functions are converted to demos. See `?utils::demo`.

Added `demo/` directory in package root.
`demo/` contains `00Index` file (required), which lists all demos and their descriptions. (A minimum of 3 spaces are required as column separator.)
`demo/` contains examples taken from `FilterState-utils.R` with some `readline` calls for pacing.
`FilterState-utils` examples call the demo under `if (interactive())`.

### DIFF
--- a/R/FilterState-utils.R
+++ b/R/FilterState-utils.R
@@ -19,6 +19,9 @@
 #' }
 #' @param ... additional arguments to be saved as a list in `private$extras` field
 #'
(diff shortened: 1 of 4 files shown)


chlebowa:
Abandoned in favor of using `getFromNamespace` in examples.

PR #513, opened 2024-01-08
Update images@pre release cleanup@main

gogonzo:
Fix images after https://github.com/insightsengineering/teal.slice/pull/506#pullrequestreview-1808818075


chlebowa:
The name of the class is `DataframeFilteredDataset` - note the lowercase `f`. This possibly results from a typo I made when renaming the class. Would you like me to correct the text or do you prefer to change the images?

PR #515, opened 2024-01-09
511 remove `check_ellipsis` function

### LINKED ISSUE #511: remove `check_ellipsis`
The `check_ellipsis` function (defined in `teal.slice` and `utils.teal`) does not seem to be used at all.
Verify that is is redundant and remove if so.

chlebowa:
Closes #511


averissimo (approved):
Seems like a safe removal of an unused non-exported function.

Last occurence of the function was removed on #165 (~6 months ago)

Issue #500, opened 2023-12-20
Separate bioconductor dependency from`teal.slice`


donyunardi:
### Summary
Prior discussion [here](https://github.com/insightsengineering/teal.slice/issues/499#issuecomment-1859921788)
In preparation to release teal.slice to CRAN, we will separate the MultiAssayExperiment and SummarizedExperiment dependencies to a separate package (extends `teal.slice`) and release that in BioConductor.

The proposed package name is `teal.slice.bio`.

### Definition of Done
* Create a repo in insightengineering for teal.slice.bio (IDR)
* teal.slice.bio will be an extension of teal.slice with MultiAssayExperiment and SummarizedExperiment dependencies
* MultiAssayExperiment and SummarizedExperiment no longer dependencies of `teal.slice`
* Update package dependencies for modules (i.e. tm.helios, tm.hermes) to use `teal.slice.bio`

donyunardi:
cc'ing @danielinteractive @BFalquet for awareness

lcd2yyz:
Agree with the proposed concept, but suggest to name new pkg to `teal.slice.bioc`

PR #518, opened 2024-01-11
Skip tests/examples when packages in Suggest section are not installed

averissimo:
- Part of pre-release activities #506 (base branch for this PR is `pre-release-cleanup@main`)
- Rationale on https://github.com/insightsengineering/idr-tasks/issues/713

### Changes description

- [x] Handles `@examples`
  - with `@examplesIf requireNamespace("MultiAssayExperiment")` as suggested [here](https://indrajeetpatil.github.io/preventive-r-package-care/#/soft-dependency-hygiene)
  - Conditional clause replacing `miniACC` with empty data.frame
- [x] Skips tests that use {MAE} or {SE}
- [x] Skips tests that use {withr}

#### Discussions

- > That's not a reason to skip tests. MAE is in Suggests so installing is the responsibility of the person who wants to run tests, i.e. not the end user. If someone uses teal with MAE data, they most likely have the package installed.
- Some generic tests and examples for `filter_state_api` are complex cases that join together normal data frames with MAE
    - These could be broken down to 2 examples.
- If we merge this it also means tests using {withr} should be skipped
  - Seems silly, but necessary to be a complete compliance

WDYT?

### DIFF
--- a/R/FilteredDataset-utils.R
+++ b/R/FilteredDataset-utils.R
@@ -40,6 +40,7 @@
 #'   shinyApp(ui, server)
 #' }
 #'
+#' @examplesIf requireNamespace("MultiAssayExperiment")
 #' # MAEFilteredDataset example
 #' library(MultiAssayExperiment)
 #' data(miniACC)
--- a/R/filter_panel_api.R
+++ b/R/filter_panel_api.R
@@ -23,22 +23,53 @@
 #' @seealso [`teal_slice`]
 #'
 #' @examples
-#' if (requireNamespace("MultiAssayExperiment", quietly = TRUE)) {
-#'   library(MultiAssayExperiment)
-#'   data(miniACC)
-#' } else {
-#'   miniACC <- data.frame()
-#' }
-#'
 #' datasets <- init_filtered_data(
 #'   x = list(
 #'     iris = list(dataset = iris),
-#'     mae = list(dataset = miniACC)
+#'     mtcars = list(dataset = mtcars)
 #'   )
 #' )
 #' fs <- teal_slices(
 #'   teal_slice(dataname = "iris", varname = "Species", selected = c("setosa", "versicolor")),
 #'   teal_slice(dataname = "iris", varname = "Sepal.Length", selected = c(5.1, 6.4)),
+#'   teal_slice(dataname = "mtcars", varname = "gear", selected = c(4, 5)),
+#'   teal_slice(dataname = "mtcars", varname = "carb", selected = c(4, 10))
+#' )
+#'
+#' # set initial filter state
+#' set_filter_state(datasets, filter = fs)
+#'
+#' # get filter state
+#' get_filter_state(datasets)
+#'
+#' # modify filter state
+#' set_filter_state(
+#'   datasets,
+#'   teal_slices(
+#'     teal_slice(dataname = "iris", varname = "Species", selected = "setosa", keep_na = TRUE)
+#'   )
+#' )
+#'
+#' # remove specific filters
+#' remove_filter_state(
+#'   datasets,
+#'   teal_slices(
+#'     teal_slice(dataname = "iris", varname = "Species"),
+#'     teal_slice(dataname = "mtcars", varname = "gear"),
+#'     teal_slice(dataname = "mtcars", varname = "carb")
+#'   )
+#' )
+#'
+#' # remove all states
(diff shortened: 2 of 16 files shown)


R/FilteredDataset-utils.R
```diff
@@ -40,6 +40,7 @@
 #'   shinyApp(ui, server)
 #' }
 #'
+#' @examplesIf requireNamespace("MultiAssayExperiment")
```
pawelru, line 43: nice 👍 good call to split and make conditional. Much better than big if inside.

pawelru:
I think this is the correct way to move forward.

I have one more thing to check that I noticed elsewhere and unfortunately I cannot check it locally for some reason. If a dependent package is missing and there are references in roxygen it will end-up with a following note:
```
* checking Rd cross-references ... NOTE
Unknown package ‘teal’ in Rd xrefs
```
Or it might be due to a given package is missing on rdrr.io / CRAN - I'm not 100% sure. I haven't deep dive into this issue but I think it might be related in this context. Will you be able to check it? Just remove package locally and run R CMD CHECK.

averissimo:
It looks like a `rdrr.io / CRAN` check. As I've modified the docs to also have some links to {future} and {mistery} (latter doesn't exist) and it only complained about {teal} and {mistery}

image

(ignore the `dev-av` warning, it's for local hacking)

R/FilteredDataset-utils.R
m7pr: wow @examplesIf is a powerful feature. great use with `requireNamespace`

R/filter_panel_api.R
```diff
@@ -23,8 +23,12 @@
...
 #'
 #' @examples
-#' library(MultiAssayExperiment)
-#' data(miniACC)
+#' if (requireNamespace("MultiAssayExperiment", quietly = TRUE)) {
+#'   library(MultiAssayExperiment)
+#'   data(miniACC)
+#' } else {
+#'   miniACC <- data.frame()
+#' }
```
chlebowa, line 31: While I don't see a better way to handle this, I don't like it. Sure, the examples will work properly, but we pass a `data.frame` instead of a `MultiAssayExperiment`, that's quite a difference.

How about we limit this example to `data.frames`, maybe add `mtcars` for the sake of having two datasets? The downside then becomes we don't have a public-facing example involving an MAE. Perhaps move the filter specification for `miniACC` to the `teal_slice` docs?

chlebowa:
So I see that `@examplesIf` prevents the examples from running in checks unless the condition is met but they are still displayed. Sounds good to me.

> ```
> * checking Rd cross-references ... NOTE
> Unknown package ‘teal’ in Rd xrefs
> ```

I'm not getting this, only 
```
❯ checking package dependencies ... NOTE
  Package suggested but not available for checking: ‘MultiAssayExperiment’
```

----

To summarize:
`examplesIf` :+1: 
remove `withr` :+1:

R/filter_panel_api.R
averissimo: The implementation seemed like the lesser of evils. Keeping it as it is makes this example quite complex, but compact.

I do like your suggestion, we can also add `mtcars` and a new section just for MAE in the example. Keeping in line the "optional" aspect of this requirement, as well as prepare this to a possible [future break](https://github.com/insightsengineering/teal.slice/issues/500)

R/filter_panel_api.R
```diff
@@ -23,22 +23,53 @@
...
+#'   teal_slices(
+#'     teal_slice(dataname = "iris", varname = "Species"),
+#'     teal_slice(dataname = "mtcars", varname = "gear"),
+#'     teal_slice(dataname = "mtcars", varname = "carb")
+#'   )
+#' )
+#'
+#' # remove all states
+#' clear_filter_states(datasets)
+#' @examplesIf requireNamespace("MultiAssayExperiment")
```
averissimo, line 65: Replaces {mae} with mtcars in main example, adds a secondary one with only {mae}

chlebowa: I endorse this :+1:

chlebowa:
I suppose we will do this in all package, not just `teal.slice`, correct? 

I have uninstalled `MultiAssayExperiment` to test this and now tests in `teal` fail, so we don't have to look far.

chlebowa (approved):
:+1:

pawelru (approved):

PR #520, opened 2024-01-12
516 `topological_sort`

### LINKED ISSUE #516: internal function `topological_sort()` is defined in `teal.slice` and `teal.data`
### What happened?

The` topological_sort` function is defined in both the [teal.slice](https://github.com/insightsengineering/teal.slice/blob/ea691771d192535bbacfa76117576e63a3332df0/R/FilteredData-utils.R#L194-L242) and [teal.data](https://github.com/insightsengineering/teal.data/blob/79d37932d21b52e21d69681de82ae156e55d307f/R/topological_sort.R#L12-L60) packages, sharing identical logic.

Should we utilize it in `teal.slice` by employing the` getFromNamespace()` function (pending feedback from CRAN for `teal.code`)? Alternatively, do we consider exporting this function in teal.data?

chlebowa:
Fixes #516 

Replaced function definition for `topological_sort` with wrapper for the same function defined in `teal.data`.
The new definition remains in `FilteredData-utils.R` to maintain documentation (would be too much to put i `zzz.R`).

### DIFF
--- a/R/FilteredData-utils.R
+++ b/R/FilteredData-utils.R
@@ -142,48 +142,5 @@ toggle_title <- function(input_id, titles, one_way = FALSE) {
 #' @seealso examples found here: `vignette("internal_function_examples", package = "teal.slice")`.
 #' @keywords internal
 topological_sort <- function(graph) {
-  # compute in-degrees
-  in_degrees <- list()
-  for (node in names(graph)) {
-    in_degrees[[node]] <- 0
-    for (to_edge in graph[[node]]) {
-      in_degrees[[to_edge]] <- 0
-    }
-  }
-
-  for (node in graph) {
-    for (to_edge in node) {
-      in_degrees[[to_edge]] <- in_degrees[[to_edge]] + 1
-    }
-  }
-
-  # sort
-  visited <- 0
-  sorted <- list()
-  zero_in <- list()
-  for (node in names(in_degrees)) {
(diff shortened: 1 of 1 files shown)


R/FilteredData-utils.R
```diff
@@ -142,48 +142,5 @@ toggle_title <- function(input_id, titles, one_way = FALSE) {
...
-  if (visited != length(in_degrees)) {
-    stop(
-      "Graph is not a directed acyclic graph. Cycles involving nodes: ",
-      paste0(setdiff(names(in_degrees), sorted), collapse = " ")
-    )
-  } else {
-    return(sorted)
-  }
+  utils::getFromNamespace("topological_sort", ns = "teal.data")(graph)
 }
```
gogonzo, line 146: Isn't this enough?
```suggestion
topological_sort <- utils::getFromNamespace("topological_sort", ns = "teal.data")
```

chlebowa: Technically yes, but there are two differences.
1. If you assign the bare call, then calling `topological_sort` will print the actual function body but with `namespace: teal.data`.
2. If you assign a wrapper, you get `namespace: teal.slice` but `topological_sort` will only print the wrapper.

Which do you prefer? I lean towards (1) myself and that is the one we have been using so far.

gogonzo: I don't have any preferences here. Wanted to know

gogonzo (approved):

Issue #533, opened 2024-01-26
[UX] Behaviour of the Interactive filters in the Filter Panel


vedhav:
Two things here:

1. The Interactive filters reload the UI when the checkbox input is changed which brings a bad UX.
~2. Is there a reason to create multiple checkbox inputs for the same filter?~

https://github.com/insightsengineering/teal.slice/assets/49812166/d249f4a1-32ec-44dc-9df7-202dabdc69a5

App code used for the test:
```r
library(teal)
data <- teal_data() |>
  within({
    set.seed(123)
    size <- 100
    DATA <- data.frame(
      id = seq(size),
      numeric = round(rnorm(size, 25, 10)),
      logical = sample(c(TRUE, FALSE), size = size, replace = TRUE),
      factor = factor(sample(LETTERS[1:4], size = size, replace = TRUE)),
      character = sample(letters, size = size, replace = TRUE),
      datetime = sample(seq(as.POSIXct('2000/01/01'), as.POSIXct('2024/01/01'), by="day"), size = size)
    )
    DATA$date <- as.Date(DATA$datetime)
  })
datanames(data) <- "DATA"
app <- teal::init(
  data = data,
  modules = modules(example_module()),
  filter = teal_slices(
    teal_slice("DATA", "numeric"),
    teal_slice("DATA", "logical"),
    teal_slice("DATA", "factor"),
    teal_slice("DATA", "character"),
    teal_slice("DATA", "datetime"),
    teal_slice("DATA", "date")
  )
)
shinyApp(app$ui, app$server)
```

vedhav:
1. I mean that after selecting/unselecting from the checkbox (example in the video) the whole UI is refreshed. The UI should not be dependant on this right?

2. This one is clear. Thanks

chlebowa:
You mean the available filters popover?

vedhav:
Yes

chlebowa:
This has been stale for over a month, I suggest we close this.

vedhav:
Agreed, I can't think of a better solution without involving JS. And I think it will be an overkill for just this. There is also an issue about improving the hover options. Would be nice to revisit this when we work on that.

PR #534, opened 2024-01-26
Improve the UX of Interactive filters

vedhav:
Partially addresses #533


chlebowa:
You killed the available filters list, I'm afraid.

https://github.com/insightsengineering/teal.slice/assets/114988527/90a734b2-2927-4788-946b-64fdb6ae3342

vedhav:
Okay, I see why it was implemented this way. Let me dig deeper.

chlebowa (changes requested):
blocking just in case

PR #536, opened 2024-01-26
options for strict tests; few enhancements

pawelru:
- part of https://github.com/insightsengineering/coredev-tasks/issues/498
- part of https://github.com/insightsengineering/coredev-tasks/issues/478
  - please read this for more info about the implementation: https://github.com/insightsengineering/coredev-tasks/issues/478#issuecomment-1909912778


tests/testthat/setup-options.R
```diff
@@ -0,0 +1,17 @@
+opts_partial_match_old <- list(
+  warnPartialMatchDollar = getOption("warnPartialMatchDollar"),
+  warnPartialMatchArgs = getOption("warnPartialMatchArgs"),
+  warnPartialMatchAttr = getOption("warnPartialMatchAttr")
+)
+opts_partial_match_new <- list(
+  warnPartialMatchDollar = TRUE,
```
averissimo, line 7: Should we also set `warn = 1`?

It might help to debug when more than 10 are presented and it was on Hadley's original comment on {rcmdcheck} and {devtools}

`?options`
> If warn is zero (the default) warnings are stored until the top–level function returns. If 10 or fewer warnings were signalled they will be printed otherwise a message saying how many were signalled...
> If warn is one, warnings are printed as they occur. 

```suggestion
  warn = 1,
  warnPartialMatchDollar = TRUE,
```

tests/testthat/setup-options.R
```diff
@@ -0,0 +1,17 @@
+opts_partial_match_old <- list(
+  warnPartialMatchDollar = getOption("warnPartialMatchDollar"),
+  warnPartialMatchArgs = getOption("warnPartialMatchArgs"),
+  warnPartialMatchAttr = getOption("warnPartialMatchAttr")
+)
```
averissimo, line 5: Do we need to keep the old options? They are never restored as `withr` handles that on its own.

averissimo (commented):
I like these changes, added very minor comments.

Great catch on `regex = ...` typo :+1:

tests/testthat/setup-options.R
pawelru: Can you please elaborate more what's the difference?

I was experimenting with `warn = 2` and abandoned this idea to keep this as warning. This has also undesirable and hard to control side-effects.

Currently we should have the following (temporarily re-introduced the old error):
```
$> devtools::test()
ℹ Testing teal.slice
✔ | F W  S  OK | Context
✔ |          9 | calls_combine_by                                                                                                                                                  
(...)                                                                                                                                  
✔ |   1     16 | FilteredDataset                                                                                                                                                   
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Warning (test-FilteredDataset.R:138:3): ui_add is pure virtual
partial argument match of 'rege' to 'regexp'
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
✔ |         39 | FilterState                                                                                                                                                       
(...)             
```

With `warn = 1` it's the same. So I don't see any difference - at least from that perspective.

tests/testthat/setup-options.R
pawelru: This was left for possible exclusions when I was mass copy&paste. This package (among others) does not have exclusions so this is actually redundant (contrary to e.g. [tern](https://github.com/insightsengineering/tern/pull/1182)). I honestly don't know. I'm fine with removing this from here.

tests/testthat/setup-options.R
averissimo: You're right, {testthat} catches all of them, I probably had a bug `1:10` as I was seeing fewer warnings.

```r
testthat::test_that("warnings", {
  (\() {
    lapply(1:100, \(.x) warning(glue::glue("warning {.x}")))
    invisible()
  })()
})
```

tests/testthat/setup-options.R
averissimo: Ok, it seems fine to leave, however a comment giving some context on how to use might be nice to have.

averissimo:
```suggestion
# May be used during tests to revert to non-strict environment
opts_partial_match_old <- list(
  warnPartialMatchDollar = getOption("warnPartialMatchDollar"),
  warnPartialMatchArgs = getOption("warnPartialMatchArgs"),
  warnPartialMatchAttr = getOption("warnPartialMatchAttr")
)
```

averissimo (approved):
LGTM :guitar: 

One _(optional)_ minor comment about adding context to an unused variable.

Feel free to resolve that as you see fit.

R/FilterState.R
```diff
@@ -149,7 +149,7 @@ FilterState <- R6::R6Class( # nolint
     set_state = function(state) {
       checkmate::assert_class(state, "teal_slice")
       if (private$is_fixed()) {
-        logger::log_warn("attempt to set state on fixed filter aborted id: { private$get_id() }")
+        warning(sprintf("attempt to set state on fixed filter aborted id: %s", private$get_id()))
```
chlebowa, line 152: `warning` doesn't need `sprintf`
```suggestion
        warning("attempt to set state on fixed filter aborted id: ", private$get_id())
```

tests/testthat/test-FilterStates.R
```diff
@@ -43,10 +43,10 @@ testthat::test_that("set_filter_state sets count_type", {
 
 testthat::test_that("set_filter_state ignores teal_slice for inexisting variables with log warning", {
   filter_states <- FilterStates$new(data = data.frame(a = 1), dataname = "test")
-  res <- utils::capture.output(
-    filter_states$set_filter_state(teal_slices(teal_slice(dataname = "test", varname = "inexisting")))
+  testthat::expect_warning(
+    filter_states$set_filter_state(teal_slices(teal_slice(dataname = "test", varname = "inexisting"))),
+    "inexisting excluded from test"
```
chlebowa, line 48: Please replace "inexisting" with "non-existent" in the four instances throughout the package.
```suggestion
    "non-existent excluded from test"
```

tests/testthat/setup-options.R
```diff
@@ -0,0 +1,20 @@
+# `opts_partial_match_old` is left for exclusions due to partial matching in dependent packages (i.e. not fixable here)
+# it might happen that it is not used right now, but it is left for possible future use
+# use with: `withr::with_options(opts_partial_match_old, { ... })` inside the test
```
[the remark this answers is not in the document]
averissimo: Awesome! Thanks

donyunardi:
Let's merge this after CRAN release.

pawelru:
CRAN release already happened - merging this

PR #546, opened 2024-01-31
Moves `{shiny}` to imports and updates examples/vignettes

### LINKED ISSUE #529: Move `shiny` to Imports
`Depends` type of dependency is generally discouraged. Would it be possible to move this to `Imports` instead?

averissimo:
# Pull Request

- Fixes #529

ℹ️ Note: The package imports all functions from `{shiny}` via `@import shiny` {roxygen2} tag, however, there's a mixed bag of prefixed (`shiny::fun(...)`) and non-prefixed calls (`fun(...)`) in the (functions') code. ~~Those were not modified.~~ (see discussion)

#### Changes description

- Remove occurences of `shiny::` from `R/` code
- Moved `{shiny}` to Imports
- Replaced shiny functions in examples by
  - add `library(shiny)` when building a shiny app to run
  - ~~`fun` -> `shiny::fun` for others~~
- Corrects vignettes accordingly

### DIFF
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -64,7 +64,7 @@
 #' ### set_filter_state
 #' library(shiny)
 #'
-#' utils::data(miniACC, package = "MultiAssayExperiment")
+#' data(miniACC, package = "MultiAssayExperiment")
 #' datasets <- FilteredData$new(list(iris = iris, mae = miniACC))
 #' fs <- teal_slices(
 #'   teal_slice(
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -40,6 +40,8 @@
 #' # use non-exported function from teal.slice
 #' FilteredData <- getFromNamespace("FilteredData", "teal.slice")
 #'
+#' library(shiny)
+#'
 #' datasets <- FilteredData$new(list(iris = iris, mtcars = mtcars))
 #'
 #' # get datanames
@@ -48,18 +50,19 @@
 #' datasets$set_filter_state(
 #'   teal_slices(teal_slice(dataname = "iris", varname = "Species", selected = "virginica"))
 #' )
-#' shiny::isolate(datasets$get_call("iris"))
+#' isolate(datasets$get_call("iris"))
 #'
 #' datasets$set_filter_state(
 #'   teal_slices(teal_slice(dataname = "mtcars", varname = "mpg", selected = c(15, 20)))
 #' )
 #'
-#' shiny::isolate(datasets$get_filter_state())
-#' shiny::isolate(datasets$get_call("iris"))
-#' shiny::isolate(datasets$get_call("mtcars"))
+#' isolate(datasets$get_filter_state())
+#' isolate(datasets$get_call("iris"))
+#' isolate(datasets$get_call("mtcars"))
 #'
 #' @examplesIf requireNamespace("MultiAssayExperiment")
 #' ### set_filter_state
+#' library(shiny)
 #'
 #' utils::data(miniACC, package = "MultiAssayExperiment")
 #' datasets <- FilteredData$new(list(iris = iris, mae = miniACC))
@@ -84,7 +87,7 @@
 #'   )
 #' )
 #' datasets$set_filter_state(state = fs)
-#' shiny::isolate(datasets$get_filter_state())
+#' isolate(datasets$get_filter_state())
 #'
 #' @keywords internal
 #'
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -48,15 +48,15 @@
 #' datasets$set_filter_state(
 #'   teal_slices(teal_slice(dataname = "iris", varname = "Species", selected = "virginica"))
 #' )
-#' isolate(datasets$get_call("iris"))
+#' shiny::isolate(datasets$get_call("iris"))
 #'
 #' datasets$set_filter_state(
 #'   teal_slices(teal_slice(dataname = "mtcars", varname = "mpg", selected = c(15, 20)))
 #' )
 #'
-#' isolate(datasets$get_filter_state())
-#' isolate(datasets$get_call("iris"))
-#' isolate(datasets$get_call("mtcars"))
+#' shiny::isolate(datasets$get_filter_state())
+#' shiny::isolate(datasets$get_call("iris"))
+#' shiny::isolate(datasets$get_call("mtcars"))
 #'
 #' @examplesIf requireNamespace("MultiAssayExperiment")
 #' ### set_filter_state
@@ -84,7 +84,7 @@
 #'   )
 #' )
 #' datasets$set_filter_state(state = fs)
-#' isolate(datasets$get_filter_state())
+#' shiny::isolate(datasets$get_filter_state())
 #'
 #' @keywords internal
 #'
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -395,7 +395,7 @@ FilteredData <- R6::R6Class( # nolint
     #' @param ... additional arguments passed to `format`.
     #'
     print = function(...) {
-      cat(shiny::isolate(self$format(...)), "\n")
+      cat(isolate(self$format(...)), "\n")
     },
 
     #' @description
@@ -405,7 +405,7 @@ FilteredData <- R6::R6Class( # nolint
     #'
     #' @return `NULL`, invisibly.
     set_filter_state = function(state) {
-      shiny::isolate({
+      isolate({
         logger::log_trace("{ class(self)[1] }$set_filter_state initializing")
         checkmate::assert_class(state, "teal_slices")
         allow_add <- attr(state, "allow_add")
@@ -434,7 +434,7 @@ FilteredData <- R6::R6Class( # nolint
     #' @return `NULL`, invisibly.
     #'
     remove_filter_state = function(state) {
-      shiny::isolate({
+      isolate({
         checkmate::assert_class(state, "teal_slices")
         datanames <- unique(vapply(state, "[[", character(1L), "dataname"))
         checkmate::assert_subset(datanames, self$datanames())
@@ -604,7 +604,7 @@ FilteredData <- R6::R6Class( # nolint
     #' @return `NULL`.
     srv_active = function(id, active_datanames = self$datanames) {
       checkmate::assert_function(active_datanames)
-      shiny::moduleServer(id, function(input, output, session) {
+      moduleServer(id, function(input, output, session) {
         logger::log_trace("FilteredData$srv_active initializing")
 
         private$srv_available_filters("available_filters")
@@ -644,9 +644,9 @@ FilteredData <- R6::R6Class( # nolint
           }
         )
 
-        output$teal_filters_count <- shiny::renderText({
+        output$teal_filters_count <- renderText({
           n_filters_active <- private$get_filter_count()
-          shiny::req(n_filters_active > 0L)
+          req(n_filters_active > 0L)
           sprintf(
             "%s filter%s applied across datasets",
             n_filters_active,
@@ -717,7 +717,7 @@ FilteredData <- R6::R6Class( # nolint
       checkmate::assert_class(active_datanames, "reactive")
       moduleServer(id, function(input, output, session) {
         logger::log_trace("FilteredData$srv_add initializing")
-        shiny::observeEvent(input$minimise_filter_add_vars, {
+        observeEvent(input$minimise_filter_add_vars, {
           shinyjs::toggle("filter_add_vars_contents")
           toggle_icon(session$ns("minimise_filter_add_vars"), c("fa-angle-right", "fa-angle-down"))
           toggle_title(session$ns("minimise_filter_add_vars"), c("Restore panel", "Minimise Panel"))
@@ -809,7 +809,7 @@ FilteredData <- R6::R6Class( # nolint
         function(input, output, session) {
           logger::log_trace("FilteredData$srv_filter_overview initializing")
 
-          shiny::observeEvent(input$minimise_filter_overview, {
+          observeEvent(input$minimise_filter_overview, {
             shinyjs::toggle("filters_overview_contents")
             toggle_icon(session$ns("minimise_filter_overview"), c("fa-angle-right", "fa-angle-down"))
             toggle_title(session$ns("minimise_filter_overview"), c("Restore panel", "Minimise Panel"))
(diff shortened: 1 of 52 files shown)


averissimo:
> Replaced shiny functions in examples by
> * add `library(shiny)` when building a shiny app to run
> * `fun` -> `shiny::fun` for others

I can easily use `library(shiny)` for all if that convention is better

chlebowa:
Since the entire `shiny` namespace is imported, I believe we should
- _not_ prefix `shiny` functions in code
- keep prefixes in tests
- add `library(shiny)` in all exapmles that require it

My reasons are:
1. We have been removing namespace prefixes from examples lately, especially in `teal`.
2. Attaching other packages in examples is common practice.

averissimo:
Ok, makes sense I'll make those changes. Hold on on the review then :smile:

averissimo:
Heads up that the double colon change will create a huge PR to review, what do you think about splitting these in 2?

1. examples/vignettes/DESCRIPTION
2. removing double colon calls to shiny

averissimo:
By huge it's 79 replacements :sweat_smile:

m7pr:
I don't think there is any reason for `removing double colon calls to shiny` in the code. We put `package::` prefix also to allow 
easier code debugging and improve code readability. Even if `shiny::` prefix is not needed when we import `shiny` into `NAMESAPCE` there are still other benefits of prefixing packages in code.

chlebowa:
> Removal of `shiny::` prefix for shiny only and leaving `package::` prefix for other packages would mix up two conventions.

Hardly. The namespace is imported. We import single functions in several places, e.g. `as`, `sertNames`. Would you have every `selectInput` prefixed? Every `div`?

m7pr:
Of course I don`t support putting `shiny::` before `selectInputs` and `divs`. After having seconds thoughts I also don't think we need to put `shiny::` prefix before `isolate` or `reactive` or `reactiveValues`. Maybe it's not that bad idea at all.

For the `NAMESPACE`, just a side note: if you have one function imported for a package, but you use two and both are not prefixed, you will get a warning for the second one that does not have a prefix and is not listed in the `NAMESPACE` that it can not be found. `NAMESPACE` in `teal.slice` doesn't have `importFrom(teal.logger, )` and that's why we need to prefix each `teal.logger` usage.

averissimo:
We could import only specific heavily used functions and a double colon for the rest.

There are around:
- 259 `div` calls
- 183 `tags$`
- 631 `isolate()`
- 215 `reactive()`

averissimo:
But that could be a hard line to define which are heavily used and those that are not.

chlebowa:
> For the `NAMESPACE`, just a side note: if you have one function imported for a package, but you use two and both are not prefixed, you will get a warning for the second one that does not have a prefix and is not listed in the `NAMESPACE` that it can not be found.

As expected.

> `NAMESPACE` in `teal.slice` doesn't have `importFrom(teal.logger, )` and that's why we need to prefix each `teal.logger` usage.

That much is obvious, what's your point?

m7pr:
> We could import only specific heavily used functions and a double colon for the rest.

That's cleaner than `@import shiny` (everything). I think those used more than 50 times could be a heavily used functions

chlebowa:
> That much is obvious, what's your point?

Just saying our NAMESPACE file is not filled for the convention where we do not prefix anything. That's all. A side-side-side note :)

`@import shiny` allows you to use all functions in the package code with no prefixes.> > We could import only specific heavily used functions and a double colon for the rest.

> > We could import only specific heavily used functions and a double colon for the rest.
> 
> That's cleaner than `@import shiny` (everything). I think those used more than 50 times could be a heavily used function

I disagree. It is cleaner to have consistent code. 50 is an arbitrary number, why not 47? And if I add or remove a usage in the future, do I have to tally and change accordingly? Ridiculous.

averissimo:
> > We could import only specific heavily used functions and a double colon for the rest.
> 
> That's cleaner than `@import shiny` (everything). I think those used more than 50 times could be a heavily used function

I would tentatively agree, the caveat is that it would create a \***long**\* `NAMESPACE` file which is kind of a deal breaker.

Re: this PR

How about:
1. Remove double colon on this PR as it's "only" 79 occurrences of (mostly) `isolate`, `reactive*`, `render*` & co
2. Continue the discussion on the original issue

m7pr:
> I disagree. It is cleaner to have consistent code. 50 is an arbitrary number, why not 47? And if I add or remove a usage in the future, do I have to tally and change accordingly? Ridiculous.

World is not ideal and sometimes you need to come up with an arbitrary number to make things easier. In business it's so-called BUSINESS VALUE based on BUSINESS EXPERIENCE : D and that sometimes needs to be revisited, as all that we do: we do revisit examples and documentation, and we do revisit NAMESPACE and DESCRIPTION.

I would actually revisit `@import package function` / `importFrom(package, )` [those functions treated as heavily used] as often as we do revisit DependsVSImports for a package. 

> 50 is an arbitrary number, why not 47

Rounded numbers are easier to remember. Do we any any more motivation for 50 being better than 47 :D?

m7pr:
This BUSINESS VALUE was a joke (added `: D`).

> And if I add or remove a usage in the future, do I have to tally and change accordingly?

I think seasonal revisit should be enough. And I think that's what we do with our revisits.

averissimo:
PR is now ready for review :smile: 

Removed all double colon in `R/` code and added a few missing from `tests/`

R/include_css_js.R
```diff
@@ -13,5 +13,5 @@ include_css_files <- function(pattern = "*") {
     system.file("css", package = "teal.slice", mustWork = TRUE),
     pattern = pattern, full.names = TRUE
   )
-  return(shiny::singleton(shiny::tags$head(lapply(css_files, shiny::includeCSS))))
+  return(singleton(tags$head(lapply(css_files, includeCSS))))
```
chlebowa, line 16: NEST policy
```suggestion
  singleton(tags$head(lapply(css_files, includeCSS)))
```

R/count_labels.R
```diff
@@ -36,6 +36,7 @@
 #'   countsnow = unname(counts)
 #' )
 #'
+#' library(shiny)
```
chlebowa, line 39: In this example packages are attached as they are needed. In others all packages are attached at the beginning of examples. 
I understand that both ways can be appropriate. Would you mind making sure that it makes sense in all cases?

R/FilteredData.R
```diff
@@ -60,6 +62,7 @@
 #'
 #' @examplesIf requireNamespace("MultiAssayExperiment")
 #' ### set_filter_state
+#' library(shiny)
 #'
 #' utils::data(miniACC, package = "MultiAssayExperiment")
```
chlebowa, line 67: Do we need a `utils::` in an example?

chlebowa (commented):
So far so good!

In some examples `isolate` be replaced with `reactiveConsole(TRUE)`. What do you think?

chlebowa:
> > not prefix shiny functions in code
> 
> Hmmm... `shiny` is on the same level as `checkmate` (and any other package in Imports) - why don't we prefix `shiny` calls and we do prefix `checkmate` (and all the other imported pkgs)? I always look for consistency in our codebase and for this reason I think we should do prefix inside the codebase.

`checkmate` is in `Imports` but we don't `@import checkmate`.

An argument can be made for `R6` but we can defer that because it does not influence the dependency structure between our packages.

R/FilteredData.R
averissimo: Missed that one as I removed the others. Good catch.

R/count_labels.R
averissimo: They are all attached as needed, not necessary before the first call though.

The reasoning was to either:
- place it before the set of expressions that use it
- or before the ShinyApp is built (this case)

image

This example is ambiguous though, I'll move it to the top just in case

chlebowa: Good enough for me :+1:

pawelru:
> > > not prefix shiny functions in code
> > 
> > 
> > Hmmm... `shiny` is on the same level as `checkmate` (and any other package in Imports) - why don't we prefix `shiny` calls and we do prefix `checkmate` (and all the other imported pkgs)? I always look for consistency in our codebase and for this reason I think we should do prefix inside the codebase.
> 
> `checkmate` is in `Imports` but we don't `@import checkmate`.
> 
> An argument can be made for `R6` but we can defer that because it does is not connected to the dependency structure between our packages.

So `#' @import shiny` is done to make shiny function available without prefix. I feel it's more a direct consequence and not the reason of the decision of not to prefix. The question is still - why don't we prefix (and have this import statement as a consequence)? (note we can still prefix and have import)
(a side note, reading this now - [this](https://roxygen2.r-lib.org/articles/namespace.html#:~:text=not%20generally%20recommended))

averissimo:
> In some examples `isolate` can be replaced with `reactiveConsole(TRUE)`. What do you think?

On one hand, it might induce the developer into thinking that it doesn't need reactivity, on the other many of these examples will be used within a reactive environment anyway, so it's a realistic snippet.

I'm not a fan of the description in `?reactiveConsole` though.

> This is an experimental feature that allows you to enable reactivity at the console, for the purposes of experimentation and learning.

I leaning toward keeping it as it is, but if you think it would be better, then I won't oppose

chlebowa:
> So `#' @import shiny` is done to make shiny function available without prefix.

*_all_ functions

> why don't we prefix

Because there are too many. That's how I see it, at least.

m7pr:
`reactiveConsole(TRUE)` call will change user's session settings : P maybe that's a point for not using that in examples?
unless we end them with `reactiveConsole(FALSE)`

chlebowa:
> `reactiveConsole(TRUE)` call will change user's session settings : P maybe that's a point for not using that in examples? unless we end them with `reactiveConsole(FALSE)`

That was the implication, yes.


> I'm not a fan of the description in `?reactiveConsole` though.

Fair enough, let's stick to `isolate`.

chlebowa (approved):
Looks good to me :+1: 
Dependency was successfully softened.
Imported namespace/function usage is consistent.

pawelru:
> > why don't we prefix
> 
> Because there are too many. That's how I see it, at least.

That's okey. I guess we all expected this. I mentioned earlier that this can be done in two step approach if that's too much for the tight deadlines.
In general I feel that prefixing is much more safer way of writing the code. I like the reasoning I have found (lined above): while this is okey right now, it does not guarantee that in the future packages will export objects with the same name and there will be name collision. Prefixing is a safeguard for such situation.

Coming back to this PR
All the changes are good but removing prefixes from the codebase feels like a adding more into a lot of work anyway. I suggest to revert those.

chlebowa:
@pawelru By that rationale we _should_ prefix every `div` and every `observeEvent`. Is this what you want?

pawelru:
> @pawelru By that rationale we _should_ prefix every `div` and every `observeEvent`. Is this what you want?

Yessir. For the same reason we have (had?) prefixed `isolate()`, `reactive()` etc.
Consistency - I'm looking for consistency. Regardless if it's 3 or 33 occurrences.

chlebowa:
> > @pawelru By that rationale we _should_ prefix every `div` and every `observeEvent`. Is this what you want?
> 
> Yessir. 

I vehemently oppose this.

> For the same reason we have (had?) prefixed `isolate()`, `reactive()` etc. Consistency - I'm looking for consistency. 

We do have consistency now (with this PR). Import nothing - prefix, import one function - don't prefix that, import whole namepsace - prefix nothing.

> Regardless if it's 3 or 33 occurrences.

I don't know what this number means.

pawelru:
> > > @pawelru By that rationale we _should_ prefix every `div` and every `observeEvent`. Is this what you want?
> > 
> > 
> > Yessir.
> 
> I vehemently oppose this.

So we have a disagreement then. Will I insist - not really. Not for this. This is a relatively small thing anyway. Will stick to what you guys agree on. But still I can't help having a feeling that removing prefixes is a step back in some sense and makes this pkg code more vulnerable for potential namespace collision.

> > For the same reason we have (had?) prefixed `isolate()`, `reactive()` etc. Consistency - I'm looking for consistency.
> 
> We do have consistency now (with this PR). Import nothing - prefix, import one function - don't prefix that, import whole namepsace - prefix nothing.

OK so there's a logic in this than. I missed this somehow.

> > Regardless if it's 3 or 33 occurrences.
> 
> I don't know what this number means.

These have no meaning. It's just to highlight that number of occurrences is irrelevant.

averissimo:
I favor the direction of using prefix on {teal.*} code. Shiny's namespace is just too big _(277 exports)_.

I revisited the "prefix" subject today when observing `shiny::code()` and `shiny::includeCSS` and `teal::include_css_files` on a PR _(getting a chill)_ and took another look at occurences

I found that I was grossly overestimating occurences on a comment before (by including vignettes and rstudio hidden folders on the `grep`).

We could open room for a hybrid approach that would import a select set of functions (UI and/or heavily used functions).

```
# {shiny} usage in {teal} (there could be false positives here `tags$div(` will appear 2x under `tags` and `div`)
# github hash used 1a9976664
26 tags
17 div
16 validate
15 need
14 observeEvent
13 moduleServer
13 icon
12 reactive
12 isolate
10 NS
10 is.reactive
# expand below to see all <10
```

<details>

<summary>Occurences of {shiny} namespace in teal</summary>

```bash
#!/bin/bash
if [ "$2" == "wc" ];then
  echo $(grep -RE "(($1\()|($1\\$))" R | grep -vE "(::${1})|([a-zA-Z0-9_.]$1)" | grep -vE ":#'" | wc -l) $1
else
  grep -RE "(($1\()|($1\\$))" R | grep -vE "(::${1})|([a-zA-Z0-9_.]$1)" | grep -vE ":#'" | grep -E --color "$1"
fi

# 1. Create "shiny_namespace.bash" with script above
# 2. create a "shiny_namespace.txt" with all functions to check"
# 3. run: $ for word in $(cat shiny_namespace.txt); do bash shiny_namespace.bash $word wc; done | grep -v "^0 " | sort -h -r
```

```r
26 tags
17 div
16 validate
15 need
14 observeEvent
13 moduleServer
13 icon
12 reactive
12 isolate
10 NS
10 is.reactive
7 showNotification
7 reactiveVal
7 p
6 tagList
5 singleton
5 showModal
5 modalDialog
5 modalButton
4 renderUI
4 removeModal
4 eventReactive
4 column
4 actionLink
4 actionButton
3 updateTextInput
3 uiOutput
3 reactiveValues
3 hr
2 textInput
2 span
2 fluidRow
1 verbatimTextOutput
1 updateSelectInput
1 textOutput
1 tagAppendChild
1 tabPanel
1 tableOutput
1 selectInput
1 req
1 renderText
1 renderTable
1 renderPrint
1 removeUI
1 reactiveValuesToList
1 pre
1 insertUI
1 h5
1 h2
1 getDefaultReactiveDomain
1 fluidPage
1 fileInput
1 downloadLink
1 downloadHandler
1 conditionalPanel
1 code
```
</details>

chlebowa:
Only 12 `reactive` and only one `req`? Only 14 `observeEvent`?? I can't believe it :raised_back_of_hand:

averissimo:
The script isn't perfect, but it's in the ballpark

image

PR #547, opened 2024-01-31
457 update filter card summary

### LINKED ISSUE #457: update filter card summary
The filter card summary put up by `ChoicesFilterState` is empty when no selection is made. As a result, the height of the card changes depending on the selection state. Perhaps there should be a note of some sort for no selection?

##### _current_
image

##### _proposed_
image

chlebowa:
Closes #457


m7pr:
Tested with the app from filter-panel-for-developers vignette

image

and now the height is the same if there is selection or no selection

image

m7pr (approved):
👍

PR #550, opened 2024-02-01
548 remove private/internal examples

### LINKED ISSUE #548: remove examples for non-exported functions
Examples for non-exported functions should be removed before the initial CRAN release and restored afterwards.
The Examples in question can be easily identified as they all contain `getFromNamespace` calls.

Remove them all in one commit (can be a squash commit of a PR) so that it can be easily reverted.

m7pr:
Closes #548


chlebowa (approved):
All offending examples have been removed.

Let's hold off merging until #536 goes in, shall we?

Also, since we haven't actually tested this approach to examples with yet (`teal.data` didn't have any), perhaps we should try sending to CRAN _with_ these examples?

m7pr:
Yes to both! I think we did send examples in this manner with `teal.code` so I don't understand why this is an issue again. Probably depending on the reviewer. I would try sending with those examples at the first attempt.

chlebowa:
> we did send examples in this manner with `teal.code`

But that was on _re_submission.

averissimo:
We want to submit this in a process that doesn't land on a human

> This information is (approximately) summarised in the following diagram by Hadley Wickham, available in the [cran-stages Github](https://github.com/edgararuiz/cran-stages) repository:

image

averissimo (approved):
Let's merge this to `main`

m7pr:
Merging to `release-candidate` branch, as we agreed during the discussion

PR #558, opened 2024-02-09
Title Case column labels in Active Filter Summary

pawelru:
Currently we have
image

### DIFF
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -881,11 +881,11 @@ FilteredData <- R6::R6Class( # nolint
               tagList(
                 lapply(
                   vapply(
-                    seq_along(datasets_df), 
+                    seq_along(datasets_df),
                     function(i) {
                       ifelse(!is.null(attr(datasets_df[[i]], "label")), attr(datasets_df[[i]], "label"), names(datasets_df)[[i]])
                     }
-                  ), 
+                  ),
                   tags$td
                 )
               )
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -843,16 +843,16 @@ FilteredData <- R6::R6Class( # nolint
               # some datasets (without keys) doesn't return subjects
               datasets_df <- transform(
                 datasets_df,
-                subjects_str_summary = ifelse(
+                subjects_summary = ifelse(
                   !is.na(subjects),
                   sprintf("%s/%s", subjects_filtered, subjects),
                   ""
                 )
               )
-              attr(datasets_df$subjects_str_summary, "label") <- "Subjects"
+              attr(datasets_df$subjects_summary, "label") <- "Subjects"
             }
 
-            all_names <- c("dataname", "obs_str_summary", "subjects_str_summary")
+            all_names <- c("dataname", "obs_str_summary", "subjects_summary")
             datasets_df <- datasets_df[, colnames(datasets_df) %in% all_names]
 
             body_html <- apply(
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -851,6 +851,7 @@ FilteredData <- R6::R6Class( # nolint
               )
               attr(datasets_df$subjects_str_summary, "label") <- "Subjects"
             }
+
             all_names <- c("dataname", "obs_str_summary", "subjects_str_summary")
             datasets_df <- datasets_df[, colnames(datasets_df) %in% all_names]
 
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -851,7 +851,8 @@ FilteredData <- R6::R6Class( # nolint
               )
               attr(datasets_df$subjects_str_summary, "label") <- "Subjects"
             }
-            datasets_df <- datasets_df[, colnames(datasets_df) %in% c("dataname", "obs_str_summary", "subjects_str_summary")]
+            all_names <- c("dataname", "obs_str_summary", "subjects_str_summary")
+            datasets_df <- datasets_df[, colnames(datasets_df) %in% all_names]
 
             body_html <- apply(
               datasets_df,
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -823,7 +823,7 @@ FilteredData <- R6::R6Class( # nolint
 
             datasets_df <- self$get_filter_overview(datanames = active_datanames())
 
-            attr(datasets_df$dataname, "label") <- "Dataname"
+            attr(datasets_df$dataname, "label") <- "Data Name"
 
             if (!is.null(datasets_df$obs)) {
               # some datasets (MAE colData) doesn't return obs column
--- a/R/FilteredData.R
+++ b/R/FilteredData.R
@@ -883,7 +883,11 @@ FilteredData <- R6::R6Class( # nolint
                   vapply(
                     seq_along(datasets_df),
                     function(i) {
-                      ifelse(!is.null(attr(datasets_df[[i]], "label")), attr(datasets_df[[i]], "label"), names(datasets_df)[[i]])
+                      ifelse(
+                        !is.null(attr(datasets_df[[i]], "label")),
+                        attr(datasets_df[[i]], "label"),
+                        names(datasets_df)[[i]]
+                      )
                     }
                   ),
                   tags$td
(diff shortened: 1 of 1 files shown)


R/FilteredData.R
```diff
@@ -876,6 +880,7 @@ FilteredData <- R6::R6Class( # nolint
             header_html <- tags$tr(
               tagList(
                 lapply(colnames(datasets_df), tags$td)
```
pawelru, line 882:
```suggestion
```

R/FilteredData.R
```diff
@@ -876,6 +880,7 @@ FilteredData <- R6::R6Class( # nolint
...
+                lapply(vapply(names(datasets_df), function(i) attr(datasets_df[[i]], "label") %||% i), tags$td)
```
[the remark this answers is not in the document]
pawelru: my godness, the alternative using base is so lengthy but I guess we just have to accept that unless rlang will be introduced or vbump R dep

R/FilteredData.R
```diff
@@ -823,31 +823,35 @@ FilteredData <- R6::R6Class( # nolint
...
+                subjects_str_summary = ifelse(
                   !is.na(subjects),
                   sprintf("%s/%s", subjects_filtered, subjects),
                   ""
                 )
               )
+              attr(datasets_df$subjects_str_summary, "label") <- "Subjects"
             }
-            datasets_df <- datasets_df[, colnames(datasets_df) %in% c("dataname", "Obs", "Subjects")]
+            datasets_df <- datasets_df[, colnames(datasets_df) %in% c("dataname", "obs_str_summary", "subjects_str_summary")]
```
pawelru, line 854:
```suggestion
            all_names <- c("dataname", "obs_str_summary", "subjects_str_summary")
            datasets_df <- datasets_df[, colnames(datasets_df) %in% all_names]
```

R/FilteredData.R
```diff
@@ -823,31 +823,36 @@ FilteredData <- R6::R6Class( # nolint
...
+                subjects_str_summary = ifelse(
                   !is.na(subjects),
                   sprintf("%s/%s", subjects_filtered, subjects),
                   ""
                 )
               )
+              attr(datasets_df$subjects_str_summary, "label") <- "Subjects"
             }
-            datasets_df <- datasets_df[, colnames(datasets_df) %in% c("dataname", "Obs", "Subjects")]
+            all_names <- c("dataname", "obs_str_summary", "subjects_str_summary")
```
pawelru, line 854:
```suggestion

            all_names <- c("dataname", "obs_str_summary", "subjects_str_summary")
```

R/FilteredData.R
```diff
@@ -875,7 +881,19 @@ FilteredData <- R6::R6Class( # nolint
 
             header_html <- tags$tr(
               tagList(
-                lapply(colnames(datasets_df), tags$td)
+                lapply(
+                  vapply(
+                    seq_along(datasets_df),
```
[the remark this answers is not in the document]
m7pr: In a for loop we could also add a warning, if `label` is missing.

R/FilteredData.R
```diff
@@ -823,31 +823,37 @@ FilteredData <- R6::R6Class( # nolint
...
               datasets_df <- transform(
                 datasets_df,
-                Subjects = ifelse(
+                subjects_str_summary = ifelse(
                   !is.na(subjects),
                   sprintf("%s/%s", subjects_filtered, subjects),
                   ""
                 )
               )
+              attr(datasets_df$subjects_str_summary, "label") <- "Subjects"
```
m7pr, line 852: Tot a fun of this long name

R/FilteredData.R
```diff
@@ -823,31 +823,37 @@ FilteredData <- R6::R6Class( # nolint
 
             datasets_df <- self$get_filter_overview(datanames = active_datanames())
 
+            attr(datasets_df$dataname, "label") <- "Dataname"
```
m7pr, line 826: ?
```suggestion
            attr(datasets_df$dataname, "label") <- "Data name"
```

m7pr:
@pawelru I left a few minor comments

R/FilteredData.R
```diff
@@ -875,7 +881,19 @@ FilteredData <- R6::R6Class( # nolint
...
               tagList(
-                lapply(colnames(datasets_df), tags$td)
+                lapply(
+                  vapply(
+                    seq_along(datasets_df),
+                    function(i) {
+                      ifelse(
+                        !is.null(attr(datasets_df[[i]], "label")),
+                        attr(datasets_df[[i]], "label"),
+                        names(datasets_df)[[i]]
```
m7pr, line 891: Anyway you can substitute `names(datasets_df)[[i]]` with `names(datasets_df[[i]])` ? Then you can assign

`ds <- datasets_df[[i]]` and have shorter usage of this part later in the code when taking `names(ds)` or `attr(ds)`?

R/FilteredData.R
pawelru: `"Data name"` OR `"Data Name"`?

R/FilteredData.R
pawelru: thanks, all valid points.

Let me reorg this a little. I was initially doing this from the webbrowser - never again...

pawelru: actually this comment is incorrect. I cannot do `names(df[[i]])`

```
r$> i <- 1

r$> names(iris)[[1]]
[1] "Sepal.Length"

r$> names(iris[[1]])
NULL
```

pawelru: this is the main reason for looping through the integers - to be able to call `names(df)[[i]]`

R/FilteredData.R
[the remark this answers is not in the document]
pawelru: re: integers
I replied in the other comment. I need it to do `names(df)[[i]]`.

re: for vs vapply
In this case, I have found `vapply` more handy because on each iteration I am appending to the result vector.
```
res <- character(0)
for (i in x) {
  res_i <- foo(i)
  res <- append(res, res_i)
}
```
vs
```
vapply(
  x
  \(i) foo(i)
)
```


re: warning
I have been thinking about it but I decided to stick to the current behaviour and focus on fixing the error and not to enhance the logic. I didn't want to make this PR big (even though I did it few times but this is nothing good IMHO)

R/FilteredData.R
m7pr: I wasn't sure about that. I think for the names it would need to be `iris[1]`. So ignore this one

R/FilteredData.R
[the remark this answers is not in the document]
m7pr: In `for` loop you can bind/concatenate/append with `c` function.
And I meant one `for` instead of a joint usage of `lapply(vapply)`. 

Instead of

```r
lapply(vapply(x, fun1), fun2)
```

do

```r
for(i in x)
x_res <- c(x_res, fun2(fun1(i))
```

or maybe you could do 

```r
vapply(x, function(i) fun2(fun1(i))
```

What do you think?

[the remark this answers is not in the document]
pawelru: Regardless of the function used - `vapply` gives this to you out-of-box.  If I don't need to capture the result for each iteration (e.g. loop to trigger side-effects) then I would probably choose `for` loop instead. But here I do, that's why I think it's handy.

`lapply(vapply(...))` was introduced to increase the readability - I separated label derivation out of html. Two loops instead of one but not nested - still of linear complexity (2O(n)).

Both are good and fully functional. It's just a balance between readability vs complexity. I just wanted this to be easily readable. Maybe I got to far? I don't know :)

R/FilteredData.R
pawelru: renamed to `subjects_summary`

R/FilteredData.R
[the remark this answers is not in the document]
m7pr: Ok, I think I'm convinced with the final output. `vapply` over `seq_along` gives you the benefit of concatenated output that would be a vector (not a list in comparison to lapply). In the end I see we do not use `lapply` for `html` derivation so there's no gonna be any `lapply(vapply`. Also the `ifelse` statement is simplfied. Thanks for the extra work. Closing

m7pr (approved):
@pawelru there is one last comment about `Data Name` usage. Other than that this is ready to go

R/FilteredData.R
pawelru:
```suggestion
            attr(datasets_df$dataname, "label") <- "Data Name"
```

PR #562, opened 2024-03-04
209 modify filter card css when expanded

### LINKED ISSUE #209: Add mouseover behavior for filter card when selected
We have a suggestion to make the filter card clear to the user when selected.

Couple ideas during discussion.
* Change in color
* Change cursor to hand

We can tackle this once we finished the functional element for the filter panel enhancement.

### Acceptance Criteria:
* Make the card obvious when selected

chlebowa:
Closes #209 

Added `boxshadow` property to and brightened background color in expanded filter card.
Selector uses the `aria-expanded` property rather than class names because it works in all BS versions and initiates `transition` immediately on click rather than after the card has been expanded.

image

### DIFF
--- a/inst/css/filter-panel.css
+++ b/inst/css/filter-panel.css
@@ -255,7 +255,7 @@ a.remove_all:hover {
 .filter-card:has(> .filter-card-header[aria-expanded="true"]) {
   transition-property: box-shadow, background-color;
   transition-duration: 150ms;
-  box-shadow: 0 0 5px 2px rgb(0 0 0 / 20%);
+  box-shadow: 0 0 5px 2px #00000033;
   background-color: #ffffffb8;
 }
 
--- a/inst/css/filter-panel.css
+++ b/inst/css/filter-panel.css
@@ -278,6 +278,12 @@ a.remove_all:hover {
   background-color: var(--bs-gray-200, #e9ecef);
 }
 
+.filter-card-header[aria-expanded="true"]:hover {
+  transition-property: background-color;
+  transition-duration: 150ms;
+  background-color: #f5f5f5;
+}
+
 .filter-card-title {
   display: flex;
   flex-direction: row;
--- a/inst/css/filter-panel.css
+++ b/inst/css/filter-panel.css
@@ -253,10 +253,15 @@ a.remove_all:hover {
 }
 
 .filter-card:has(> .filter-card-header[aria-expanded="true"]) {
-  transition-property: box-shadow, background-color;
+  transition-property: background-color;
+  transition-duration: 150ms;
+  background-color: #ffffffb8;
+}
+
+.filter-card:has(> .filter-card-header[aria-expanded="true"]):hover {
+  transition-property: box-shadow;
   transition-duration: 150ms;
   box-shadow: 0 0 5px 2px #00000033;
-  background-color: #ffffffb8;
 }
 
 .filter-card-header {
--- a/inst/css/filter-panel.css
+++ b/inst/css/filter-panel.css
@@ -255,13 +255,13 @@ a.remove_all:hover {
 .filter-card:has(> .filter-card-header[aria-expanded="true"]) {
   transition-property: background-color;
   transition-duration: 150ms;
-  background-color: #ffffffb8;
+  background-color: var(--bs-gray-100, #f8f9fa);
 }
 
 .filter-card:has(> .filter-card-header[aria-expanded="true"]):hover {
   transition-property: box-shadow;
   transition-duration: 150ms;
-  box-shadow: 0 0 5px 2px #00000033;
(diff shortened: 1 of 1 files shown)


donyunardi:
Looking good. I like transition setup that you did. The white background on the filter card also makes it popup.
However, not a big fan with `boxshadow`, but this is personal preference so let's hear another opinion.

Just as an alternative, I removed the `boxshadow` and preserve the hover color on filter-header when a filter card is selected.

https://github.com/insightsengineering/teal.slice/assets/8597300/5e9503a1-dd50-46a3-bc70-3ed8bc60eb89

WDYT?

chlebowa:
Your cards have no hover beharior.

lcd2yyz:
I don't know if there are (better) styling options than `boxshadow`, but at least I find it helpful to draw one's eyes to the card in its entirety, including the label/value along with the selection UIs, making the visualization "wholesome". 

With respect to @donyunardi 's preserved color-change in the card title - I kind of expected the color change to follow my hover, so it did feel a bit weird seeing two card titles highlighted (one expanded and one collapsed) at the same time. But also a personally preference.

So without other options to compare it to, I would say I like the new styling @chlebowa added a bit more.

inst/css/filter-panel.css
```diff
@@ -252,6 +252,18 @@ a.remove_all:hover {
...
+  transition-property: background-color;
+  transition-duration: 150ms;
+  background-color: #ffffffb8;
+}
+
+.filter-card:has(> .filter-card-header[aria-expanded="true"]):hover {
+  transition-property: box-shadow;
+  transition-duration: 150ms;
+  box-shadow: 0 0 5px 2px #00000033;
+}
```
vedhav, line 265: The box-shadow is distracting. Even without the shadow, the change in the background color is enough to distinguish between the selected and unselected filter.
Screenshot 2024-03-11 at 12 04 15 PM

inst/css/filter-panel.css
```diff
@@ -267,11 +279,16 @@ a.remove_all:hover {
   border-bottom-right-radius: 0;
 }
 
-.filter-card-header:hover,
-.filter-card-body:hover {
+.filter-card-header:hover {
```
vedhav, line 282:
```suggestion
.filter-card-header:hover {
  cursor: pointer;
```
One important requirement to show that the filter-header card can be clicked was to implement a cursor pointer.

inst/css/filter-panel.css
```diff
@@ -267,11 +279,16 @@ a.remove_all:hover {
...
   background-color: var(--bs-gray-200, #e9ecef);
 }
 
+.filter-card-header[aria-expanded="true"]:hover {
+  transition-property: background-color;
+  transition-duration: 150ms;
+  background-color: #f5f5f5;
```
vedhav, line 289:
```suggestion
  background-color: #eaecef;
```
It would be nice to use the same background that we use for the active hover of the card header, regardless of whether the card is expanded or not.

https://github.com/insightsengineering/teal.slice/assets/49812166/f06fbe3f-2a9a-4858-b902-bc879cbd4af3

vedhav (changes requested):
Added a few comments. `Pointer` is the main change I'm requesting. Others are optional.

inst/css/filter-panel.css
chlebowa: With the default hover-darkening the contrast between the highlighted card body and the darkened card header is jarring.

inst/css/filter-panel.css
vedhav: In that case I would recommend a mild scale transition along with increased opacity on the close buttons on hover WDYT?

vedhav: We already seem to do the opacity change to the outer buttons
Screenshot 2024-03-11 at 6 29 22 PM

chlebowa: When you change filter state, the rewind and reset buttons will appear in the header, which have full opacity.

donyunardi:
Another idea, how about changing the header color when the filter card is selected and persist it until the card is collapsed?

https://github.com/insightsengineering/teal.slice/assets/8597300/066c4c07-5ede-4737-9c29-375c0af8dd87

inst/css/filter-panel.css
donyunardi: I tried the pointer suggestion and I think this can easily be missed by user.

chlebowa:
@donyunardi so you want to replace `boxshadow` on the filter card with `background-color` on the header, correct?

Note that two cards may be unfolded at the same time if they correspond to different datasets. With no hover style there is no distinction between them.

donyunardi (approved):
We'll move forward with @chlebowa design.

vedhav (approved):

Issue #445, opened 2023-08-16
unify use of `tags$*` across the package


chlebowa:
`shiny` functions that create `HTML` tags (re-exported from `htmltools`) can be used directly or as elements of the `tags` list: 
```
div("some content")
tags$div("some content")
```

In `teal.slice` most `div` calls are direct but `span` calls usually use `tags$`.
Pick one and apply throughout the package.

--------------------------------------------------
## similar changes done here as part of this PR
- [x] teal https://github.com/insightsengineering/teal/pull/1122
- [ ] teal.data (not required)
- [ ] teal.code (not required)
- [x] teal.transform https://github.com/insightsengineering/teal.transform/pull/203
- [ ] teal.reporter (not required)
- [ ] teal.slice  https://github.com/insightsengineering/teal.slice/pull/566
- [x] teal.widgets https://github.com/insightsengineering/teal.widgets/pull/230
- [ ] teal.logger (not required)
- [ ] teal.modules.clinical https://github.com/insightsengineering/teal.modules.clinical/pull/1098
- [ ] teal.modules.general https://github.com/insightsengineering/teal.modules.general/pull/708
- [x] teal.goshawk https://github.com/insightsengineering/teal.goshawk/pull/268
- [ ] teal.modules.helios  https://github.com/insightsengineering/teal.modules.helios/pull/124
- [x] teal.modules.hermes https://github.com/insightsengineering/teal.modules.hermes/pull/367
- [ ] teal.osprey https://github.com/insightsengineering/teal.osprey/pull/259

kartikeyakirar:
The majority of common HTML tags are directly accessible through re-exports from the` htmltools` package, including but not limited to` a, br, code, div, em, h1 through h6, hr, HTML, htmlTemplate, img, includeCSS, includeHTML, includeMarkdown, includeScript, includeText, is.singleton, p, pre, singleton, span, strong, suppressDependencies, tag, tagAppendAttributes, tagAppendChild, tagAppendChildren, tagGetAttribute, tagHasAttribute, tagList, tags, tagSetChildren, validateCssUnit, and withTags.` However, certain tags like `td, tr, table,` and` label `are exceptions and used in `teal.slice`and are not re-exported in this manner; thus, they require `tags$ `prefix .

so for unifying I prefer to do prefix` tags$`  for all HTML tags  to ensures a standardized codebase.

WDYT?

chlebowa:
It's probably the way to go.

[...]

PR #566, opened 2024-03-05
445 unify html tags

kartikeyakirar:
fixes https://github.com/insightsengineering/teal.slice/issues/445


chlebowa (changes requested):
`shiny::` prefixes are superfluous because of `@import shiny`. Please revert.

averissimo (approved):
Looks good! I believe you caught everything!

tests still have `shiny::` prefix, but that shouldn't block this.

@chlebowa anything to add?

PR #568, opened 2024-03-06
335 modify Rd database

### LINKED ISSUE #335: `teal_slice` should show up in package index
The `teal_slice` help topic extensively documents the `teal_slice` and `teal_slices` classes and explains at length how filter states are defined and handled. Although all the functions in that topic are not exported, I believe `teal_slice` should be visible in the package index in RStudio's Help tab.

chlebowa:
Fixes #335 

Package index is built during the installation process but only once the package has been actually installed.
A file is created, `<package>/html/00Index.html`, that is the source for the index page.
The file is built based on a parsed Rd  database that is stored in `<package>/Meta/Rd.rds`.

Here, a script is created that will modify the existing `Rd.rds` and add it to `inst/Meta`, so that it the modified version can be used during index creation.

Note that if documentation files are changed and the index should be updated, this script must be run again. It should be done before every release at least.


m7pr:
For custom operations during the build I would suggest creating a script that does everything needed to have the package properly build, that would include so far:

- runs custom build operation (like your script)
- (maybe) runs some custom functions that edit pkgdown sites
- actually builds the package and packs it in the binary

And then move to the convention where we use scripts for building packages and we don't use custom RStudio buttons nor shortcuts nor custom commands.

m7pr:
I prefer this solution over https://github.com/insightsengineering/teal.slice/pull/569 as we modify the package before the build. So nobody needs an access to R library during or after package installation, which removes the issue that @pawelru mentioned here https://github.com/insightsengineering/teal.slice/pull/569#issuecomment-1981245678 (inability to over-write package files).

chlebowa:
Well, unfortunately it results in a check warning: 
```
* checking package subdirectories ... WARNING
Found the following non-empty subdirectories of ‘inst’ also used by R:
  inst/Meta
It is recommended not to interfere with package subdirectories used by
R.
```

PR #565, opened 2024-03-04
335 bring `teal_slice` and `teal_slices` to package INDEX

m7pr:
Proposition on how to solve #335

- Removed `@rdname teal_slice` from utility functions for `teal_slice` 
- Created customized `@usage` section to display both `teal_slice` and utility functions on the same page
- Result: `teal_slice` appearing in package index, where utility functions only appear in `teal_slice` documentation page

If this is acceptable, I will do the same for `teal_slices`

index
documentation

PR #569, opened 2024-03-06
335 remove aliases from package index

### LINKED ISSUE #335: `teal_slice` should show up in package index

chlebowa:
Fixes #335 

Package index is built during the installation process but only once the package has been actually installed.
A file is created, `<package>/html/00Index.html`, that is the source for the index page.

Here, a function is added that modifies `html/00Index.html` to remove HTML elements that link to a specified topic through aliases.
The function is called in the loading hook.

`@keywords internal` is removed from all functions that are included in `teal_slice`, `teal_slices`, and `filter_state_api`.


chlebowa:
Then no changes are made
```
  if (identical(index_file, "") || !isTRUE(utils::file_test("-w", index_file))) {
    return(invisible(FALSE))
  }
```

gogonzo:
I would like this solution if `pkgdown` used installed package index - I run `pkgdown::build_site` and it seems that `pkgdown` build site from the source including all aliases in the rendered site

m7pr:
Maybe for `pkgdown` we need some custom code in `pkgdown.yml` to be executed so that the pkgdown-index gets cleaned the same way it's done for package index.

pawelru:
> Then no changes are made

I am afraid that this would make it working only for subset of users. To my surprise, centrally managed environments is quite popular in the pharma world - even internally we have a BEE environment managed like that.

Is this really the only way to address this problem? Can it be done with modification of roxygen tags? Maybe we can slightly change the approach of documentation to obey this issue? Just thinking loudly...

m7pr:
I actually like the idea, if this plays with `pkgdown` as it also clears aliases for `filter-panel-api` that are also duplicated (4 times).

chlebowa:
`pkgdown` provides a way to exclude a topic from the reference but raises an error if you do :unamused:

chlebowa:
> Is this really the only way to address this problem? Can it be done with modification of roxygen tags? Maybe we can slightly change the approach of documentation to obey this issue? Just thinking loudly...

Be my guest.

m7pr:
@pawelru 

> Can it be done with modification of roxygen tags? 

It's a bigger problem and we research multiple approaches. You can read https://github.com/insightsengineering/teal.slice/issues/335 and other PR https://github.com/insightsengineering/teal.slice/pull/565

pawelru:
> > Can it be done with modification of roxygen tags?
> 
> It's a bigger problem and we researched multiple approaches. You can read #335 and other PR #565

> > Is this really the only way to address this problem? Can it be done with modification of roxygen tags? Maybe we can slightly change the approach of documentation to obey this issue? Just thinking loudly...
> 
> Be my guest.

I tried to avoid going deeper if possible. I need to focus on something else at the current moment. I can do that but a little bit later. Just wanted to flag this out for you in case you are unaware.

chlebowa:
> `pkgdown` provides a way to exclude a topic from the reference but raises an error if you do 😒

This seems to be a bug that was supposedly fixed in 2022. I will raise the issue again but even if they do fix it, I don't think they will do it soon.

chlebowa:
> I tried to avoid going deeper if possible. I need to focus on something else at the current moment. I can do that but a little bit later. Just wanted to flag this out for you in case you are unaware.

We are four days in, join us when you can :wink:

m7pr:
>> pkgdown provides a way to exclude a topic from the reference but raises an error if you do 😒

> This seems to be a bug that was supposedly fixed in 2022. I will raise the issue again but even if they do fix it, I don't think they will do it soon

Maybe we can run it with `suppressWarnings/suppressMessages` to avoid having it be displayed during GHA builds?

chlebowa:
> Maybe we can run it with `suppressWarnings/suppressMessages` to avoid having it be displayed during GHA builds?

```
-- Building function reference -------------------------------------------------
Error: 
! in callr subprocess.
Caused by error in `check_missing_topics(rows, pkg)`:
! All topics must be included in reference index
✖ Missing topics: filter_state_api
ℹ Either add to _pkgdown.yml or use @keywords internal
ℹ See `$stdout` for standard output.
Type .Last.error to see the more details.
```

Good luck.

m7pr:
Got it, how do you run it? Did you add `reference` section to `pkgdown.yml` and run `build_site` or did you use `build_reference`?

chlebowa:
image

m7pr:
Looks like promising path!

chlebowa:
@pawelru Re: file permissions, I forgot to mention that a package is test-loaded _during installation_ and the index file is modified then.

pawelru:
> @pawelru Re: file permissions, I forgot to mention that a package is test-loaded _during installation_ and the index file is modified then.

That's great to hear. It means that my concern is not valid and I no longer see any blockers. However, I do encourage you to continue with looking for alternatives and do compare different approaches. I personally would prefer a fix to be incorporated in the already built package compared to apply fix on install. That's of course as long as it is possible.

Now I cannot say it's buggy or incorrect. It's just non-standard and I am afraid that I am just unable to assess all possible outcomes or edge-cases. Why I do care? Non-standard ways of working introduce some difficulty and requires some/deep knowledge and generally speaking i do want to have a low and cheap maintenance code. I really hope you know what I am mean.
I did a very quick search on GH CRAN mirror: https://github.com/search?q=org%3Acran+path%3A*.R+00Index.html&type=code
Out of 26k+ repos only 16 of them are using `00Index.html` file and looking briefly most of them do read only. (It looks like only one `knitr` does write but I don't want to spend my time trying to understand it - that's not the point to get an exact number).

I know I am sitting on a very comfy reviewer position but I think that there is enough IQ points already there to came up with something good :) You are all clever folks and I believe in you 💪

chlebowa:
> Out of 26k+ repos only 16 of them are using `00Index.html` file

False.
The file is built by internal functions of the `tools` package upon installation. Every package that has a package index page (which I believe means virtually every package in existence) has that page. Why the page is not built with the `INDEX` file from the package root directory, which can be modified in the source package - I don't know.

> I do encourage you to continue with looking for alternatives and do compare different approaches.

Believe me, I tried. This is the best I was able to come up with. See https://github.com/insightsengineering/teal.slice/pull/568 for an inferior alternative.

Now, is this bulletproof? I don't know. For example, my way of dropping rows from the HTML file is rather primitive and makes some assumptions.

pawelru:
Ahhh I knew you will find something in my statement :) I really tried my best! Let me correct myself. That statement should go as follows

Out of 26k+ packages on CRAN, only 16 are _accessing_ `00Index.html` file in its sources. _Accessing_ for _read_ / _write_ or any other purpose. They all might _have_ it (and _use_ it) but only very small number of packages is doing actual modifications.

Anyway, that's just to show why I think it's non-standard. I will try to have a look at this but in the middle of next week the earliest.

chlebowa:
> Out of 26k+ packages on CRAN, only 16 are _accessing_ `00Index.html` file in its sources.

Fair :smiley: One of the reasons it is rarely accessed is it does not exist in the source. I agree, it is nonstandard.

chlebowa:
I raised an issue in the `pkgdown` repo but no help will be coming.

This is what I have been able to accomplish with the current state of affairs:

image

image


I suggest we apply it as is.

PR #589, opened 2024-05-08
588 remove link to `teal` package form docs

### LINKED ISSUE #588: R CMD CHECK note: `Package unavailable to check Rd xrefs: ‘teal’`
https://github.com/r-universe/pharmaverse/actions/runs/8984414175/job/24676515113
```
* checking Rd cross-references ... NOTE
Package unavailable to check Rd xrefs: ‘teal’
```

This is the cause: https://github.com/insightsengineering/teal.slice/blob/4f8dabc43c1c97a3e1b91c486c68050c86e30900/R/teal_slices.R#L84

`teal.slice` cannot use `teal` in its docs because it does not have `teal` in its dependencies (hard or soft). Moreover, we cannot add `teal` to deps because that would be a circular dependency A -> B and B -> A. Let's remove it

m7pr:
Fix #588

### DIFF
--- a/R/teal_slices.R
+++ b/R/teal_slices.R
@@ -81,7 +81,7 @@
 #'
 #' @seealso
 #' - [`teal_slice`] for creating constituent elements of `teal_slices`
-#' - [`teal::slices_store`] for robust utilities for saving and loading `teal_slices` in `JSON` format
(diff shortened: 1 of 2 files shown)


m7pr:
I have read the CLA Document and I hereby sign the CLA

pawelru (approved):

PR #593, opened 2024-05-23
Reduce unwanted reactivity on filter panel changes

### LINKED ISSUE #62: Implement bindCache wherever possible to avoid unnecessary reactivity on "unfiltered" data.

vedhav:
Closes #62 

Pending things before it can be undrafted:
1. Make the `private$reactive_call` for a given `sid`. Right now it will not work when `sid` is provided. A possible solution is to make the `sid` a private object and use it inside the server logic.
2. Figure out the right place to observe `self$get_call()` to extract the call.
3. Test for normal `data.frame`. `CDISC` `data.frame` and `MAE` object.

Example app to test:

```r
library(shiny)
library(teal.data)

  list(
    iris = iris,
    ADSL = rADSL, ADAE = rADAE,
    MAE = hermes::multi_assay_experiment
  ),
  join_keys = default_cdisc_join_keys[c("ADSL", "ADAE")]
)

ui <- fluidPage(
  fluidRow(
    column(
      width = 9,
      DT::DTOutput("iris_table"),
      DT::DTOutput("adsl_table"),
      DT::DTOutput("adae_table"),
      shiny::verbatimTextOutput("mae_text")
    ),
    column(width = 3, datasets$ui_filter_panel("filter_panel"))
  )
)

  datasets$srv_filter_panel("filter_panel")
  iris_filtered_data <- reactive(datasets$get_data(dataname = "iris", filtered = TRUE))
  adsl_filtered_data <- reactive(datasets$get_data(dataname = "ADSL", filtered = TRUE))
  adae_filtered_data <- reactive(datasets$get_data(dataname = "ADAE", filtered = TRUE))
  mae_filtered_data <- reactive(datasets$get_data(dataname = "MAE", filtered = TRUE))

  observeEvent(iris_filtered_data(), {
    print("IRIS changed!!!")
  })
  observeEvent(adsl_filtered_data(), {
    print("ADSL changed!!!")
  })
  observeEvent(adae_filtered_data(), {
    print("ADAE changed!!!")
  })
  observeEvent(mae_filtered_data(), {
    print("MAE changed!!!")
  })
  output$iris_table <- DT::renderDT(head(iris_filtered_data()))
  output$adsl_table <- DT::renderDT(head(adsl_filtered_data()))
  output$adae_table <- DT::renderDT(head(adae_filtered_data()))
  output$mae_text <- renderPrint({
    mae_filtered_data()
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/R/FilteredDataset.R
+++ b/R/FilteredDataset.R
@@ -42,6 +42,7 @@ FilteredDataset <- R6::R6Class( # nolint
       private$dataname <- dataname
       private$keys <- keys
       private$label <- if (is.null(label)) character(0) else label
+      private$reactive_call <- reactiveVal()
 
       # function executing reactive call and returning data
       private$data_filtered_fun <- function(sid = "") {
@@ -53,7 +54,7 @@ FilteredDataset <- R6::R6Class( # nolint
         }
         env <- new.env(parent = parent.env(globalenv()))
         env[[dataname]] <- private$dataset
-        filter_call <- self$get_call(sid)
+        filter_call <- private$reactive_call()
         eval_expr_with_msg(filter_call, env)
         get(x = dataname, envir = env)
       }
@@ -301,6 +302,10 @@ FilteredDataset <- R6::R6Class( # nolint
             }
           )
 
+          observeEvent(self$get_call(), ignoreNULL = FALSE, {
+            private$reactive_call(self$get_call())
+          })
+
           observeEvent(self$get_filter_state(), {
             shinyjs::hide("filter_count_ui")
             shinyjs::show("filters")
@@ -374,6 +379,7 @@ FilteredDataset <- R6::R6Class( # nolint
     dataname = character(0),
     keys = character(0),
     label = character(0),
+    reactive_call = NULL, # reactiveVal
 
     # Adds `FilterStates` to the `private$filter_states`.
     # `FilterStates` is added once for each element of the dataset.
--- a/R/FilteredDataset.R
+++ b/R/FilteredDataset.R
@@ -44,10 +44,6 @@ FilteredDataset <- R6::R6Class( # nolint
       private$label <- if (is.null(label)) character(0) else label
       private$reactive_call <- reactiveVal()
 
-      observeEvent(self$get_call(), ignoreNULL = FALSE, {
-        private$reactive_call(self$get_call())
-      })
-
       # function executing reactive call and returning data
       private$data_filtered_fun <- function(sid = "") {
         checkmate::assert_character(sid)
@@ -305,6 +301,10 @@ FilteredDataset <- R6::R6Class( # nolint
               private$get_filter_states()[[x]]$srv_active(id = x)
             }
           )
+
+          observeEvent(self$get_call(), ignoreNULL = FALSE, {
+            private$reactive_call(self$get_call())
+          })
 
           observeEvent(self$get_filter_state(), {
             shinyjs::hide("filter_count_ui")
--- a/R/FilteredDataset.R
+++ b/R/FilteredDataset.R
@@ -44,6 +44,10 @@ FilteredDataset <- R6::R6Class( # nolint
       private$label <- if (is.null(label)) character(0) else label
       private$reactive_call <- reactiveVal()
 
+      observeEvent(self$get_call(), ignoreNULL = FALSE, {
+        private$reactive_call(self$get_call())
+      })
+
       # function executing reactive call and returning data
       private$data_filtered_fun <- function(sid = "") {
         checkmate::assert_character(sid)
@@ -301,10 +305,6 @@ FilteredDataset <- R6::R6Class( # nolint
               private$get_filter_states()[[x]]$srv_active(id = x)
             }
           )
-
-          observeEvent(self$get_call(), ignoreNULL = FALSE, {
-            private$reactive_call(self$get_call())
-          })
 
           observeEvent(self$get_filter_state(), {
             shinyjs::hide("filter_count_ui")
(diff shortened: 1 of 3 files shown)


gogonzo:
Looks good already. `sid` problem is a minor thing - it is only for reactive-counts in `FilterState`

vedhav:
@gogonzo I was trying to figure out what was broken when this `sid` is not passed properly and I can't seem to figure out a way to test this. The only thing I know is that it is used [here](https://github.com/insightsengineering/teal.slice/blob/71cee08aca11618a555649a4b5836626281ea460/R/FilterStates.R#L742)

gogonzo:
> @gogonzo I was trying to figure out what was broken when this `sid` is not passed properly and I can't seem to figure out a way to test this. The only thing I know is that it is used [here](https://github.com/insightsengineering/teal.slice/blob/71cee08aca11618a555649a4b5836626281ea460/R/FilterStates.R#L742)

Yu, I just noticed that when `count_type = "all"` it uses the same dataset for each `FilterState`.

What if you try to bind on a call here?
https://github.com/insightsengineering/teal/blob/5c3e3fd4380ef0fc7efc53ec1cdf4f5a7c922dd3/R/module_nested_tabs.R#L278

R/FilteredDataset.R
```diff
@@ -42,6 +42,11 @@ FilteredDataset <- R6::R6Class( # nolint
       private$dataname <- dataname
       private$keys <- keys
       private$label <- if (is.null(label)) character(0) else label
+      private$reactive_call <- reactiveVal()
+
+      observeEvent(self$get_call(), ignoreNULL = FALSE, {
```
gogonzo, line 47: this is constructor - it should be outside of a reactive context. It is only because `srv_teal` `insertUI` which is wrong and will be removed sooner or later.

vedhav: Reverted that change 👍🏽 
I was surprised that observing outside a module server worked just fine. Right now the observer is inside the `srv_active` which is also not correct. It should be somewhere similar to the global server init. I will try to move this to some server logic in the `FilterData` class.
But, this adds a new problem. The dataset is only dependant on the `private$reactive_call ()` which needs shiny to work. So, we have some unit tests failing. As, those tests used to be reactive on the `self$get_call()` before.

R/FilteredDataset.R
```diff
@@ -53,7 +54,7 @@ FilteredDataset <- R6::R6Class( # nolint
         }
         env <- new.env(parent = parent.env(globalenv()))
         env[[dataname]] <- private$dataset
-        filter_call <- self$get_call(sid)
+        filter_call <- private$reactive_call()
```
gogonzo, line 57: It won't work for filter-counts which has non-empty `sid` as an argument. Let me explain briefly how `sid` works and why you can't cache these object in `FilteredDataset`.

For given `FilterState(s)` configuration there are two kind of calls (and data) produced in the same time. When the filter is modified possibly multiple calls of `data_filtered_fun` are invoked. This means that this can't be cached here because this function needs to be called several times with different `sid`. 

1. All FilterState are combined into one call and it is returned to the module. For this case your current solution works just perfect
image

2. reactive counts in a single (each) FilterState are calculated on data filtered by every filter but not "this" filter.
image

-----

Why are counts so weird?
When you filter single column you'd like to still see the counts of the unfiltered ranges. If we don't show these counts user wouldn't know how many observations are outside of the filtered range.

| correct | incorrect |
| -------  | --------- |
|  |  |


Even more technical explaination [here](https://github.com/insightsengineering/teal.slice/pull/205)

<details>
<summary> example app </summary>

```r
# options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
library("teal.slice")
library(teal)
library(scda)
library(SummarizedExperiment)

data <- teal_data() |> within({
  # MAE <- hermes::multi_assay_experiment
  ADSL$empty <- NA
  ADSL$logical1 <- FALSE
  ADSL$numeric <- rnorm(nrow(ADSL))
  ADSL$categorical <- sample(letters[1:10], size = nrow(ADSL), replace = TRUE)

  ADSL$numeric[sample(1:nrow(ADSL), size = 10, )] <- NA
  ADSL$numeric[sample(1:nrow(ADSL), size = 10, )] <- Inf
  ADSL$logical[sample(1:nrow(ADSL), size = 10, )] <- NA
  ADSL$date[sample(1:nrow(ADSL), size = 10, )] <- NA
  ADSL$datetime[sample(1:nrow(ADSL), size = 10, )] <- NA
  ADSL$categorical[sample(1:nrow(ADSL), size = 10, )] <- NA

})

teal.data::datanames(data) <- c(
  #  "MAE",
  "ADSL", "ADTTE", "ADRS"
)
teal.data::join_keys(data) <- teal.data::default_cdisc_join_keys[teal.data::datanames(data)]

app <- init(
  data = data,
  modules = list(
    example_module("mod1"),
    modules(
      example_module("mod2")
    )
  ),
  filter = teal_slices(
    teal_slice("ADSL", "numeric", id = "numeric1"),
    teal_slice("ADSL", "numeric", id = "numeric2"),
    module_specific = TRUE,
    mapping = list(
      mod1 = "numeric1",
      mod2 = "numeric2"
    ),
    count_type = "all"
  )
)

runApp(app)

``` 

</details>

gogonzo (changes requested):
filter-counts are broken

donyunardi:
@vedhav @gogonzo 
Is this urgent? Otherwise, should we hold off on this movement until we finished other priorities?

gogonzo:
> @vedhav @gogonzo Is this urgent? Otherwise, should we hold off on this movement until we finished other priorities?

This is not urgent and working solution hasn't been provided. Let's deprioritize this

PR #597, opened 2024-06-07
revert `logger::log_shiny_input_changes`

m7pr:
Part of https://github.com/insightsengineering/coredev-tasks/issues/553


averissimo (approved):
LGTM!

Does as expected and there's no longer a barrage of log messages

PR #602, opened 2024-07-26
fix: duplicated slices are lingering the filter panel after DDL reload

averissimo:
# Pull Request

Companion of https://github.com/insightsengineering/teal/pull/1275

Note: There's a similar solution that is cleaner. I'll push it on Monday

#### Changes description

- Keeps track of `remove observers` on filter panel
- Destroys `remove observer` when clearing filter panel

### DIFF
--- a/R/FilterStates.R
+++ b/R/FilterStates.R
@@ -372,19 +372,10 @@ FilterStates <- R6::R6Class( # nolint
               added_state_names <- vapply(added_states(), function(x) x$get_state()$id, character(1L))
               logger::log_debug("FilterStates$srv_active@2 triggered by added states: { toString(added_state_names) }")
               lapply(added_states(), function(state) {
-                fs_callback <- state$server(id = fs_to_shiny_ns(state))
-                state_id <- state$get_state()$id
-                remove_observer <- observeEvent(
-                  once = TRUE, # remove button can be called once, should be destroyed afterwards
-                  ignoreInit = TRUE, # ignoreInit: should not matter because we destroy the previous input set of the UI
-                  eventExpr = fs_callback(), # when remove button is clicked in the FilterState ui
-                  handlerExpr = private$state_list_remove(state_id)
+                state$server(
+                  id = fs_to_shiny_ns(state),
+                  function() private$state_list_remove(state$get_state()$id)
                 )
-
-                # Keep track of all remove observers
-                remove_id_prefix <- sprintf("remove_%s", state_id)
-                remove_id <- sprintf("%s_%s", remove_id_prefix, sum(grepl(remove_id_prefix, names(private$observers))))
-                private$observers[[remove_id]] <- remove_observer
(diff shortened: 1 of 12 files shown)


R/FilterStates.R
```diff
@@ -373,12 +373,20 @@ FilterStates <- R6::R6Class( # nolint
...
                   once = TRUE, # remove button can be called once, should be destroyed afterwards
                   ignoreInit = TRUE, # ignoreInit: should not matter because we destroy the previous input set of the UI
                   eventExpr = fs_callback(), # when remove button is clicked in the FilterState ui
-                  handlerExpr = private$state_list_remove(state$get_state()$id)
+                  handlerExpr = private$state_list_remove(state_id)
                 )
+
+                # Each filter state_id can only have 1 active remove observer at the time
+                remove_id <- sprintf("remove_%s", state_id)
+                if (!is.null(private$observers[[remove_id]])) {
```
gogonzo, line 386: This shouldn't be here. If we have one instance of `FilteredData` and multiple `srv_active` (for each module), then it is removing observers from other tabs. For example when `module_specific = FALSE` and two modules use the same filter - it will be possible to remove filter from the last-rendered tab only.

averissimo: We still need to keep track of the observers that are created.

I removed the overwrite, but keeping any additional observers with `remove_<state_id>_<count>`

Any observer that has the state id on its name is removed on method call `state_list_remove` via

```r
state$destroy_observers() # already
lapply(
  Filter(function(x) grepl(state$get_state()$id, x, fixed = TRUE), names(private$observers)),
  function(x) private$observers[[x]]$destroy()
)
```

Ideally we would push the observer to the state$private$observers, but that is private and requires new API on `<teal.slice>/R/FilterState.R`

R/FilterStates.R
```diff
@@ -373,12 +373,18 @@ FilterStates <- R6::R6Class( # nolint
...
                   ignoreInit = TRUE, # ignoreInit: should not matter because we destroy the previous input set of the UI
                   eventExpr = fs_callback(), # when remove button is clicked in the FilterState ui
-                  handlerExpr = private$state_list_remove(state$get_state()$id)
+                  handlerExpr = private$state_list_remove(state_id)
                 )
+
+                # Keep track of all remove observers
+                remove_id_prefix <- sprintf("remove_%s", state_id)
+                remove_id <- sprintf("%s_%s", remove_id_prefix, sum(grepl(remove_id_prefix, names(private$observers))))
+                private$observers[[remove_id]] <- remove_observer
```
gogonzo, line 387: For sake of simplicity, you can also:
- move `FilterState$<private>$observers` to `FilterState$<public>$observers`
- set this `remove_observer` as an element of `FilterState$observers` class
- then `state_list_remove` can remain unchanged and will call `state$destroy_observers()`

PR #603, opened 2024-07-31
Feature branch PR

gogonzo:
- reduce the size of filter-panel by moving `ui/srv_add` to `ui/srv_active`
- change log level
- introduce `self$finalize` method to destroy all the observers and inputs (needed after data reloading)


R/FilterState.R
```diff
@@ -267,16 +268,27 @@ FilterState <- R6::R6Class( # nolint
...
+            handlerExpr = remove_callback()
+          )
+
           private$destroy_shiny <- function() {
-            logger::log_trace("Destroying FilterState inputs and observers; id: { private$get_id() }")
+            logger::log_debug("Destroying FilterState inputs and observers; id: { private$get_id() }")
+
+            if (session$isEnded()) {
+              return(NULL)
+            } # skip input removal if session has ended
```
averissimo, line 283: Finalize error is fixed by ignoring `session$input` cleanup when `session` has ended.

We should not try to clean this ourselves outside of a `teal` session, where we need to make sure the input is removed.

Same fix applied to `FilterStateExpr` that doesn't inherit from this class

averissimo (approved):
Let's go!

PR #604, opened 2024-07-31
Fix +/- add new filter buttons when changing data using teal_data_module

vedhav:
Fixes the +/- icon behavior on the filter panel.
Please checkout to the `fix-add-remove-ui@669_insertUI@main` in both `teal` and `teal.slice`
In `teal.slice` it just reverts one commit to the working version of the show/hide logic.

When the data changes using `teal_data_module` the filter panel observers are being called once again thus running the scope of the observers as many times the data changes.
The fix is to reinitialize the filter panel with new NS so the observers are not connected to the active UI anymore.

### Example app to test
```r
pkgload::load_all("teal")
library(MultiAssayExperiment)

data <- teal::teal_data_module(
  ui = function(id) {
    ns <- shiny::NS(id)
    shiny::tagList(
      shiny::textInput(ns("username"), label = "Username"),
      shiny::passwordInput(ns("password"), label = "Password"),
      shiny::actionButton(ns("submit"), label = "Submit")
    )
  },
  server = function(id, ...) {
    shiny::moduleServer(id, function(input, output, session) {
      shiny::eventReactive(input$submit, {
        data <- teal.data::teal_data() |>
          within(
            {
              logger::log_debug("Loading data...")
              iris <- iris
              data("miniACC", package = "MultiAssayExperiment", envir = environment())
            },
            username = input$username,
            password = input$password
          )
        data
      })
    })
  },
  once = FALSE
)

app <- init(
  data = data,
  modules = example_module()
)

shiny::runApp(app$ui, app$server)
```


R/FilteredDataset.R
```diff
@@ -235,16 +235,12 @@ FilteredDataset <- R6::R6Class( # nolint
...
+                  actionLink(
+                    ns("toggle_add_panel"),
+                    label = "",
+                    icon = icon("plus", lib = "font-awesome"),
                     class = "remove pull-right",
-                    tags$i(
-                      class = "fa fa-plus",
-                      title = "fold/expand transform panel",
-                      onclick = sprintf(
-                        "togglePanelItem(this, '%s', 'fa-minus', 'fa-plus');",
```
gogonzo, line 244: what is the root problem of `togglePanelItem`?

vedhav: The root problem is that the [shinyjs::toggle](https://github.com/insightsengineering/teal.slice/blob/3707239ec62a3ad2ce9df5037bd75c048598b79f/R/FilteredDataset.R#L346-L348) observer gets loaded as many times as the `srv_active` is called.

```
observeEvent(input$toggle_add_panel, {
  shinyjs::toggle("add_panel")
})
```

gogonzo: I think `observeEvent` can be removed when `togglePanelItem` is used. I forgot to remove `observeEvent` after replacing `actionLink` with `tags$a` with js function.

vedhav: Ah! true. Removing it should do it.

vedhav: But, the teal PR is still valid as it fixes other observers not getting re-initialized

vedhav: I still have some issue with the add. Let me test it a little.

gogonzo: We can have a call about `ns` of the filter-panel tomorrow with @averissimo . I think we are mixing few issues in the same time

gogonzo (approved):
please merge to feature branch once it is fine on your side

R/FilteredDataset.R
vedhav: I did talk with @averissimo today and the observer destroy issue that he is working on is for the filter state (The filter card observers that need to be removed when filter states are removed). And this is regarding removing observers of the whole filter panel (well the filter panel that we're using)

vedhav:
Can we please switch to the `shinyjs::toggle()` implementation @gogonzo?

The two problems with the current `togglePanelItem()` implementation are:
1. The file `togglePanelItem.js` is actually implemented and sourced inside the `teal` package. And, in `teal.slice` this is not sourced as well so it fails when a non-teal shiny app is created with `teal.slice`.
2. The reason why `togglePanelItem()` works well with the `teal`  is because the `targetID` to be shown/hidden is shown from the start so just toggling the display attribute is able to show/hide the target element. However, in this case, we want to initialize the `add_panel` element in the hidden state (display: none) and sometimes the UI initialized in the server not rendered unless some reactivity is triggering it when the display is visible.
 Additionally, there are two other places where we have the same show/hide logic implemented using shines in this package `minimise_filter_overview` and `minimise_filter_active`. And, the `togglePanelItem()` logic works for the former and not the latter.


https://github.com/user-attachments/assets/e805677e-a60a-4fef-b3ca-66b26caf9742

gogonzo:
> Can we please switch to the `shinyjs::toggle()` implementation @gogonzo?

We can switch but not at this moment with current solutions (switching to toggle and rerender filter-panel with new namespace). I don't like the idea of handling toggling on the server side for so many panels, I'd like to move them all to the client side. Let's look for more alternatives.

> The file togglePanelItem.js is actually implemented and sourced inside the teal package. And, in teal.slice this is not sourced as well so it fails when a non-teal shiny app is created with teal.slice. But, if we really wanted this we can get over this with the name change or sourcing this inside the filter UI again and it will override the function from teal.

Good point but we just need to include js source by fixing in `FilteredDataset` a line `include_js_files(pattern = "icons|togglePanelItem"),`

> The reason why togglePanelItem() works well with the teal is because the targetID to be shown/hidden is shown from the start so just toggling the display attribute is able to show/hide the target element. However, in this case, we want to initialize the add_panel element in the hidden state (display: none) and sometimes the UI initialized in the server not rendered unless some reactivity is triggering it when the display is visible.

Fair enough. I see now, that in `FilterStates$ui_add` we `renderUI`, which doesn't compute once it is hidden, and envoking `togglePanelItem` doesn't trigger shiny reactivity, so ui add is not rendered. Render happens when any reactive event triggers within this `FilterStates$srv_add` module. We can either:
- Fix `togglePanelItem` to trigger renderXyz contained within panel.
- `suspendWhenHidden = FALSE` not a good idea but not the worst neither as it involves comparison of colnames vs active states variables but in all teal_modules x datasets.

vedhav:
@gogonzo Made a few improvements to the filter panel icons and their actions.
These are the rules that we follow from the UI/UX perspective:
1. The collapse/remove icons are only displayed when they can be used. In other words, clicking them always performs some UI change, if this is not possible they are hidden.
2. The add filter icon is moved along with the dataset name so that it is separated from the collapse/hide icon which can disappear based on the filter state.
3. The add filter adds new UI which is easily identifiable from the active filters because after clicking the add button it transforms into a tab with the new UI inside the tab body.
4. Collapsing the UI should also collapse the add/remove filters tab.

Screenshot 2024-08-05 at 8 01 32 AM

Example of how the add/remove tab button would look when the dataset name is too long.
Screenshot 2024-08-05 at 8 12 00 AM

R/FilterState.R
```diff
@@ -284,6 +284,8 @@ FilterState <- R6::R6Class( # nolint
             lapply(private$observers, function(x) x$destroy())
           }
 
+          private$state_history <- reactiveVal(list())
```
vedhav, line 287: FYI, This ensures that when we call the server again (happens when teal tab is switched). The state history is clear, else we would get the `back` and `reset` buttons during init.

Screenshot when a new teal tab was clicked by the user and this was the state of the filter panel during init.
Screenshot 2024-08-05 at 8 20 30 AM

R/FilteredData.R
```diff
@@ -552,23 +554,27 @@ FilteredData <- R6::R6Class( # nolint
...
-            class = "remove_all pull-right"
-          ),
-          actionLink(
-            inputId = ns("remove_all_filters"),
-            label = "",
-            icon("circle-xmark", lib = "font-awesome"),
-            title = "Remove active filters",
-            class = "remove_all pull-right"
+          # class = "filter-panel-active-header",
+          style = "display: flex; justify-content: space-between;",
```
vedhav, line 560: There are a lot of inline styles added because when we work on the sidebar panel redesign we would have to work on a lot of CSS and inline CSS makes it easier to know what style is added for this element without switching back and forth.
CSS is added only is cases where it was necessary because we needed to switch class in the JS to change the style and inline CSS could not help there,

R/FilteredData.R
```diff
@@ -552,23 +554,27 @@ FilteredData <- R6::R6Class( # nolint
       tags$div(
         id = id, # not used, can be used to customize CSS behavior
         class = "well",
+        include_js_files(pattern = "togglePanelItem"),
```
vedhav, line 557: Note that we are including the `togglePanelItem` in both `ui_filter_panel` and `ui_active` because teal app only uses the active module directly.

gogonzo (commented):
See the comments. I think `is_filter_removable` and toggle panel can be split into two PRs, so that comments over one improvement doesn't block the other.

R/FilteredDataset.R
```diff
@@ -322,17 +340,47 @@ FilteredDataset <- R6::R6Class( # nolint
...
 
+          output$collapse_ui <- renderUI({
+            req(self$is_filter_collapsible())
+            tags$a(
+              id = session$ns("collapse"),
+              class = "filter-icon",
+              tags$i(
+                id = session$ns("collapse_icon"),
+                class = "fa fa-angle-down",
+                title = "fold/expand transform panel",
```
vedhav, line 351:
```suggestion
                title = "fold/expand dataset filters",
```

R/FilteredData.R
```diff
@@ -620,8 +620,25 @@ FilteredData <- R6::R6Class( # nolint
...
+                sapply(
+                  self$get_filter_state(),
+                  function(x) x$anchored
+                )
+              )
+            ) {
+              filter_active <- TRUE
+            }
+          }
+          filter_active
```
gogonzo, line 637:
```suggestion
          non_anchored <- Filter(function(x) x$anchored, self$get_filter_state())
          isTRUE(length(non_anchored) > 0)
```

R/FilteredData.R
```diff
@@ -616,10 +621,21 @@ FilteredData <- R6::R6Class( # nolint
...
+            inputId = session$ns("remove_all_filters"),
+            label = "",
+            icon("circle-xmark", lib = "font-awesome"),
+            title = "Remove active filters",
+            class = "remove_all"
+          )
+        })
+
         private$observers[[session$ns("get_filter_count")]] <- observeEvent(
           eventExpr = private$get_filter_count(),
```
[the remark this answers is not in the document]
vedhav: They are now reactives 👍🏽

gogonzo (commented):
Some errors in the js console 

image

vedhav:
The JS errors in the console are fixed in the commit 988a75feccaeb14df635123b31d564e8eab06d1e. It was because the observer was triggered before the UI is initialized because in teal we call the server module inside the `renderUI`.

R/FilteredDataset.R
```diff
@@ -322,17 +311,59 @@ FilteredDataset <- R6::R6Class( # nolint
             }
           )
 
-          private$observers[[session$ns("get_filter_state")]] <- observeEvent(self$get_filter_state(), {
+          is_filter_collapsible <- reactive({
+            length(self$get_filter_state()) != 0
```
gogonzo, line 315: Let's create a `filter_count <- reactive()` and  remove `FilteredDataset$get_filter_count()`. 
If you remove `$get_filter_count()` ocurence here you'll be able to remove `get_filter_count` from all the classes.
 
```suggestion
            length(filter_count()) != 0
```

gogonzo (approved):
flawless 👍

PR #609, opened 2024-08-06
fix duplicated FilterState when reloading

gogonzo:
Fix Bug found by @vedhav when reloading data.

Actions to reproduce this bug:
- Initiate a teal app with `teal_data_module(once = FALSE)` and one active Species filter
- Remove the active filter and add the same Species filter
- Change the data
- Navigate to the teal app you'll notice two slices
- The filter state is not properly initialised when we do this
- This Bug occurs only when `include_varnames` is set

<details>
<summary> example </summary>

```r
pkgload::load_all("teal")

data <- teal::teal_data_module(
  ui = function(id) {
    ns <- shiny::NS(id)
    shiny::tagList(
      shiny::textInput(ns("username"), label = "Username"),
      shiny::passwordInput(ns("password"), label = "Password"),
      shiny::actionButton(ns("submit"), label = "Submit")
    )
  },
  server = function(id, ...) {
    shiny::moduleServer(id, function(input, output, session) {
      shiny::eventReactive(input$submit, {
        data <- teal.data::teal_data() |>
          within(
            {
              logger::log_debug("Loading data...")
              iris <- iris
            },
            username = input$username,
            password = input$password
          )
        data
      })
    })
  },
  once = FALSE
)

init(
  data = data,
  modules = modules(example_module(), example_module()),
  filter = teal_slices(
    teal_slice(
      dataname = "iris",
      varname = "Species",
      selected = "virginica"
    ),
    include_varnames = list(
      iris = c("Species")
    ),
    count_type = "all",
    allow_add = TRUE
  )
) |>
  shiny::runApp()
```

</details>


R/FilterStates.R
```diff
@@ -477,13 +479,6 @@ FilterStates <- R6::R6Class( # nolint
                   teal_slice(dataname = private$dataname, varname = input$var_to_add)
                 )
               )
-              logger::log_debug(
```
gogonzo, line 480: logs on exit are relevant but not worth it to double number of logs. We log only on entry only so we can keep track of the process order

R/FilterStates.R
```diff
@@ -440,12 +440,13 @@ FilterStates <- R6::R6Class( # nolint
...
           })
 
-
           output$add_filter <- renderUI({
             logger::log_debug(
               "FilterStates$srv_add@1 updating available column choices, dataname: { private$dataname }"
             )
             if (length(avail_column_choices()) == 0) {
+              # because input UI is not rendered on this condition but shiny still holds latest selected value
+              shinyjs::runjs(sprintf("Shiny.setInputValue('%s', null);", session$ns("var_to_add")))
```
averissimo, line 449: A method to avoid going through javascript. We use this in `FilterState`.

```suggestion
              .subset2(session$input, "impl")$.values$remove(session$ns("var_to_add"))
```

R/FilterStates.R
```diff
@@ -477,13 +479,6 @@ FilterStates <- R6::R6Class( # nolint
...
               )
-              logger::log_debug(
-                sprintf(
-                  "FilterStates$srv_add@2 added FilterState of variable %s, dataname: %s",
-                  input$var_to_add,
-                  private$dataname
-                )
-              )
             }
           )
```
averissimo, line 484:
```suggestion

          # Extra observer to destroy that clears all input values
          private$observers$input_remove <- list(
            destroy = function() {
              if (!session$isEnded()) {
                lapply(session$ns(names(input)), .subset2(input, "impl")$.values$remove)
              }
            }
          )
```

averissimo (commented):
I think we can avoid going through `javascript`,

There might be 2 possible solutions:

- Take advantage of `private$observers` to clear all inputs when finalize is called
- (alternatively) Remove the `input$var_to_add` directly from session object instead of `runjs`

R/FilterStates.R
gogonzo: I'm not a fan of this suggestion. I don't really know the difference between `setInputValue(nulll)` and `.$values$remove`. I agree that it is feels safer do interact with shiny socket through the session object than from javascipt.

R/FilterStates.R
gogonzo: I do like this suggestion. I'll implement this instead. Tested locally, works as expected 👍

gogonzo: Thanks @averissimo for this suggestion. I've included remove-inputs-callback to the list of `observers`. I've changed `observers` to `shiny_bindings` instead as it is not observers only. Feel free to suggest different field name (`shiny_bindings` is not precise enough as it contains observers and `list(destroy = function() kill input)`

R/utils.R
```diff
@@ -67,7 +67,16 @@ make_c_call <- function(choices) {
...
+#' Destroys inputs and observers stored in `private$session_bindings`
+#'
+#' @description
+#' Call a `destroy` method to remove `observer` and `input` from obsolete session which happens
+#' when `filter_panel_srv` is called again in new `FilteredData` object.
+#' Inputs are not stored directly in a field as they don't have `destroy` method. Instead, we
+#' store callback `destroy` function for inputs which removes bindings from a `session`.
+#' @param self,private slots of a `R6` class
+#' @return `NULL` invisibly
+.finalize_session_bindings <- function(self, private) {
```
averissimo, line 79:
```suggestion
#' @keywords internal
.finalize_session_bindings <- function(self, private) {
```

averissimo (approved):
Looks good! I think we need to add `@keywords internal`

Tested again and the problem is still gone.

PR #622, opened 2024-10-15
Allow non-standard datanames in filter data

averissimo:
# Pull Request

Fixes https://github.com/insightsengineering/teal/issues/1366

Related:

- https://github.com/insightsengineering/teal/pull/1382
- https://github.com/insightsengineering/teal.slice/pull/622
- https://github.com/insightsengineering/teal.data/pull/340

### Changes description

- Removed assertion on datanames that start with alphabetic character
- [x] Fix problem with JS namespace in filter panel
- [x] Fix crash when filtering using MAE (both SE and Matrix)
- [x ] ~Fix upload of snapshot file that is not compatible~
- [x] Ignore datanames that contain functions, language, expression (and other non-data objects)
    - https://github.com/insightsengineering/teal/issues/1352


averissimo:
Error in filter panel, see https://github.com/insightsengineering/teal.data/pull/340#issuecomment-2429154741

averissimo:
One way to solve that problem globally for `teal.slice` is to replace the `NS` and `moduleServer` that are used to create the Shiny's namespace.

The solution in 23c2bcfda03a44bd072f3cc144e11519c1a4d94b would add minimal code to the existing codebase.

WDYT?

averissimo:
An alternative to this proposal would be to identify call in the codebase that generates an id, and encode that id manually.

This would add a lot of noise as well as being prone to errors.

note: The proposal needs some extra tests to ensure all dataname types are working. As of now I only tested with `data.frame`-based dataname.

gogonzo:
> One way to solve that problem globally for teal.slice is to replace the NS and moduleServer that are used to create the Shiny's namespace.

I like the approach, not sure yet if we should make `NS/moduleServer` or `NS2/moduleServer2` just to avoid a confusion. 

> An alternative to this proposal would be to identify call in the codebase that generates an id, and encode that id manually.

There is a lot to cover and a lot to miss. I prefer to make own `NS/moduleServer`

R/FilterStates.R
```diff
@@ -165,7 +165,12 @@ FilterStates <- R6::R6Class( # nolint
       )
       if (length(filter_items) > 0L) {
         filter_function <- private$fun
-        data_name <- str2lang(private$dataname_prefixed)
+        data_name <- tryCatch(
```
gogonzo, line 168: I think it should be done where `dataname_prefixed` is created. Here it is little too late as dataname_prefixed could be `multi assay experiment[["experiment 1"]]`. Please search for `private$dataname_prefixed <-`. Also, I think it is better to not handle error but

```
private$dataname_prefixed <- if (!identical(datanames, make.names(dataname)) {
  sprintf("`%s`", dataname)
} else {
  dataname
}
```

R/FilteredData.R
```diff
@@ -306,10 +306,8 @@ FilteredData <- R6::R6Class( # nolint
     #'
     set_dataset = function(data, dataname) {
       checkmate::assert_string(dataname)
+      checkmate::assert_character(dataname, min.len = 1, any.missing = FALSE)
```
gogonzo, line 309: two above asserts are not okey

R/FilteredDataset.R
```diff
@@ -30,7 +30,7 @@ FilteredDataset <- R6::R6Class( # nolint
     #' @return Object of class `FilteredDataset`, invisibly.
     #'
     initialize = function(dataset, dataname, keys = character(0), label = attr(dataset, "label", exact = TRUE)) {
-      check_simple_name(dataname)
+      checkmate::assert_character(dataname, min.len = 1, any.missing = FALSE)
```
gogonzo, line 33: how dataname can be of length > 1? 
```suggestion
      checkmate::assert_string(dataname)
```

R/FilteredData.R
averissimo: Corrected

R/FilterStates.R
averissimo: Yup, that's right and it still needs some more extra conversions.

I'm playing with ``format(as.name("`yada`"))`` instead of that if clause

My tests today with MAE are also failing, initially because of this error, but also downstream in `FilterState$private$get_varname_prefixed`.

<details>

<summary>App</summary>

```r
pkgload::load_all("teal.data")
pkgload::load_all("teal")


options(
  teal.log_level = "TRACE",
  teal.log_layout = paste(
    "{crayon::bold(logger::colorize_by_log_level(level, levelr))}",
    "from {namespace}",
    "[{crayon::italic(format(time, \"%Y-%m-%d %H:%M:%S\"))}]",
    "{logger::grayscale_by_log_level(msg, levelr)}"
  ),
  teal.show_js_log = TRUE,
  # teal.bs_theme = bslib::bs_theme(version = 5),
  shiny.bookmarkStore = "server"
)

app <- teal::init(
  data = teal_data() |> within({
    `iris raw` <- cbind(iris, data.frame(date = Sys.Date() + sample(seq(100), nrow(iris), replace = TRUE)))
    `%iris%` <- iris
    env_temp <- new.env()
    data("miniACC", package = "MultiAssayExperiment", envir = env_temp)
    `a MAE` <- env_temp$miniACC
    miniACC_copy <- env_temp$miniACC
    rm(env_temp)
  }), 
  # filter = teal_slices(
  #   teal_slice("iris raw", "Species", selected = "setosa")
  # ),
  modules = modules(
    example_module("exotic"),
    example_module("exotic-specified", datanames = "iris raw")
  )
)

runApp(app)
```

</details>

averissimo:
> I like the approach, not sure yet if we should make NS/moduleServer or NS2/moduleServer2 just to avoid a confusion.

Both have drawbacks, 

\1. `NS`/`moduleServer` is not completely transparent to developers as I would expect those to be imported from Shiny

\2. `NS2`/`moduleServer2` is more explicit, but involves a bunch of find/replace as well as it might be easy for us to not use them when fixing some bug or adding new features

We could mitigate `2.` by overriding `NS`/`moduleServer` in `teal.slice` with a stop error saying it should not be used.

I tend towards `1.` as it's the cleaner solution and it has a very stable API. In the end we're just forwarding with modified id

R/FilterStates.R
averissimo: Works now for MAE and Matrix.

Flagged a new bug here that was already present in `teal.slice`

https://github.com/insightsengineering/teal.slice/issues/625

R/utils.R
```diff
@@ -81,3 +56,49 @@ make_c_call <- function(choices) {
...
+
+#' `moduleServer` wrapper to sanitize ids for shiny
+#'
+#' Special characters and spaces are not allowed in shiny ids (in JS)
+#'
+#' @noRd
+moduleServer <- function(id, module, session = getDefaultReactiveDomain()) { # nolint: object_name.
+  id <- js_encode(id)
+  shiny::moduleServer(id, module, session)
+}
```
averissimo, line 104: @gogonzo we can continue the discussion on the best solution to handle shiny ids normalization in this thread

R/utils.R
```diff
@@ -81,3 +56,49 @@ make_c_call <- function(choices) {
...
+#' We replace `%` with `__html_XX__` where `XX` is the HTML representation of
+#' the character.
+#'
+#' @param id (`character(1)`) The id string.
+#'
+#' @return Sanitized string that removes special characters and spaces.
+#'
+#' @keywords internal
+js_encode <- function(id) {
+  gsub("%([0-9]{2})", "__html_\\1__", utils::URLencode(as.character(id)))
```
averissimo, line 78: I think this is good enough.

However, a more complex alternative would be to convert special characters to utf8 _(the resulting id is more readable)_

``` r
js_encode <- function(x) {
  x_split <- strsplit(gsub("%", " %", utils::URLencode(as.character(x))), " ")[[1]]
  if (length(x_split) == 1) return(x)
  pattern <- "%[0-9]{2}"
  regmatches(x_split, regexpr(pattern, x_split)) <- paste(
    sep = "",
    "__u", 
    as.hexmode(
      sapply(
        utils::URLdecode(regmatches(x_split, regexpr(pattern, x_split))), 
        utf8ToInt, 
        USE.NAMES = FALSE
      )
    ), 
    "__"
  )
  paste(collapse = "", x_split)
}
js_encode(" an id with spaces and %")
#> [1] "__u20__an__u20__id__u20__with__u20__spaces__u20__and__u20____u25__"
js_encode("an_id_with_spaces")
#> [1] "an_id_with_spaces"
```

<sup>Created on 2024-10-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

averissimo: nevermind this. URLencode is already showing utf8 codes

R/utils.R
gogonzo: I lean towards `moduleServer` and `NS` as well 👍

gogonzo: I don't think we need a `js_encode`. I think `URLencode` is enough to convert weird ids.
```
utils::URLencode(as.character("iris"))
# [1] "iris"
utils::URLencode(as.character("%iris%"))
# [1] "%25iris%25"
utils::URLencode(as.character("iris.raw"))
# [1] "iris.raw"
utils::URLencode(as.character("raw iris"))
# [1] "raw%20iris"
utils::URLencode(as.character(".raw_data"))
# [1] ".raw_data"
```

R/utils.R
gogonzo: I don't think we need a `js_encode`. I think `URLencode` is enough to convert weird ids.
```
utils::URLencode(as.character("iris"))
# [1] "iris"
utils::URLencode(as.character("%iris%"))
# [1] "%25iris%25"
utils::URLencode(as.character("iris.raw"))
# [1] "iris.raw"
utils::URLencode(as.character("raw iris"))
# [1] "raw%20iris"
utils::URLencode(as.character(".raw_data"))
# [1] ".raw_data"
```

gogonzo: Quick question - is `"%"` harmful for id?

averissimo: It crashes jquery. You can test in a console with jquery loaded

```js
$("#an-id%20")
```

image

R/FilteredDataset.R
averissimo: I kept the previous assert in `check_simple_name`. I don't think it can in this initialize

R/utils.R
gogonzo: > It crashes jquery. You can test in a console with jquery loaded

I think it is ok to use urlEncode if id doesn't work natively with this. I think we should just prepend syntacticly incorrect characters with [escape character](https://api.jquery.com/category/selectors/).

averissimo: I went with urlEncode as it's a very wide range escape solution.

Let me test it out by escaping using that list. Although I think that we still need a special escape for space `" "`.

averissimo: The same with ``"`'`` (double quote, backtick, single quote)

gogonzo: > I think that we still need a special escape for space " " (in particular, to prevent problems if id is used in class)

Yes, space is included in this set ` !"#$%&'()*+,./:;<=>?@[\]^`{|}~`

averissimo: I meant that "space", and a bunch of other characters, may need to be a special case of escape character.

`<div id="a name.that:hover">`  is not a problem with ids `$("#a\\ name\.that\\:hover")` returns fine

But I'm very wary of the ramifications of having reserved selectors or separators as valid ids. Does that `.` or `:` have a impact? what if the HTML tag ends in `\"`? does it create a corrupt HTML?

averissimo: From my tests escaping the characters doesn't work, as the id parameter in the `UI`/`moduleServer` is used verbatim, we cannot say that the `id` parameter is `"%cbind%"` and in some cases use "\\%cbind\\%"

Nor can we use `"\\%cbind\\%"` as dataname as that will also need to be escaped when used in JQuery.

**Bottom line:** I believe this would be very hard to support and using urlEncode with `%XX -> __uXX__` replacement is the best solution we can hope to get.

R/utils.R
```diff
@@ -81,3 +56,57 @@ make_c_call <- function(choices) {
...
+#' and a hash of the original id is added to the beginning of the id.
+#'
+#' @param id (`character(1)`) The id string.
+#'
+#' @return Sanitized string that removes special characters and spaces.
+#'
+#' @keywords internal
+sanitize_id <- function(id) {
+  id_converted <- make.names(id)
+  if (identical(make.names(id), id)) {
```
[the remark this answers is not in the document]
averissimo: I agree, that's why two lines below it replaces `.` with `_`

I dropped that specific list as it will create a R-safe name.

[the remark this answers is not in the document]
averissimo: Check the latest commit, which adds some tests via `NS` call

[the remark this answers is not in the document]
gogonzo: Yes, but look, if your `id = <dataname>` contains dots, it might be considered as valid and it will be returned as id (`return(id)`). It means that if dataset is named `my.data` - the id will be `<namespace>-my.data` which will cause problems

```
make.names("not.good.for.shiny.id")
# "not.good.for.shiny.id"
```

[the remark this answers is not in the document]
averissimo: Ahh, good point.

Reverted back to pattern conversion.

tests/testthat/test-utils.R
```diff
@@ -1,25 +1,83 @@
...
+  testthat::it("should replace dots with `_` when id is otherwise valid", {
+    id <- "a.b"
+    ns <- NS("app")
+    testthat::expect_identical(
+      ns(id),
+      paste0("app-h", substr(rlang::hash(id), 1, 4), "_a_b")
+    )
+  })
+
+  testthat::it("should take vector input", {
```
averissimo, line 18: Vector input/output support is required

tests/testthat/test-utils.R
```diff
@@ -1,25 +1,83 @@
...
+      c(
+        paste0("app-h", substr(rlang::hash(id[1]), 1, 4), "_a_b"),
+        "app-a",
+        "app-b",
+        paste0("app-h", substr(rlang::hash(id[4]), 1, 4), "__c")
+      )
+    )
+  })
+
+  testthat::it("should allow for integer input", {
```
averissimo, line 32: Current usage also requires us to allow integer input

R/utils.R
```diff
@@ -81,3 +56,63 @@ make_c_call <- function(choices) {
...
+}
+
+#' `moduleServer` wrapper to sanitize ids for shiny
+#'
+#' Special characters and spaces are not allowed in shiny ids (in JS)
+#'
+#' @noRd
+moduleServer <- function(id, module, session = getDefaultReactiveDomain()) { # nolint: object_name.
+  id <- sanitize_id(id)
+  shiny::moduleServer(id, module, session)
```
averissimo, line 117: There is a potential problem with this PR, as a call to `session$ns(" a%problem ")` inside a moduleServer doesn't sanitize the id.

We can:

1. Live with this limitation (as it doesn't appear to affect us)
2. Modify the environment of the session to include our copy of `NS` function.
    - See suggestion

```suggestion

  ns_local <- NS
  ns_env <- environment(session$ns)
  ns_env$NS <- ns_local

  shiny::moduleServer(id, module, session)
```

averissimo: I think it's better to live with this problem, than to modify the `session` environment manually, this is a poor solution, but I couldn't find an alternative.

Opinions?

gogonzo (approved):
Wonderful! 👍

R/utils.R
```diff
@@ -81,3 +56,59 @@ make_c_call <- function(choices) {
...
+#' and a hash of the original id is added to the beginning of the id.
+#'
+#' @param id (`character(1)`) The id string.
+#'
+#' @return Sanitized string that removes special characters and spaces.
+#'
+#' @keywords internal
+sanitize_id <- function(id) {
+  # Left square bracket needs to be first in the pattern to avoid errors with pattern
+  pattern_escape <- "[] !\"#$%&'()*+,./:;<=>?@[\\^`{|}~]"
```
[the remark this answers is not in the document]
averissimo: Implemented and added a test to check for unicode characters outside of scope.

gogonzo (approved):

PR #624, opened 2024-10-22
Accept functions

gogonzo:
Part of https://github.com/insightsengineering/teal/issues/1352

Ignores other datasets than `data.frame` and `MultiAssayExperiment`. Unfilterable datasets shouldn't be displayed in the filter-panel as they will be printed independently in the teal-module_data_summary.


chlebowa (commented):
When we made the filter panel "data agnostic" it was considered an overdue success (I cannot find the exact comment that celebrated it).
This change reverses that, the filter panel becomes limited to strictly specified classes again.

R/FilteredDataset-utils.R
```diff
@@ -141,9 +141,5 @@ init_filtered_dataset.default <- function(dataset,
                                           parent, # ignored
                                           join_keys, # ignored
                                           label = attr(dataset, "label", exact = TRUE)) {
-  DefaultFilteredDataset$new(
-    dataset = dataset,
-    dataname = dataname,
-    label = label
-  )
+  NULL
```
chlebowa, line 144: How about an error message?

R/FilteredData-utils.R
```diff
@@ -29,7 +29,10 @@ init_filtered_data <- function(x, join_keys = teal.data::join_keys(), code, chec
       "init_filtered_data(check = 'No longer supported')"
     )
   }
-  FilteredData$new(x, join_keys = join_keys)
+  FilteredData$new(
+    Filter(function(obj) inherits(obj, c("data.frame", "MultiAssayExperiment")), x),
+    join_keys = join_keys
+  )
```
pawelru, line 35: I echo @chlebowa comment [here](https://github.com/insightsengineering/teal.slice/pull/624#pullrequestreview-2385205495). I think this should be delegated lower into the dataset constructor. Conceptually - a `function` (or anything like that) cannot become a filtered dataset.

gogonzo:
Closing this PR as changes in `teal.slice` are not needed. [PR in teal](https://github.com/insightsengineering/teal/pull/1393) alone fixes the problem.

PR #628, opened 2024-11-04
Update scheduled.yaml

pawelru:
Currently the max strategy fails with the following:
```
Error : ! Could not solve package dependencies:
* deps::/tmp/RtmpZGFxMH/filec219ed9d76: Can't install dependency insightsengineering/teal.data
* insightsengineering/teal.data: Can't install dependency teal.code (>= 0.5.0.9011)
```

`teal.code` is an indirect dependency via `teal.data`


averissimo (approved):
:+1:

PR #631, opened 2024-12-09
add setup-r-dependencies

donyunardi:
Related to https://github.com/insightsengineering/nestdevs-tasks/issues/65
Switched to setup-r-dependencies and added these packages for lookup-refs:
* teal.code
* teal.data
* teal.logger
* teal.widgets
* formatters
* rtables

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -58,6 +58,13 @@ jobs:
       )
     name: revdepcheck ↩️
     uses: insightsengineering/r.pkg.template/.github/workflows/revdepcheck.yaml@main
+    with:
+      lookup-refs: |
+        insightsengineering/teal.data
+        insightsengineering/teal.logger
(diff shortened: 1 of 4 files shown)


.github/workflows/scheduled.yaml
```diff
@@ -58,6 +58,14 @@ jobs:
       )
     name: revdepcheck ↩️
     uses: insightsengineering/r.pkg.template/.github/workflows/revdepcheck.yaml@main
+    with:
+      lookup-refs: |
```
m7pr, line 62: Do we need `deps-installation-method: setup-r-dependencies` in here?

m7pr: Looks like not, revdepcheck uses setup-r-dependencies already
https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/revdepcheck.yaml#L93

m7pr (approved):

PR #634, opened 2025-01-09
Fix `scheduled` inputs

averissimo:
Part of https://github.com/insightsengineering/nestdevs-tasks/issues/98

`extra-deps` was badly formed

PR #637, opened 2025-01-22
Support more MAE and SE versions in CI Pipelines

m7pr:
A follow-up after https://github.com/insightsengineering/teal.slice/pull/634 that allows to use older MAE and SE versions

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -43,6 +43,8 @@ jobs:
       additional-env-vars: |
         PKG_SYSREQS_DRY_RUN=true
       additional-repos: ${{ matrix.pair.repos }}
+      extra-deps: |
+        matrixStats (>= 1.5.0)
   branch-cleanup:
     if: >
       github.event_name == 'schedule' || (
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,11 +56,11 @@ Imports:
     utils
 Suggests:
     knitr (>= 1.42),
-    MultiAssayExperiment (>= 1.32.0),
+    MultiAssayExperiment,
     rmarkdown (>= 2.23),
-    SummarizedExperiment (>= 1.36.0),
-    testthat (>= 3.1.8),
-    withr (>= 2.1.0)
+    SummarizedExperiment,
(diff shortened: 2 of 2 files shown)


m7pr:
https://github.com/insightsengineering/teal.slice/actions/runs/12909783684

m7pr:
Hi, cc @llrs-roche, @gogonzo, and @donyunardi, as you previously expressed interest in the recent updates to fixing our pipelines.

Initially, I thought we needed to specify higher version requirements for `MultiAssayExperiment` and `SummarizedExperiment` (which are listed in `Suggests`) to ensure the Scheduled job scenarios could install all dependencies without issues. This was because one of the dependencies, `matrixStats`, was causing a crash. However, I later realized we only need to specify this third-level dependency (`matrixStats`) to be built from the latest Bioconductor release. This allows us to avoid version-locking `MultiAssayExperiment` and `SummarizedExperiment`, keeping them compatible with older versions of those packages as well.

Since this approach worked for `teal.slice`, I applied the same reduction of version requirements to `teal` and `tmg`. I’ll also make similar changes to `tmc` and other `teal.modules.*` packages.

Here are the relevant pull requests:  
- [teal PR #1460](https://github.com/insightsengineering/teal/pull/1460)](https://github.com/insightsengineering/teal/pull/1460)  
- [teal.modules.general PR #830](https://github.com/insightsengineering/teal.modules.general/pull/830)(https://github.com/insightsengineering/teal.modules.general/pull/830)

On a personal note, I find this to be an unnecessary complication since `MultiAssayExperiment` and `SummarizedExperiment` are only listed in `Suggests`.

donyunardi:
> On a personal note, I find this to be an unnecessary complication since MultiAssayExperiment and SummarizedExperiment are only listed in Suggests.

We need to start looking into this issue:
* https://github.com/insightsengineering/teal.slice/issues/500

Let's do it this year.

llrs-roche (approved):
The change makes sense and a the checks remained with the same failure as [main](https://github.com/insightsengineering/teal.slice/actions/runs/12850106779/job/35829410339) on [this branch](https://github.com/insightsengineering/teal.slice/actions/runs/12909783684/job/35998292831)

PR #642, opened 2025-01-30
Make sure the code chunks does not break because of warnings

vedhav:
Part of vignette review https://github.com/insightsengineering/nestdevs-tasks/issues/99

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,6 +55,7 @@ Imports:
     teal.widgets (>= 0.4.2),
     utils
 Suggests:
+    DT,
     knitr (>= 1.42),
     MultiAssayExperiment,
     rmarkdown (>= 2.23),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,7 +55,7 @@ Imports:
     teal.widgets (>= 0.4.2),
     utils
 Suggests:
-    DT,
+    DT (>= 0.3.0),
     knitr (>= 1.42),
     MultiAssayExperiment,
     rmarkdown (>= 2.23),
--- a/vignettes/teal-slice.Rmd
+++ b/vignettes/teal-slice.Rmd
@@ -61,8 +61,9 @@ To facilitate communication with `FilteredData`, we provide the filter panel API
 The example below uses the `set_filter_state` function to set state and the state is specified with `teal_slices` and `teal_slice` functions. 
 For details please see the *Filter panel for developers* vignette.
 
-```{r}
+```{r warning=FALSE, message=FALSE}
 library(shiny)
+library(DT)
 library(teal.slice)
 
 # create a FilteredData object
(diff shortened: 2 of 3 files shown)


llrs-roche:
I think it would be better to use the `collapse = TRUE` knitr option to avoid breaking the code chunks for each output than to hide everything. 

````

```{r, collapse=TRUE}
````` 

See the [examples on the book/guide](https://bookdown.org/yihui/rmarkdown-cookbook/opts-collapse.html)

File: vignettes/teal-slice.Rmd [hunk 1]
```diff
@@ -61,8 +61,9 @@ To facilitate communication with `FilteredData`, we provide the filter panel API
 The example below uses the `set_filter_state` function to set state and the state is specified with `teal_slices` and `teal_slice` functions. 
 For details please see the *Filter panel for developers* vignette.
 
-```{r}
+```{r warning=FALSE, message=FALSE}
 library(shiny)
+library(DT)
```
m7pr, line 66: @vedhav maybe instead of loading `DT` we can use `shiny::renderDataTable` and `shiny::dataTableOutput`

DESCRIPTION
```diff
@@ -55,6 +55,7 @@ Imports:
     teal.widgets (>= 0.4.2),
     utils
 Suggests:
+    DT,
```
m7pr, line 58: but I rather use shiny equivalents, it this is just for one example in a vignette

```suggestion
    DT (>= 0.3.0),
```

vignettes/teal-slice.Rmd
vedhav: That's the problem I wanted to solve. [We get this warning](https://insightsengineering.github.io/teal.slice/main/articles/teal-slice.html) if we do not use DT.

```
## `shiny::dataTableOutput()` is deprecated as of shiny 1.8.1.
## Please use `DT::DTOutput()` instead.
## Since you have a suitable version of DT (>= v0.32.1), shiny::dataTableOutput() will automatically use DT::DTOutput() under-the-hood.
## If this happens to break your app, set `options(shiny.legacy.datatable = TRUE)` to get the legacy datatable implementation (or `FALSE` to squelch this message).
## See <https://rstudio.github.io/DT/shiny.html> for more information.
## `shiny::dataTableOutput()` is deprecated as of shiny 1.8.1.
## Please use `DT::DTOutput()` instead.
## Since you have a suitable version of DT (>= v0.32.1), shiny::dataTableOutput() will automatically use DT::DTOutput() under-the-hood.
## If this happens to break your app, set `options(shiny.legacy.datatable = TRUE)` to get the legacy datatable implementation (or `FALSE` to squelch this message).
## See <https://rstudio.github.io/DT/shiny.html> for more information.
``

DESCRIPTION
vedhav: In that case I think it will be nice to add version in teal as well.

vedhav: never mind it is on its way to CRAN. It can stay like that for now.

m7pr (approved):

PR #644, opened 2025-03-06
Implement bslib and use Bootstrap 5

vedhav:
Part of https://github.com/insightsengineering/coredev-tasks/issues/587


gogonzo (approved):
👍

PR #645, opened 2025-03-12
Fix reactivity

gogonzo:
Fixed logs and sealed some setter functions to not enter when unnecessary.


m7pr (approved):
Hey, I see mostly changes in the logging so not many complex changes.
When it comes to the reactivity, looks like all setters were modified so they work in the isolated mode (some already were isolated, like `set_selected`, or `set_list_remove`). 

They also change the object only if it's different that when it was before (based on `identical()` function).

PR #652, opened 2025-06-18
fix scheduled jobs

m7pr:
Running dependency tests: 

- 1 https://github.com/insightsengineering/teal.slice/actions/runs/15727995505
- 2 https://github.com/insightsengineering/teal.slice/actions/runs/15729228463
- 3 https://github.com/insightsengineering/teal.slice/actions/runs/15729938736
- 4 https://github.com/insightsengineering/teal.slice/actions/runs/15731093270
- 5 https://github.com/insightsengineering/teal.slice/actions/runs/15823883739

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -44,7 +44,7 @@ jobs:
         PKG_SYSREQS_DRY_RUN=true
       additional-repos: ${{ matrix.pair.repos }}
       extra-deps: |
-        matrixStats (>= 1.5.0)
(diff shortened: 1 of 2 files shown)


m7pr:
Hey @llrs-roche let's skip setting versions for Bioconductor packages.
This build will use the specific Bioconductor snapshot so will pull always the same version. We have no way testing smaller/older versions. This pipeline uses the newest version of Bioconductor and the newest Bioconductor packages were pulled.

We had this conversation with the team in the past before, about those 2 bioc packages in teal.slice. It should be ok as it is.

PR #654, opened 2025-07-09
chore: bump teal.logger dependency to 0.4.0 and remove from Remotes

m7pr:
This PR updates all DESCRIPTION files to require teal.logger (>= 0.4.0) and removes it from Remotes.


donyunardi (changes requested):
Just occurred to me, shouldn't logger version also updated to 0.4.0 in Imports since teal.logger 0.4.0 requires logger 0.4.0?

donyunardi (approved):

PR #658, opened 2025-07-23
Fixes R6 warnings when loading package

### LINKED ISSUE #657: [Bug]: R6 `public$finalize()` warnings when loading package
### What happened?

R6 has added a deprecation message 4 years ago about `finalize()` methods needing to be private. This appears when loading the package via `pkgload::load_all()` but not with `library()` calls.

Image

@llrs-roche provided with context:

> Is a message and we might be contacted by the R6 team in advance as it would break many CRAN packages. Not sure how easy is it to fix, but it's been like that for 5 years (I asked if they have an estimation for it to become a warning or an error: https://github.com/r-lib/R6/issues/146). Given that it is on teal.slice it could be done on the following release.
>
> https://github.com/r-lib/R6/issues/146

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

averissimo:
# Pull Request

- Fixes #657

### Changes description

- Move `finalize()` methods
- Create public `destroy()` methods as these methods are called recursively

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ### Miscellaneous
 
-* In R6 classes `finalize()` method is moved to private and is replaced in public API by `destroy()`.
+* In R6 classes `finalize()` method is moved to private and is replaced in public API by `destroy()` (#658).
 
 # teal.slice 0.6.0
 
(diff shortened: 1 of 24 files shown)


llrs-roche (approved):
Easy enough 😄🚀

PR #665, opened 2025-08-19
Increase idle duration

vedhav:
Increasing the timeout to fix the failures in servers with less compute resources.

This is the failure in our integration test runs:

```
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test-shinytest2_DataframeFilteredDataset.R:72:5'): Toggle button shows and hide: 'Active Filter Summary' panel ──
  is_visible(app_driver, "#filter_panel-overview-table") is not TRUE
  
  `actual`:   FALSE
  `expected`: TRUE 
  ── Failure ('test-shinytest2_DataframeFilteredDataset.R:126:3'): Clicking add button on the datasets shows add filter panel ──
  is_existing(app_driver, "#filter_panel-active-mtcars-add_filter_icon") is not TRUE
  
  `actual`:   FALSE
  `expected`: TRUE 
  ── Error ('test-shinytest2_DataframeFilteredDataset.R:127:3'): Clicking add button on the datasets shows add filter panel ──
  Error in `app_find_node_id(self, private, input = input, output = output, 
      selector = selector)`: Cannot find HTML element with selector #filter_panel-active-mtcars-add_filter_icon
  Backtrace:
      ▆
   1. └─app_driver$click(selector = "#filter_panel-active-mtcars-add_filter_icon") at test-shinytest2_DataframeFilteredDataset.R:127:3
   2.   └─shinytest2:::app_click(...)
   3.     └─shinytest2:::app_find_node_id(self, private, input = input, output = output, selector = selector)
   4.       └─shinytest2:::app_abort(...)
   5.         └─rlang::abort(..., app = self, call = call)
  
  [ FAIL 3 | WARN 0 | SKIP 0 | PASS 837 ]
```

### DIFF
--- a/tests/testthat/test-shinytest2_DataframeFilteredDataset.R
+++ b/tests/testthat/test-shinytest2_DataframeFilteredDataset.R
@@ -71,14 +71,14 @@ testthat::describe("Toggle button shows and hide", {
     app_driver <- local_app_driver()
     testthat::expect_true(is_visible(app_driver, "#filter_panel-overview-table"))
     app_driver$click(selector = "#filter_panel-overview-main_filter_accordion * button")
-    app_driver$wait_for_idle(timeout = default_idle_timeout * 2)
+    app_driver$wait_for_idle(timeout = default_idle_timeout * 8)
     testthat::expect_false(is_visible(app_driver, "#filter_panel-overview-table"))
   })
   it("'Filter Data' panel", {
     app_driver <- local_app_driver()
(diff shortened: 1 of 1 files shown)


averissimo (approved):
I think this timeout was placed due to Github CI timing out, `* 8` seems like the necessary value for gitlab.

I'm wondering if on the long-term we should override this with a environment variable with defaulting to `* 2`

Or have a function in `TealAppDriver` if this continues to appear

Let's say `TealAppDriver$wait_for_long_idle()`

PR #680, opened 2026-05-21
add more unit tests

donyunardi:
Fixes https://github.com/insightsengineering/coredev-tasks/issues/754

Increased coverage to 87.16%

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,4 +1,3 @@
-# nocov start
 .onLoad <- function(libname, pkgname) { # nolint
   # adapted from https://github.com/r-lib/devtools/blob/master/R/zzz.R
   teal_default_options <- list(teal.threshold_slider_vs_checkboxgroup = 5)
@@ -46,7 +45,6 @@
   teal.data::datanames
   teal.widgets::optionalSelectInput
 }
-# nocov end
 
 
 ### END ENSURE CHECK PASSES
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,3 +1,4 @@
+# nocov start
 .onLoad <- function(libname, pkgname) { # nolint
   # adapted from https://github.com/r-lib/devtools/blob/master/R/zzz.R
   teal_default_options <- list(teal.threshold_slider_vs_checkboxgroup = 5)
@@ -45,6 +46,7 @@
   teal.data::datanames
   teal.widgets::optionalSelectInput
 }
+# nocov end
 
 
 ### END ENSURE CHECK PASSES
--- a/tests/testthat/test-DatetimeFilterState.R
+++ b/tests/testthat/test-DatetimeFilterState.R
@@ -324,7 +324,7 @@ testthat::test_that("content_summary shows NA xmark icon when keep_na is FALSE a
   testthat::expect_match(summary_html, "text-danger")
 })
 
-testthat::test_that("content_summary shows NA check icon when keep_na is NULL", { # nolint
+testthat::test_that("content_summary shows NA check icon when keep_na is NULL", { # nolint: line_length.
   filter_state <- DatetimeFilterState$new(
     c(posixct, NA),
     slice = teal_slice(dataname = "data", varname = "variable")
--- a/tests/testthat/test-DatetimeFilterState.R
+++ b/tests/testthat/test-DatetimeFilterState.R
@@ -284,6 +284,165 @@ testthat::test_that("format returns a properly formatted string representation",
   )
 })
 
+# content_summary ----
+testthat::test_that("content_summary shows datetime range in summary-value span", {
+  filter_state <- DatetimeFilterState$new(
+    posixct,
+    slice = teal_slice(dataname = "data", varname = "variable")
+  )
+  summary_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$content_summary())
+  )
+  testthat::expect_match(summary_html, "summary-value\">2000-01-01 12:00:00")
+  testthat::expect_match(summary_html, "&ndash;")
+  testthat::expect_match(summary_html, "2000-01-01 12:00:09")
+})
+
+testthat::test_that("content_summary shows NA check icon when keep_na is TRUE and NA values exist", {
+  filter_state <- DatetimeFilterState$new(
+    c(posixct, NA),
+    slice = teal_slice(dataname = "data", varname = "variable", keep_na = TRUE)
+  )
+  summary_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$content_summary())
+  )
+  testthat::expect_match(summary_html, "fa-check")
+  testthat::expect_match(summary_html, "text-success")
+})
+
+testthat::test_that("content_summary shows NA xmark icon when keep_na is FALSE and NA values exist", {
+  filter_state <- DatetimeFilterState$new(
+    c(posixct, NA),
+    slice = teal_slice(
+      dataname = "data", varname = "variable", keep_na = FALSE
+    )
+  )
+  summary_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$content_summary())
+  )
+  testthat::expect_match(summary_html, "fa-xmark")
+  testthat::expect_match(summary_html, "text-danger")
+})
+
+testthat::test_that("content_summary shows NA check icon when keep_na is NULL", { # nolint
+  filter_state <- DatetimeFilterState$new(
+    c(posixct, NA),
+    slice = teal_slice(dataname = "data", varname = "variable")
+  )
+  summary_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$content_summary())
+  )
+  testthat::expect_match(summary_html, "fa-check")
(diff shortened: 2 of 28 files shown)


tests/testthat/test-DatetimeFilterState.R
```diff
@@ -284,6 +284,165 @@ testthat::test_that("format returns a properly formatted string representation",
...
+    )
+  )
+  summary_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$content_summary())
+  )
+  testthat::expect_match(summary_html, "fa-xmark")
+  testthat::expect_match(summary_html, "text-danger")
+})
+
+testthat::test_that("content_summary shows NA check icon when keep_na is NULL", { # nolint
```
osenan: please better specify which "#nolint" is better to ingore only the one relevant for this line

tests/testthat/test-EmptyFilterState.R
```diff
@@ -1,3 +1,60 @@
+# ui_inputs ----
+testthat::test_that("ui_inputs renders 'Variable contains missing values only' span", {
```
osenan: please include in the description of the test a reference to the `EmptyFilterState` because we use `ui_inputs` in other tests.

tests/testthat/test-EmptyFilterState.R
```diff
@@ -1,3 +1,60 @@
...
+    slice = teal_slice(dataname = "data", varname = "variable")
+  )
+  summary_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$content_summary())
+  )
+  testthat::expect_match(summary_html, "All empty", fixed = TRUE)
+})
+
+# cache_state ----
+testthat::test_that("cache_state raises an error because private$get_state() does not exist", {
```
osenan: can we rephrase this test description?


```suggestion
testthat::test_that("cache_state raises an error if private$get_state() does not exist", {
```
To me makes more sense as a test that it raises an error if something is missing.

tests/testthat/test-FilteredData-utils.R
```diff
@@ -0,0 +1,117 @@
...
+  testthat::with_mocked_bindings(
+    runjs = function(js_expr) {
+      captured_js <<- js_expr
+      invisible(NULL)
+    },
+    fn(...),
+    .package = "shinyjs"
+  )
+  captured_js
+}
```
osenan: whenever we need a function for test we tend to place it at the beginning of the file

tests/testthat/test-FilteredData-utils.R
```diff
@@ -0,0 +1,117 @@
...
+      teal.slice:::toggle_title, "panel_toggle", c("Show panel", "Hide panel")
+    )
+    testthat::expect_match(
+      generated_js,
+      "if (curr == 'Show panel') { $(button_id).attr('title', 'Hide panel');",
+      fixed = TRUE
+    )
+  })
+
+  testthat::it("sets title back to titles[1] in the else branch", {
```
osenan: I found this test too specific of the implementation but actually don't testing any defined behavior. Do we want to test a default if the title does not match enything? Either change the description to reflect that this scenario is a default or just remove this test if what we are testing is specific to this else block

tests/testthat/test-FilteredData.R
```diff
@@ -1267,3 +1267,106 @@ testthat::describe("test FilterStateExpr server", {
...
+    ui_html, 'id="test-remove_all_filters_ui"',
+    fixed = TRUE
+  )
+})
+
+testthat::test_that("ui_active renders the filters_active_count container hidden", {
+  filtered_data <- FilteredData$new(list(iris = iris))
+  ui_html <- as.character(
+    shiny::isolate(filtered_data$ui_active("test"))
+  )
```
osenan: Do we need four different tests? All tests are actually checking that specific components, which only depend on the implementation, are there. Maybe we could check the expected class (test 1) and in second test we verify that the expected content is there and we check the title and some content. I think the titles are too specific and they could be simplified.

R/zzz.R
osenan, line 1: Is this required for the autovalidate? We avoid using " nocov". Indeed, I do not think it makes a difference in the package coverage.

tests/testthat/test-FilteredData.R
```diff
@@ -1267,3 +1267,106 @@ testthat::describe("test FilterStateExpr server", {
...
+})
+
+testthat::test_that("ui_overview renders the table output placeholder", {
+  filtered_data <- FilteredData$new(list(iris = iris))
+  ui_html <- as.character(
+    shiny::isolate(filtered_data$ui_overview("test"))
+  )
+  testthat::expect_match(ui_html, 'id="test-table"', fixed = TRUE)
+})
+
```
osenan: Again, it seems to me we are creating four test but testing the same behavior, the ui_component has specific elements that are included in the implementation. Maybe we can simplify this with two tests, or a single test with different expectations. The first implementation we check that the class is the expected, the second we check that it has specific content, like the table and the main filter accordion

tests/testthat/test-LogicalFilterState.R
```diff
@@ -139,6 +140,123 @@ testthat::test_that("get_call adds is.na(variable) to returned call if keep_na i
...
+    c(TRUE, TRUE, TRUE),
+    slice = teal_slice(
+      dataname = "data", varname = "variable",
+      selected = c(TRUE, FALSE), multiple = TRUE
+    )
+  )
+  testthat::expect_true(
+    shiny::isolate(filter_state$.__enclos_env__$private$is_any_filtered())
+  )
+})
```
osenan: Maybe this test is redundant? Selected equals at all choices but in reality we is not as different as the previus test.

tests/testthat/test-LogicalFilterState.R
```diff
@@ -139,6 +140,123 @@ testthat::test_that("get_call adds is.na(variable) to returned call if keep_na i
...
+
+testthat::test_that("is_choice_limited is always FALSE for LogicalFilterState", {
+  filter_state <- LogicalFilterState$new(
+    c(TRUE, FALSE, NA),
+    slice = teal_slice(dataname = "data", varname = "variable", choices = FALSE)
+  )
+  testthat::expect_false(filter_state$.__enclos_env__$private$is_choice_limited)
```
osenan: if we do no have a public method, our rule is not to test unexported / private functions. I think we can remove this test.

tests/testthat/test-LogicalFilterState.R
```diff
@@ -139,6 +140,123 @@ testthat::test_that("get_call adds is.na(variable) to returned call if keep_na i
...
+    slice = teal_slice(dataname = "data", varname = "variable", multiple = FALSE)
+  )
+  ui_html <- as.character(
+    shiny::isolate(filter_state$.__enclos_env__$private$ui_inputs("test"))
+  )
+  testthat::expect_match(ui_html, "shiny-input-radiogroup")
+  testthat::expect_no_match(ui_html, "shiny-input-checkboxgroup")
+})
+
+testthat::test_that("ui_inputs renders checkboxGroupInput when multiple is TRUE", {
```
osenan:
```suggestion
testthat::test_that("ui_inputs renders the expected component when multiple is TRUE", {
```
It is better to describe the behavior than the specific implementation

osenan (changes requested):
Good job! We have greatly improved the coverage, but I think some test can be simplified into one and some descriptions can be changed as well. In general is better not to include specific implementation in the description.

R/zzz.R
donyunardi: I agree, let me remove the nocov. It only raises it by 0.3%.

tests/testthat/test-DatetimeFilterState.R
donyunardi: Done in 27708b83.

tests/testthat/test-EmptyFilterState.R
donyunardi: Done in 27708b83

tests/testthat/test-FilteredData-utils.R
donyunardi: Done in 27708b83

R/zzz.R
donyunardi: Done in 27708b83

tests/testthat/test-FilteredData.R
donyunardi: Updated in 27708b83

tests/testthat/test-LogicalFilterState.R
donyunardi: Updated in 27708b83

tests/testthat/test-LogicalFilterState.R
donyunardi: Removed in 27708b83

tests/testthat/test-FilteredData.R
donyunardi: Updated in 27708b83

tests/testthat/test-FilteredData-utils.R
donyunardi: Removed in d3fedb04

osenan (approved):
LGTM now!

PR #688, opened 2026-07-23
Fix reactiveVal bug with latest shiny release

### LINKED ISSUE #682: Heads-up: shiny session cleanup will affect teal_slice tests (temporarily deferred for 1.14.0)
Hi! A heads-up from the shiny team, and a request for your thoughts since some of this is still in flight.

In shiny's upcoming 1.14.0 we added module-scoped reactive cleanup ([rstudio/shiny#4372](https://github.com/rstudio/shiny/pull/4372)): a `reactiveValues` now registers with the active reactive domain when it's created and is torn down when that session is destroyed. `teal_slice()` builds its object with `reactiveValues()` (`R/teal_slice.R:140`, `:163`), so a slice created while a session is active becomes bound to that session.

We caught this in our reverse-dependency checks: in your tests (e.g. `test-FilterStates.R` "Selecting a new variable…", "Adding 'var_to_add'…") a `FilterStates` object is created outside `testServer()`, mutated inside it, then read *after* the block via `get_filter_state()`. Since `testServer()` closes its session on exit, the slices get destroyed and the later read errors with *"its module session has been destroyed."*

**You don't need to do anything for 1.14.0.** We've temporarily disabled this cleanup inside `testServer()` for the release so your current tests keep passing — see [rstudio/shiny#4396](https://github.com/rstudio/shiny/pull/4396). We do plan to re-enable it in a later release (timing not yet fixed), so we wanted to flag it early.

When you're ready, the fix we'd suggest — which also makes things more robust in real apps — is to create the slice's `reactiveValues` outside any reactive domain, since a `teal_slice` is meant to be session-independent:

```r
ans <- withReactiveDomain(NULL, do.call(reactiveValues, args))
```

That said, we're still feeling out the long-term story here, so if this binding behavior is awkward for teal.slice's design we'd genuinely like to hear it — it may influence how/whether we re-enable.

averissimo:
# Pull Request

- Fixes #682 (part of)

### DIFF
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -107,7 +107,7 @@ FilterState <- R6::R6Class( # nolint
             varlabel
           }
 
-        private$state_history <- reactiveVal(list())
+        private$state_history <- withReactiveDomain(NULL, reactiveVal(list()))
 
         invisible(self)
       })
--- a/R/FilterState.R
+++ b/R/FilterState.R
@@ -43,7 +43,6 @@ FilterState <- R6::R6Class( # nolint
 
   # public methods ----
   public = list(
-
     #' @description
     #' Initialize a `FilterState` object.
     #'
--- a/R/FilterStates.R
+++ b/R/FilterStates.R
@@ -64,7 +64,7 @@ FilterStates <- R6::R6Class( # nolint
       }
       private$data <- data
       private$data_reactive <- data_reactive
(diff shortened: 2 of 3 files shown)


llrs-roche (approved):
Tested with examples on teal.osprey I haven't found any other reactiveVal outside a reactive environment

